PR #1, opened 2026-01-13
docs: Batch inference design doc

lioraron:
## Description

This PR adds a design document for batch inference.


docs/design/batch_inference_architecture.md
```diff
@@ -0,0 +1,116 @@
...
+
+Jan 2026  
+Revision 4
+
+## Customer use cases
+
+The use case is running asynchronously a set of inference requests which are grouped and tracked together in a job, where the completion of these inference requests is more time tolerant relative to interactive (time-sensitive) inference requests.  
+
+The time tolerance for processing of interactive inference requests is typically on the scale of seconds and up to minutes.  
+The time tolerance for processing batch jobs is typically on the scale of hours, for example 12 house, 24 hours, 48 hours.
```
yizhaodev: `12 house` is typo?

docs/design/batch_inference_architecture.md
```diff
@@ -0,0 +1,116 @@
...
+
+The API should be compatible with [Batch API - OpenAI API](https://platform.openai.com/docs/guides/batch).  
+
+1. The user submits a batch job that includes metadata and a reference to a jsonl file, where each line in the file specifies an independent inference request, that includes a model id, inference api, prompt, and inference parameters. The batch job metadata may include:
+   - An SLO requirement, for example 24 hours. The SLO can be defined as a requirement for beginning processing or for completing processing.
+   - A maximum charging threshold, and/or a maximum error threshold, where exceeding these thresholds would trigger stopping the job.
+   - Retries policies.
+   - Default or override inference parameter values on various levels, e.g. global, api type, model level.
+1. The output of the inference requests is written to an output file and an error file, which the user can then retrieve / access. Responses are correlated to requests using a custom id.
+1. It is possible to support multiple input files and multiple output files per job.
```
j-mok-dev: It might be too early for a detailed discussion, but I’m concerned about how we’ll define the job status when only some files succeed.
OpenAI’s Batch API considers a job completed even if there are errors in individual request lines. We need to decide whether we will follow this same behavior, especially in cases where an entire file is corrupted or fails. Should the entire job be marked as failed, or should we introduce a `partial_success` status to indicate that some files were processed while others were skipped? We can follow the openAI way but it can be misleading. if a user expects 10 files but only 1 file is proceed, it still shows `completed`

docs/design/batch_inference_architecture.md
lioraron: Done, thanks.

docs/design/batch_inference_architecture.md
```diff
@@ -0,0 +1,116 @@
...
+1. Users can list their batch jobs, track the status of their batch jobs, and can control their jobs, for example pause, resume, cancel or delete a job.
+
+## High level system requirements
+
+1. Process a large number of grouped inference requests in an asynchronous way.
+1. Minimum interference to satisfying interactive requests SLOs.
+1. Satisfy batch jobs’ SLOs.
+1. When there is less interactive workload (enabling resource capacity) - increase the processing rate of batch workload, and when there is more interactive workload - decrease the processing rate of batch workload.
+1. Control the flow of interactive and batch requests to maximize GPU utilization, minimize resource costs, and satisfy SLOs.
```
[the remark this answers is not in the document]
lioraron: > 1. Queue Sorting & SLO: How will the queue be prioritized? Should we implement a sorting logic that calculates SLOs to ensure urgent jobs are dispatched first? If so, which component will be responsible for this sorting (the Queue itself, or an external Orchestrator)? For MVP, I guess we might start with FIFO but do we have plans to implement queue sorting?

Assuming this is the batch jobs priority queue: Currently will be sorted based on the target time = submission time + SLO value. The DB layer will do the sorting.

> 2. Control Authority & Preemption: Since the Autoscaler controls the number of Processor pods based on metrics from EPP flow control, what happens during an interactive traffic spike?

This should be explored in the context of the batch auto-scaling logic.
I'm guessing something like:
1. First priority is to scale the model servers. In this case it will be increase interactive servers, decrease batch servers.
2. Seconds priority: If we reached limits on scaling model servers and still can't satisfy SLOs - scale the batch processors. In this case scale these down.

>    Can the Autoscaler terminate currently running pods to free up resources immediately? then how will we handle the loss of partially processed data and the need for reprocessing?

- At this time - no preemption.
- But the batch processor should handle recovery for abnormal termination / crashes, with some checkpointing mechanism? Was this implemented in wx? If not, it can be post mvp.

>    If it only stops creating new pods, the interactive requests might be blocked or delayed until the current batch pods finish their tasks.

Yes.

>    Should we consider implementing a pause/resume mechanism for the processors to avoid data loss while freeing up resources?

Yes, but not for mvp.

> 3. Role of the Auto-scaler vs. Orchestrator: Is the Auto-scaler intended to replace the Batch Orchestrator role (similar to our previous implementation)? Will the Auto-scaler pull a job from the queue and pass the Job ID to a new Processor? Or will it simply spawn a Worker pod, which then independently pulls a job from the queue?

The auto-scaler only modifies the k8s deployment of the batch processor - updating the number of instances.
The batch processor should be similar to that in wx. It's responsible to pull a job, process it, write the results, and update the status.

docs/design/batch_inference_architecture.md
lioraron: The more detailed info we provide the better, with keeping openai compatibility. Namely, we can provide additional info, in addition to the openai indicators.
Pls provide a proposal for the statuses?

yizhaodev:
LGTM, thanks

j-mok-dev (commented):
Thanks, Lior. lgtm :D

j-mok-dev (approved):

PR #4, opened 2026-01-21
feat: Go Client to interact with the Batch Processer and Scheduler

RishabhSaini:
Core Implementation:
- OpenAI-compatible HTTP client (`/v1/chat/completions`, `/v1/completions`)
- Exponential backoff retry with jitter (configurable: max retries, backoff factor, jitter fraction)
- Error categorization (rate limit, auth, server, invalid request) with retry logic
- Connection pooling with configurable timeouts

Testing:
  - Unit: Client logic, retry algorithm, error handling, configuration
  - Integration: Real HTTP behavior, latency simulation, failure injection
  - Each integration test spawns isolated `llm-d-inference-sim` Docker mock server with specific config (latency, failure injection)

Fixes:
  - Fix `BatchStatus` constant naming conflict with Ginkgo (`Pending` → `StatusPending`, etc.)
  - Update all references in `worker.go`

### DIFF
--- a/internal/shared/batch/http_inference_client_test.go
+++ b/internal/shared/batch/http_inference_client_test.go
@@ -30,7 +30,22 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestNewHTTPInferenceClient(t *testing.T) {
+// TestInferenceClient aggregates all HTTPInferenceClient test cases
+// Run with: go test -run TestInferenceClient
+func TestInferenceClient(t *testing.T) {
+	t.Run("NewHTTPInferenceClient", testNewHTTPInferenceClient)
+	t.Run("Generate", testGenerate)
+	t.Run("ErrorHandling", testErrorHandling)
+	t.Run("RetryLogic", testRetryLogic)
+	t.Run("TLSConfiguration", testTLSConfiguration)
+	t.Run("Authentication", testAuthentication)
+	t.Run("NetworkErrors", testNetworkErrors)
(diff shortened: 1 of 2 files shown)


internal/shared/batch/http_inference_client.go
```diff
@@ -0,0 +1,280 @@
...
+		RequestID: req.RequestID,
+		Response:  resp.Body(),
+		RawData:   rawData,
+	}, nil
+}
+
+// handleRequestError processes request-level errors (network, timeout, cancellation)
+func (c *HTTPInferenceClient) handleRequestError(ctx context.Context, err error, req *InferenceRequest) (*InferenceResponse, *InferenceError) {
+	// Check if context was cancelled or timed out
+	if ctx.Err() == context.Canceled {
```
yizhaodev: safe even if ctx.Err() is wrapped
```
if errors.Is(ctx.Err(), context.Canceled) {
}
```

internal/shared/batch/http_inference_client_test.go
```diff
@@ -0,0 +1,1113 @@
...
+	"net/http/httptest"
+	"reflect"
+	"strings"
+	"testing"
+	"time"
+)
+
+// Test helper functions
+
+func assertEqual(t *testing.T, got, want interface{}, msgAndArgs ...interface{}) {
```
yizhaodev: maybe we should use https://github.com/stretchr/testify

PR #15, opened 2026-01-28
docs: Async inference handling design diagram

lioraron:
Adding a diagram of alternatives designs for async inference handling.


yizhaodev:
we still call it "inference proxy"?

PR #18, opened 2026-01-28
InferenceClient between BatchProcesser and Scheduler

RishabhSaini:


internal/inference/client_integration_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+	client := NewHTTPClient(HTTPClientConfig{
+		BaseURL: fmt.Sprintf("http://localhost:%d", testPort),
+		Timeout: 10 * time.Second,
+	})
+
+	t.Run("should successfully make text completion request", func(t *testing.T) {
+		req := &GenerateRequest{
+			RequestID: "test-completion-001",
+			Model:     "fake-model",
```
yizhaodev: so we have to put model id twice in GenerateRequest? is the GenerateRequest.model needed?

internal/inference/client_test.go
```diff
@@ -0,0 +1,1245 @@
...
+				MaxRetries:     3,
+				InitialBackoff: 500 * time.Millisecond,
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			client := NewHTTPClient(tt.config)
+			assert.NotNil(t, client)
```
yizhaodev: if `client` is nil, `assert.NotNil` will fail the case and move to next line, then you will have panic
should use `require.NotNil`, which will abort subcase when nil

RishabhSaini: updated

internal/inference/client_test.go
```diff
@@ -0,0 +1,1245 @@
...
+
+		req := &GenerateRequest{
+			RequestID: "test",
+			Model:     "gpt-4",
+			Endpoint:  "/v1/chat/completions",
+			Params:    map[string]interface{}{"model": "gpt-4"},
+		}
+
+		ctx, cancel := context.WithCancel(context.Background())
+		cancel() // Cancel immediately
```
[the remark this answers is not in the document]
RishabhSaini: updated

yizhaodev (approved):
thanks @RishabhSaini for the great work! Overall  it looks good to me.
looks like @j-mok-dev requested another change before merge, you can work with her.

PR #2, opened 2026-01-19
Initial draft for rest api server

yizhaodev:
This is initial draft for rest api server, which include
* rest server using native golang http server
* unit test cases using native golang test framework
* `k8s.io/klog/v2` based logger
* OpenAI object struct definition
* endpoints:
    * ` /batch`: placeholder and not implemented
    * `/files`: placeholder and not implemented
    * `/metrics`: add 3 simple prometheus metrics for now
    * `/health`: simple health support GET and HEAD
* middleware to intercept all requests:
    * SecurityHeadersMiddleware: adding security headers
    * RequestMiddleware: 
        * generate request id and logger for each request, and attach to request context
        * auto increase/decrease counter metrics
    * RecoveryMiddleware: catch and recover from all panic generated by business endpoints like `/batch`

@lioraron please let me know who else should be invited to review.

PR #39, opened 2026-02-13
feat(postgresql): implement PostgreSQL database backend

acardace:
## Summary

Implements `DBClient[T, Q]` using PostgreSQL via `pgx/v5`, supporting both `BatchDBClient` and `FileDBClient`.

## Design

The implementation uses a **table-descriptor pattern** with a shared `pgCore` engine that handles all SQL operations generically. Each domain type provides a `TableDescriptor` that declares:

- **Table name and schema DDL** (via `TableName()` / `Schema()`) — each type can define a different table structure
- **Column layout** (via `HasStatus()` / `ExtraColumns()`) — controls which columns exist beyond the common set (`id`, `tenant_id`, `expiry`, `tags`, `spec`)
- The core builds INSERT, SELECT, UPDATE, and DELETE SQL dynamically from the descriptor

This means adding a new domain type requires only a new descriptor, a schema file, and a thin client wrapper — with zero changes to the SQL engine.

### Concrete descriptors

- **`batchDescriptor`**: includes a `status` JSONB column (`HasStatus() = true`), no extra columns.
- **`fileTableDescriptor`**: no status column, adds a `purpose` extra column for file-level filtering.

### Schema

Each descriptor owns its table DDL via `go:embed`. Tables share a common shape (`id`, `tenant_id`, `expiry`, `tags`, `spec`) with descriptor-specific columns. Indexes on `tenant_id` and `expiry` (partial) support the primary query patterns.

## Changes

- Add `pgx/v5` and `pgxmock/v4` dependencies
- Add `PostgresConfig` with validation and connection string building
- Add `pgxPool` interface for testability (satisfied by `*pgxpool.Pool` and `pgxmock`)
- Add `TableDescriptor` interface and `pgCore` shared SQL engine (`db_core.go`)
- Add `batchDescriptor` + `PostgresBatchDBClient` (`batch_db.go`)
- Add `fileTableDescriptor` + `PostgresFileDBClient` (`file_db.go`)
- Add embedded SQL schemas (`batch_schema.sql`, `file_schema.sql`)
- Tests using `pgxmock` covering CRUD operations, config validation, context management

### DIFF
--- a/internal/database/postgresql/file_db.go
+++ b/internal/database/postgresql/file_db.go
@@ -1,0 +1,129 @@
+/*
+Copyright 2026 The llm-d Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package postgresql
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+
+	"github.com/llm-d-incubation/batch-gateway/internal/database/api"
+	"k8s.io/klog/v2"
+)
+
+//go:embed file_schema.sql
+var fileSchemaSql string
+
+const (
+	colPurpose = "purpose"
+)
+
+// fileTableDescriptor implements TableDescriptor for file items.
+type fileTableDescriptor struct{}
+
+func (fileTableDescriptor) TableName() string      { return "file_items" }
+func (fileTableDescriptor) Schema() string         { return fileSchemaSql }
+func (fileTableDescriptor) ExtraColumns() []string { return []string{colPurpose} }
+
+// PostgresFileDBClient implements api.FileDBClient using PostgreSQL.
+type PostgresFileDBClient struct {
+	*pgCore
+}
+
+var _ api.FileDBClient = (*PostgresFileDBClient)(nil)
+
+// NewPostgresFileDBClient creates a new PostgreSQL file database client.
+func NewPostgresFileDBClient(ctx context.Context, config *PostgreSQLConfig) (*PostgresFileDBClient, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+
+	pgCore, err := newPgCore(ctx, config, fileTableDescriptor{})
+	if err != nil {
+		return nil, err
+	}
+
+	klog.FromContext(ctx).Info("NewPostgresFileDBClient: client created successfully")
+	return &PostgresFileDBClient{pgCore}, nil
+}
+
+func (c *PostgresFileDBClient) Close() error {
+	return c.close()
+}
+
+func (c *PostgresFileDBClient) DBStore(ctx context.Context, item *api.FileItem) (err error) {
+	if item == nil {
+		err = fmt.Errorf("item is nil")
+		return
+	}
+
+	if err = c.store(ctx, &item.BaseIndexes, &item.BaseContents, map[string]any{
+		colPurpose: item.Purpose,
+	}); err != nil {
+		return
+	}
+	return
+}
+
+func (c *PostgresFileDBClient) DBGet(
+	ctx context.Context, query *api.FileQuery,
+	includeStatic bool, start, limit int,
+) (items []*api.FileItem, cursor int, expectMore bool, err error) {
+	if query == nil {
+		return
+	}
+
+	var extraFilters map[string]any
+	if query.Purpose != "" {
+		extraFilters = map[string]any{colPurpose: query.Purpose}
+	}
+
+	indexes, contents, extras, cursor, expectMore, err := c.get(
+		ctx, &query.BaseQuery, includeStatic, start, limit, extraFilters)
+	if err != nil {
+		return
+	}
+
+	items = make([]*api.FileItem, len(indexes))
+	for i := range indexes {
+		items[i] = &api.FileItem{
+			BaseIndexes:  *indexes[i],
+			BaseContents: *contents[i],
+			Purpose:      extras[i][colPurpose],
+		}
+	}
+
+	return
+}
+
+func (c *PostgresFileDBClient) DBUpdate(ctx context.Context, item *api.FileItem) (err error) {
+	if item == nil {
+		err = fmt.Errorf("item is nil")
+		return
+	}
+	if err = c.update(ctx, &item.BaseIndexes, &item.BaseContents); err != nil {
+		return
+	}
+	return
+}
+
+func (c *PostgresFileDBClient) DBDelete(ctx context.Context, ids []string) (deletedIDs []string, err error) {
+	if deletedIDs, err = c.delete(ctx, ids); err != nil {
+		return
+	}
+	return
+}
--- a/internal/database/postgresql/file_db_test.go
+++ b/internal/database/postgresql/file_db_test.go
@@ -1,0 +1,260 @@
+/*
+Copyright 2026 The llm-d Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package postgresql
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/pashagolub/pgxmock/v4"
+
+	"github.com/llm-d-incubation/batch-gateway/internal/database/api"
+)
+
+const (
+	fileTestTable    = "file_items"
+	fileTestTenantID = "tenant-1"
+)
+
+func newTestFileClient(t *testing.T) (*PostgresFileDBClient, pgxmock.PgxPoolIface) {
+	t.Helper()
+
+	mock, err := pgxmock.NewPool()
+	if err != nil {
+		t.Fatalf("failed to create pgxmock pool: %v", err)
+	}
+
+	client := &PostgresFileDBClient{
+		pgCore: &pgCore{pool: mock, desc: fileTableDescriptor{}},
+	}
+
+	return client, mock
+}
+
+func newTestFileItem(id, tenantID string) *api.FileItem {
+	return &api.FileItem{
+		BaseIndexes: api.BaseIndexes{
+			ID:       id,
+			TenantID: tenantID,
+			Expiry:   time.Now().Add(24 * time.Hour).Unix(),
+			Tags:     api.Tags{"purpose": "batch"},
+		},
+		Purpose: "batch",
+		BaseContents: api.BaseContents{
+			Spec: []byte(`{"filename":"input.jsonl","bytes":1024}`),
+		},
+	}
+}
+
+func TestFileStore(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("stores file item successfully", func(t *testing.T) {
+		client, mock := newTestFileClient(t)
+		defer mock.Close()
+
+		item := newTestFileItem("file-1", fileTestTenantID)
+
+		mock.ExpectExec("INSERT INTO "+fileTestTable).
+			WithArgs("file-1", fileTestTenantID, pgxmock.AnyArg(), pgxmock.AnyArg(), pgxmock.AnyArg(), pgxmock.AnyArg(), pgxmock.AnyArg()).
+			WillReturnResult(pgxmock.NewResult("INSERT", 1))
+
+		if err := client.DBStore(ctx, item); err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+
+		if err := mock.ExpectationsWereMet(); err != nil {
+			t.Fatalf("unmet expectations: %v", err)
+		}
+	})
+
+	t.Run("returns error for nil item", func(t *testing.T) {
+		client, mock := newTestFileClient(t)
+		defer mock.Close()
+
+		if err := client.DBStore(ctx, nil); err == nil {
+			t.Fatal("expected error for nil item")
+		}
+	})
+}
+
+func TestFileGet(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("returns nil for nil query", func(t *testing.T) {
+		client, mock := newTestFileClient(t)
+		defer mock.Close()
+
+		items, cursor, expectMore, err := client.DBGet(ctx, nil, true, 0, 10)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		if items != nil || cursor != 0 || expectMore {
+			t.Fatal("expected empty result for nil query")
+		}
+	})
+
+	t.Run("gets file by tenant ID with purpose", func(t *testing.T) {
+		client, mock := newTestFileClient(t)
+		defer mock.Close()
+
+		item := newTestFileItem("file-1", fileTestTenantID)
+		tags, _ := packTags(item.Tags)
+
+		rows := pgxmock.NewRows([]string{colID, colTenantID, colExpiry, colTags, colPurpose, colStatus, colSpec}).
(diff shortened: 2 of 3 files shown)


internal/database/postgresql/batch_schema.sql
```diff
@@ -0,0 +1,15 @@
+-- Copyright 2026 The llm-d Authors
+-- Licensed under the Apache License, Version 2.0
```
lioraron: Possibly use the same copyright comment as in the golang files.

internal/database/postgresql/batch_schema.sql
```diff
@@ -0,0 +1,15 @@
...
+    tenant_id  TEXT NOT NULL,
+    expiry     BIGINT,
+    tags       JSONB,
+    spec       JSONB,
+    status     JSONB
+);
+
+CREATE INDEX IF NOT EXISTS idx_batch_items_tenant_id ON batch_items(tenant_id);
+CREATE INDEX IF NOT EXISTS idx_batch_items_expiry ON batch_items(expiry) WHERE expiry IS NOT NULL;
+CREATE INDEX IF NOT EXISTS idx_batch_items_tags ON batch_items USING GIN (tags);
```
lioraron: For the tags index there is no need for WHERE ... IS NOT NULL?

internal/database/postgresql/db_core.go
```diff
@@ -0,0 +1,462 @@
...
+type TableDescriptor interface {
+	// TableName returns the PostgreSQL table name.
+	TableName() string
+	// Schema returns the DDL SQL to create the table and indexes.
+	// Must be idempotent (use IF NOT EXISTS).
+	Schema() string
+	// HasStatus returns true if the table has a status JSONB column.
+	HasStatus() bool
+	// ExtraColumns returns names of additional indexed columns
+	// beyond the common set (id, tenant_id, expiry, tags, spec, [status]).
```
lioraron: `spec` and `status` are not indexed columns, so should they be included in this comment?

internal/database/postgresql/db_core.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+	if timeLimit > 0 {
+		return context.WithTimeout(parentCtx, timeLimit)
+	}
+	return context.WithTimeout(parentCtx, defaultContextTimeout)
+}
+
+func (c *pgCore) close() error {
+	if c.pool != nil {
+		c.pool.Close()
```
lioraron: This function doesn't return an error?

internal/database/postgresql/batch_db.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/klog/v2"
+)
+
+//go:embed batch_schema.sql
+var batchSchemaSql string
+
+// batchDescriptor implements TableDescriptor for batch items.
+type batchDescriptor struct{}
+
+func (batchDescriptor) TableName() string      { return "batch_items" }
```
lioraron: In case the tables names appear in more than one location, consider defining these as const in one place then reusing.

internal/database/postgresql/batch_db.go
```diff
@@ -0,0 +1,125 @@
...
+	includeStatic bool, start, limit int,
+) (items []*api.BatchItem, cursor int, expectMore bool, err error) {
+	if query == nil {
+		return
+	}
+
+	indexes, contents, _, cursor, expectMore, err := c.get(
+		ctx, &query.BaseQuery, includeStatic, start, limit, nil)
+	if err != nil {
+		return
```
lioraron: No error logging?

lioraron:
@acardace Here is the code review from Claude:

  I've reviewed the pull request thoroughly. This is a substantial addition implementing PostgreSQL support with a clean architecture. Here's my detailed analysis:

  ---
  Overview

  The PR implements a PostgreSQL backend for the batch-gateway using pgx/v5, with a table-descriptor pattern for extensibility.

  Stats: +1,948 additions, -3 deletions across 11 files

  ---
  ✅ Strengths

  1. Excellent Architecture - Table Descriptor Pattern

  The TableDescriptor interface is elegant and extensible:
  type TableDescriptor interface {
      TableName() string
      Schema() string
      HasStatus() bool
      ExtraColumns() []string
  }

  This allows adding new domain types without modifying core SQL logic. Very clean separation of concerns.

  2. Comprehensive Test Coverage

  - Uses pgxmock for isolated unit testing
  - Tests cover: nil inputs, validation errors, DB failures, pagination, tag selectors (AND/OR), expiry conditions
  - Round-trip testing for tag serialization
  - Edge cases like empty queries, non-existent IDs

  3. Proper SQL Practices

  - Parameterized queries - All queries use $1, $2 placeholders preventing SQL injection
  - GIN indexes on JSONB columns for efficient tag queries
  - Partial indexes on nullable columns (expiry, purpose)
  - RETURNING clause in DELETE for tracking deleted IDs
  - DB-side NOW() for expiry checks to avoid clock skew

  4. Good Error Handling

  - Validation before DB operations
  - Nil-safe checks throughout
  - Context timeout support (30s default)
  - Pool cleanup on initialization failures

  5. Clean Code Organization

  - Embedded SQL schemas (//go:embed)
  - Consistent logging with structured fields
  - Clear separation between core logic and domain clients

  ---
  🔍 Critical Issues

  1. Password Escaping Bug ⚠️

  The connection string escaping is incorrect:

  escapedPassword := "'" + strings.NewReplacer(`\`, `\\`, `'`, `\'`).Replace(c.Password) + "'"
  return fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=%s",
      c.Host, port, c.User, escapedPassword, c.Database, sslMode)

  Problems:
  1. The password shouldn't be quoted manually - pgx handles this internally
  2. Other fields aren't escaped - host, user, database could contain special chars
  3. Wrong escaping approach - This creates: password='escaped_value' which pgx will interpret as a literal string including the quotes

  Fix: Use pgx's built-in config parsing:
  func (c *PostgresConfig) connectionString() string {
      sslMode := c.SSLMode
      if sslMode == "" {
          sslMode = SSLModeDisable
      }
      port := c.Port
      if port == 0 {
          port = defaultPort
      }

      // Use pgx's config parser which handles escaping properly
      connStr := fmt.Sprintf("host=%s port=%d user=%s password=%s dbname=%s sslmode=%s",
          c.Host, port, c.User, c.Password, c.Database, sslMode)
      return connStr
  }

  Or better yet, use pgx.ParseConfig and set fields individually:
  func newPool(ctx context.Context, config *PostgresConfig) (pgxPool, error) {
      if config == nil {
          return nil, fmt.Errorf("config is nil")
      }
      if err := config.Validate(); err != nil {
          return nil, fmt.Errorf("invalid config: %w", err)
      }

      // Let pgx handle escaping
      pgxConfig, err := pgxpool.ParseConfig("")
      if err != nil {
          return nil, err
      }

      pgxConfig.ConnConfig.Host = config.Host
      pgxConfig.ConnConfig.Port = uint16(config.Port)
      pgxConfig.ConnConfig.User = config.User
      pgxConfig.ConnConfig.Password = config.Password
      pgxConfig.ConnConfig.Database = config.Database
      // Set SSL mode appropriately

      pool, err := pgxpool.NewWithConfig(ctx, pgxConfig)
      // ... rest of function
  }

  2. Thread Safety - Close() Race Condition

  The Close() methods check c.pool != nil without synchronization:

  func (c *pgCore) close() error {
      if c.pool != nil {
          c.pool.Close()
      }
      return nil
  }

  Issue: If multiple goroutines call Close() concurrently, there's a data race on the pool field check.

  Fix: Use sync.Once:
  type pgCore struct {
      pool      pgxPool
      desc      TableDescriptor
      closeOnce sync.Once
  }

  func (c *pgCore) close() error {
      c.closeOnce.Do(func() {
          if c.pool != nil {
              c.pool.Close()
          }
      })
      return nil
  }

  3. Missing Index on Combined Queries

  The schema has individual indexes on tenant_id, tags, and expiry, but common queries combine these:

  WHERE tenant_id = $1 AND tags @> $2 AND expiry > 0 AND expiry <= NOW()

  Recommendation: Add a composite index for hot path queries:
  CREATE INDEX IF NOT EXISTS idx_batch_items_tenant_tags ON batch_items(tenant_id)
      INCLUDE (tags, expiry) WHERE expiry IS NOT NULL;

  Or use a GIN composite index for JSONB + scalar columns if query patterns justify it.

  ---
  ⚠️ Potential Issues

  4. Tag Selector Query Performance

  The tag selector logic builds JSONB containment queries:

  tagJSON, _ := json.Marshal(map[string]string{k: v})
  tagConds = append(tagConds, fmt.Sprintf(colTags+" @> $%d::jsonb", argIdx))

  For TagsLogicalCond=OR with many selectors, this becomes:
  WHERE (tags @> '{"env":"prod"}' OR tags @> '{"team":"ml"}' OR ...)

  Issue: GIN index may not optimize OR queries well. Each condition requires a separate index scan.

  Recommendation:
  - Document this limitation
  - Consider materializing common tag combinations
  - For OR queries with >5 conditions, warn about performance

  5. Silent Error Swallowing

  In packTags, JSON marshaling errors are returned, but callers sometimes ignore them:

  tags, _ := packTags(idx.Tags)  // Error ignored in some places

  While unlikely to fail (string maps marshal reliably), it's inconsistent.

  Recommendation: Always check the error or document why it's safe to ignore.

  6. No Connection Pool Configuration

  The newPool function creates a pool with default settings. Production workloads may need tuning:
  - Max connections
  - Min connections
  - Connection lifetime
  - Acquire timeout

  Recommendation: Add pool configuration to PostgresConfig:
  type PostgresConfig struct {
      // ... existing fields
      MaxConns     int32
      MinConns     int32
      MaxConnLifetime time.Duration
  }

  7. Schema Migration Strategy

  The PR uses CREATE TABLE IF NOT EXISTS which:
  - ✅ Works for initial deployment
  - ❌ Doesn't handle schema changes (new columns, index modifications)
  - ❌ No version tracking

  Recommendation:
  - Add schema version table
  - Document that this is bootstrap-only
  - Or integrate a migration tool (golang-migrate, goose)

  8. No Prepared Statement Optimization

  Each query builds SQL strings dynamically without preparing statements. For hot paths (DBGet with same query patterns), this causes:
  - Repeated parsing overhead
  - No query plan caching

  Recommendation: Consider prepared statements for common query patterns in a future enhancement.

  ---
  📝 Code Quality Issues

  9. Error Messages Lack Context

  Many errors don't include operation context:
  if result.RowsAffected() == 0 {
      return fmt.Errorf("DBUpdate: item %s not found", idx.ID)
  }

  Should include table name:
  return fmt.Errorf("DBUpdate: item %s not found in %s", idx.ID, c.desc.TableName())

  10. Magic Numbers

  const defaultContextTimeout = 30 * time.Second
  const defaultPort = 5432

  Good! But missing:
  - Default max/min pool connections
  - Default acquire timeout

  11. Validation Inconsistency

  PostgresConfig.Validate() checks port range but allows Port: 0 (uses default). This is fine, but the validation message is confusing:

  if c.Port < 0 || c.Port > 65535 {
      return fmt.Errorf("port must be between 0 and 65535")
  }

  Should be: "port must be between 0 and 65535 (0 uses default 5432)"

  12. Missing Tests

  While coverage is good, missing:
  - Connection string generation tests (especially password escaping!)
  - Concurrent Close() test
  - Context timeout test
  - Pool initialization failure handling
  - Schema creation failure

  ---
  🎯 Minor Observations

  13. Empty Tags Serialization

  if len(tags) == 0 {
      return "", nil
  }

  Empty tags serialize to "" which gets stored as NULL. This is fine, but inconsistent with storing {} for empty JSON objects.

  Consider: Always store {} for consistency with JSONB expectations.

  14. Cursor Implementation

  cursor = start + len(indexes)
  expectMore = len(indexes) == limit

  This works but is opaque. The cursor is just an offset.

  Consider: Document that cursor semantics are simple offset-based, not cryptographic tokens.

  15. SQL Formatting

  Some queries use string concatenation:
  expiryCondition = colExpiry + " > 0 AND " + colExpiry + " <= EXTRACT(EPOCH FROM NOW())::BIGINT"

  Style preference: Use fmt.Sprintf for readability:
  expiryCondition = fmt.Sprintf("%s > 0 AND %s <= EXTRACT(EPOCH FROM NOW())::BIGINT", colExpiry, colExpiry)

  ---
  ✨ Suggestions

  1. Add integration tests - Use testcontainers-go to spin up real PostgreSQL for testing
  2. Connection pool metrics - Export pool stats (active connections, wait time) as Prometheus metrics
  3. Query logging - Add slow query logging for debugging production issues
  4. Batch operations - Implement DBStoreBatch() using COPY or batch inserts for better performance
  5. Upsert support - Add DBUpsert() using ON CONFLICT for idempotent operations
  6. Read replicas - Support read/write splitting for scalability
  7. Table name validation - Prevent SQL injection via malicious table descriptors:
  func validateTableName(name string) error {
      if !regexp.MustCompile(`^[a-z_][a-z0-9_]*$`).MatchString(name) {
          return fmt.Errorf("invalid table name: %s", name)
      }
      return nil
  }

  ---
  🎯 Verdict

  Overall: Good foundation, but needs critical fixes before merge ⚠️

  The architecture is solid, test coverage is comprehensive, and the code is well-organized. However, there are critical issues that must be addressed.

  Blockers (MUST fix before merge):
  1. Fix password escaping in connection string (#1)
  2. Add thread safety to Close() (#2)

  Strongly recommended (fix before or immediately after merge):
  3. Add connection pool configuration (#6)
  4. Add missing tests for connection string generation (#12)
  5. Document schema migration strategy (#7)

  Nice-to-have (future enhancements):
  - Composite indexes for common queries (#3)
  - Performance optimization for OR tag queries (#4)
  - Integration tests (#suggestions)
  - Prepared statement caching (#8)

  ---
  Action Items

  For the PR author:
  1. Fix connection string escaping immediately
  2. Add sync.Once to Close()
  3. Add tests for connection string with special characters in password
  4. Add concurrent Close() test
  5. Document schema migration expectations

  For reviewers:
  - Verify password escaping fix with unit tests
  - Test connection with passwords containing: ', \, space, @, =
  - Verify thread safety of Close() with race detector

  Would you like me to:
  1. Create a patch for the critical fixes?
  2. Write the missing tests?
  3. Draft an integration test using testcontainers-go?
  4. Review any specific part more deeply?

internal/database/postgresql/db_core.go
```diff
@@ -0,0 +1,462 @@
...
+		if err != nil {
+			return nil, nil, nil, 0, false, err
+		}
+
+		indexes = append(indexes, idx)
+		contents = append(contents, cnt)
+		extras = append(extras, extra)
+	}
+
+	if err := rows.Err(); err != nil {
```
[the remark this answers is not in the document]
acardace: Yes this is the correct way to do it according to the docs https://pkg.go.dev/github.com/jackc/pgx/v5#Rows.

internal/database/postgresql/batch_db.go
```diff
@@ -0,0 +1,125 @@
...
+//go:embed batch_schema.sql
+var batchSchemaSql string
+
+// batchDescriptor implements TableDescriptor for batch items.
+type batchDescriptor struct{}
+
+func (batchDescriptor) TableName() string      { return "batch_items" }
+func (batchDescriptor) Schema() string         { return batchSchemaSql }
+func (batchDescriptor) HasStatus() bool        { return true }
+func (batchDescriptor) ExtraColumns() []string { return nil }
```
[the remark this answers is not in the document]
acardace: It's an empty struct, there's no benefit in using a pointer receiver. Also is pretty common in Go to use value receivers for types like this.

internal/database/postgresql/batch_db.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return &PostgresBatchDBClient{pgCore}, nil
+}
+
+func (c *PostgresBatchDBClient) GetContext(parentCtx context.Context, timeLimit time.Duration) (context.Context, context.CancelFunc) {
+	return c.getContext(parentCtx, timeLimit)
+}
+
+func (c *PostgresBatchDBClient) Close() error {
+	return c.close()
```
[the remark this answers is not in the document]
acardace: Fixed.

internal/database/postgresql/db_core.go
```diff
@@ -0,0 +1,448 @@
...
+}
+
+// selectColumns returns the column list for a SELECT query.
+func (c *pgCore) selectColumns(includeStatic bool) []string {
+	cols := []string{colID, colTenantID, colExpiry, colTags}
+	cols = append(cols, c.desc.ExtraColumns()...)
+	cols = append(cols, colStatus)
+	if includeStatic {
+		cols = append(cols, colSpec)
+	}
```
lioraron: Maybe more efficient to call append once:
```
	if includeStatic {
		cols = append(cols, colStatus, colSpec)
	} else {
    	cols = append(cols, colStatus)
    }
```

lioraron (approved):
Thanks!
Add minor comments, and pls also ensure that all the comments from copilot are inspected.

PR #53, opened 2026-02-18
Refactor batch processor phase-1 pipeline and add test coverage

j-mok-dev:
**Summary**
Refactored processor internals around a phase-1 flow (poller/status-updater/planner split) to improve module boundaries and testability.
- Added architecture doc.
- Added phase-1 preprocessing behavior for local input persistence and per-model plan/index generation.
- Updated processor config/helm values for new runtime knobs (work_dir, max_open_files, readiness/shutdown behavior) while keeping processor disabled by default.
- Expanded unit tests across processor config, metrics, planner/poller, and phase-1 worker flow.

**Known limitations**
- Phase-2 execution path is not implemented yet (processJob path remains pending).
- Processor client wiring in cmd/batch-processor still uses mock clients for now.


internal/processor/worker/worker.go
```diff
@@ -147,238 +197,440 @@ func (p *Processor) RunPollingLoop(ctx context.Context) error {
...
-				return
-			default:
-			}
-			// TODO:: line parsing
-			// TODO:: check allowed methods
-			// TODO:: request validation
+		// parse the line (minimal check for the model id) for plan file writing
+		var req planRequestLine
+		trimmedLine := bytes.TrimSuffix(line, []byte{'\n'})
+		if err := json.Unmarshal(trimmedLine, &req); err != nil {
```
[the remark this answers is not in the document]
j-mok-dev: we write the raw input first to preserve the original file, then parse minimally for plan generation. for MVP, i chose to fail fast - if any line is invalid, the job fails. i think we can revisit partial processing later.

internal/processor/worker/fileio.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, nil, fmt.Errorf("failed to get input file metadata: %w", err)
+	}
+	if len(items) == 0 {
+		return nil, nil, fmt.Errorf("input file %q not found in db", inputFileID)
+	}
+
+	// unmarshal file object into openai file object to get inputfile filename and tenant id
+	fileItem := items[0]
+	fileObj := &openai.FileObject{}
+	if err := json.Unmarshal(fileItem.Spec, fileObj); err != nil {
```
yizhaodev: internal/shared/converter/file_converter.go has a function `DBItemToFile` to do this

internal/processor/worker/fileio.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil, nil, fmt.Errorf("failed to unmarshal file object: %w", err)
+	}
+
+	reader, metadata, err := p.clients.files.Retrieve(ctx, fileObj.Filename, fileItem.TenantID)
```
yizhaodev: TenantID can't be used as folder name directly
We may have to write a function to sanitize TenantID into valid folder name
please add a TODO before we wrote that function?

internal/processor/worker/worker.go
```diff
@@ -58,27 +72,56 @@ func NewProcessorClients(
...
 	clients *ProcessorClients
 }
 
+var ErrCancelled = errors.New("batch job cancelled")
+
 func NewProcessor(
 	cfg *config.ProcessorConfig,
 	clients *ProcessorClients,
 ) *Processor {
+	sem := make(chan struct{}, cfg.NumWorkers)
```
yizhaodev: use `golang.org/x/sync/semaphore`?
https://pkg.go.dev/golang.org/x/sync/semaphore

```
sem        = semaphore.NewWeighted(int64(maxWorkers))
sem.Acquire(ctx, 1)
sem.Release(1)
```

yizhaodev: hmm, semaphore is not standard library....

internal/processor/worker/worker.go
```diff
@@ -119,25 +167,27 @@ func (p *Processor) RunPollingLoop(ctx context.Context) error {
...
 		}
 
 		// check queue for available tasks
-		task := p.getTaskFromQueue(ctx)
+		task, err := poller.DequeueOne(ctx)
+
+		// polling error
+		if err != nil {
+			p.release()
+			continue
```
yizhaodev: so if failed to DequeueOne, 
`continue` will make the for loop to DequeueOne again without sleep for PollInterval?

internal/processor/worker/worker.go
```diff
@@ -147,236 +197,438 @@ func (p *Processor) RunPollingLoop(ctx context.Context) error {
...
+
+			p.release()
+			metrics.RecordJobProcessed(metrics.ResultSkipped, metrics.ReasonNotRunnableState)
+			continue
+		}
 
 		// process job
-		go func(wid int, j *db.BatchItem) {
+		p.wg.Add(1)
+		go func(c context.Context, jobItem *db.BatchItem, jobInfo *batch_types.JobInfo, task *db.BatchJobPriority) {
```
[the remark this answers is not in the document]
j-mok-dev: i’ve refactored the RunPollingLoop into specialized helper components (Poller, Planner, Preprocessor, JobRunner, etc.) across separate files. a significant portion of the diff is dedicated to new unit tests (*_test.go) for each decoupled module, which was previously difficult to test.

Terribly sorry for the monsterous size, and I truly appreciate your patience and thorough review! 🙏

internal/processor/worker/worker.go
```diff
@@ -58,325 +67,255 @@ func NewProcessorClients(
...
+}
 
-	// status update
-	p.clients.status.StatusSet(jobctx, job.ID, 24*60*60, []byte(batch.StatusFinalizing))
+func (p *Processor) release() {
+	select {
+	case p.tokens <- struct{}{}:
+		return
+	default:
+		panic("token channel is full")
```
yizhaodev: are we allowed to use panic to control flow ? @lioraron

PR #70, opened 2026-02-25
refactor: restructure ServerConfig and fix REST API issues

yizhaodev:
The PR fixed various issues in REST api

  Reorganize ServerConfig
  - Introduce BatchAPIConfig
  - Update Helm chart accordingly

  Fix Batch API handler issues:
  - Rename BatchApiHandler → BatchAPIHandler (Go naming convention)
  - Validate input file existence on CreateBatch
  - Store output_expires_after as database tags on batch creation
  - Use common.DefaultTenantID instead of hardcoded "unknown" in RequestMiddleware
  - Add validation that output_expires_after.anchor must be "created_at"

  Fix File API handler issues:
  - Rename FileApiHandler→ FileAPIHandler(Go naming convention)
  - Fix ListFiles to use query.Purpose field instead of tag selectors for purpose filtering

  Add unit tests


internal/apiserver/common/config.go
```diff
@@ -52,50 +52,64 @@ const (
...
+type BatchAPIConfig struct {
+	BatchEventTTLSeconds int `yaml:"batch_event_ttl_seconds"`
+}
+
+// GetBatchEventTTLSeconds returns the configured value or default if invalid
+func (b *BatchAPIConfig) GetBatchEventTTLSeconds() int {
+	if b.BatchEventTTLSeconds <= 0 {
+		return DefaultBatchEventTTLSeconds
+	}
+	return b.BatchEventTTLSeconds
```
lioraron: Is there a point in time where a config object is initialized from storage (configmap)?
If so, is it possible to check and set these defaults once at that time (instead of checking every time we access a param)?
I think that Get* methods are useful to enable dynamic config refresh, but not clear why we check for nil value on every access.

lioraron (approved):
Added questions.

PR #74, opened 2026-02-26
doc: add doc for resource lifecycle

yizhaodev:
This doc is to summarize the lifecycle of each resource.
It need both API server, processor and cleanup service to implement this together

@acardace @lioraron @j-mok-dev

### DIFF
--- a/docs/design/resource-lifecycle.md
+++ b/docs/design/resource-lifecycle.md
@@ -1,0 +1,54 @@
+# Resource Lifecycle
+
+This document describes how each type of object in the batch gateway is created and deleted,
+following the [OpenAI Batch API](https://platform.openai.com/docs/api-reference/batch) behavior.
+
+## Batch
+
+- **Creation:** Created when the user calls `POST /v1/batches`. Exists in database only, no physical files.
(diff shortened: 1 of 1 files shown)


docs/design/resource-lifecycle.md
```diff
@@ -0,0 +1,53 @@
+# Resource Lifecycle
+
+This document describes how each type of object in the batch gateway is created and deleted,
+following the [OpenAI Batch API](https://platform.openai.com/docs/api-reference/batch) behavior.
+
+## Batch
+
+- **Creation:** Created when the user calls `POST /v1/batches`. Exists in database only, no physical files.
+- **Deletion:** Never deleted — retained for audit and history. OpenAI does not expose a `DELETE /batches` endpoint either.
```
lioraron: Isn't there some type of ttl?
How can this be a boundless db of items?
The job cancel api doesn't have an option to delete the batch record?

yizhaodev: from openai doc
https://developers.openai.com/api/reference/resources/batches/methods/cancel
- there is delete API 
- the `cancel` API will mark batch as `cancelled `

PR #87, opened 2026-03-04
fix: use last tenant header value to prevent ext_authz header injection

pierDipi:
The external auth service may append response headers instead of overwriting them. A client can send a spoofed tenant header which gets concatenated with the auth-injected value. This takes the last value from multi-value tenant headers, which is always the trusted auth-injected one per Envoy ext_authz ordering guarantees.

Adds tests for tenant header extraction including the multi-value workaround, request ID handling, and metrics/health path skipping.


pierDipi:
@yizhaodev can we merge?

PR #89, opened 2026-03-04
feat: support pass-through headers from batch creation to inference r…

yizhaodev:
- Add new conf to api server, to configure pass-through header names
```
pass_through_headers:
   - "X-MaaS-Username"
   - "X-MaaS-Group"
```
-  api server will extract those header values from user request, and store in DB
- processor will send inference requests with those headers
- E2E test is updated to cover this.


j-mok-dev (approved):
I think it's all good. I'll include an unit test checking if the header is given and passed to inference client on worker side in next pr.

PR #92, opened 2026-03-05
fix: use last header value for pass-through headers to prevent spoofing

pierDipi:
Switch from Header.Get (first value) to Header.Values with last-entry selection so that Envoy ext_authz-injected headers take precedence over client-supplied ones. Add tests covering single value, multi-value last-wins, absent header, and multiple configured headers.

Similar to https://github.com/llm-d-incubation/batch-gateway/pull/87


yizhaodev (approved):
LGTM, Thanks!

PR #97, opened 2026-03-06
feat: add multi-gateway inference endpoint support

j-mok-dev:
## Summary

Adds per-model inference gateway routing with optional per-gateway API keys.
When a batch contains requests for multiple models served on different endpoints,
the processor can now route each request to the correct gateway instead of
sending everything to a single URL.

## How it works

**Configuration** uses a default + override pattern:

```yaml
inference_config:
  gateway_url: "http://gateway-a:8000"       # default for all models
  model_gateways:                             # optional per-model overrides
    "llama-3":
      url: "http://gateway-a:8000"
    "mistral":
      url: "http://gateway-b:8000"
      api_key_name: "gateway-b-api-key"       # key name in /etc/.secrets/
```

### DIFF
--- a/charts/batch-gateway/values.yaml
+++ b/charts/batch-gateway/values.yaml
@@ -13,7 +13,7 @@ global:
   appSecretName: "batch-gateway-secrets"
 
   # Database backend type: "redis", or "postgresql"
-  databaseType: "redis"
+  databaseType: "postgresql"
 
   # OpenTelemetry configuration
   otel:
--- a/cmd/batch-processor/main.go
+++ b/cmd/batch-processor/main.go
@@ -295,7 +295,7 @@ func buildProcessorClients(ctx context.Context, cfg *config.ProcessorConfig) (*c
 	clients, err := clientset.NewClientset(
 		ctx,
 		cfg.DatabaseType,
-		nil,
+		&cfg.PostgreSQLCfg,
 		redisCfg,
 		cfg.FileClientCfg.Type,
 		&cfg.FileClientCfg.FSConfig,
--- a/internal/apiserver/common/config.go
+++ b/internal/apiserver/common/config.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/llm-d-incubation/batch-gateway/internal/database/postgresql"
 	fsclient "github.com/llm-d-incubation/batch-gateway/internal/files_store/fs"
 	s3client "github.com/llm-d-incubation/batch-gateway/internal/files_store/s3"
 	"gopkg.in/yaml.v3"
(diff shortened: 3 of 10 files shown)


charts/batch-gateway/values.yaml
```diff
@@ -263,7 +263,16 @@ processor:
...
+      gatewayUrl: "http://localhost:8000"  # default for all models
+      # Per-model overrides. Lookup: modelGateways[model] -> gatewayUrl (fallback).
+      # apiKeyName is a key name in the app secret (mounted at /etc/.secrets/).
+      # When omitted, the default inference-api-key is used.
+      # modelGateways:
+      #   "llama-3":
+      #     url: "http://gateway-a:8000"
+      #   "mistral":
+      #     url: "http://gateway-b:8000"
+      #     apiKeyName: "gateway-b-api-key"
```
lioraron: Maybe add in the comments that a single gateway override per model is supported.

internal/inference/gateway_resolver.go
```diff
@@ -0,0 +1,132 @@
...
+	defaultClient, err := NewHTTPClient(baseCfg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create default inference client: %w", err)
+	}
+
+	// Pool of clients for reuse. The key is the URL + API key.
+	// This covers the case where multiple models share the same URL but different API keys.
+	pool := map[clientKey]Client{
+		{baseCfg.BaseURL, baseCfg.APIKey}: defaultClient,
+	}
```
lioraron: Recommendation from claude:

Yes! For 2 strings specifically, here are the best alternatives:

1. Array Key [2]string (Recommended)

// Instead of:
type Key struct {
    UserID   string
    TenantID string
}
m := make(map[Key]Value)
m[Key{userID, tenantID}] = value

// Use:
m := make(map[[2]string]Value)
m[[2]string{userID, tenantID}] = value

Pros:
10-20% faster than struct (simpler memory layout, faster hashing)
No allocation overhead
Still type-safe and comparable

Cons: 
Less readable field access (key[0] vs key.UserID)

internal/inference/gateway_resolver.go
```diff
@@ -0,0 +1,132 @@
...
+type GatewayResolver struct {
+	defaultClient Client
+	modelClients  map[string]Client
+}
+
+// clientKey is a deduplication key for the HTTP client pool.
+// Gateways sharing both the same URL and API key reuse a single client.
+type clientKey struct {
+	url    string
+	apiKey string
```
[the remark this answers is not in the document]
j-mok-dev: good catch. clientKey was redundant since GatewayClientConfig can be used as a map key directly. Removed clientKey and the conversion helper. Updated.

PR #110, opened 2026-03-09
feat: add OTel instrument for postresql

yizhaodev:
The PR will add OTel instrument for postresql operations

- Each DB span is added to corresponding parent span
- Each DB span has its own detail tags.

image

image


j-mok-dev (approved):
lgtm

PR #112, opened 2026-03-09
feat: SLO enforcement

j-mok-dev:
## Summary

Implements SLO enforcement for batch jobs based on the `expires_at` field set from `completion_window`.

- **`context.WithDeadline`** on the job execution context (`sloCtx`) to bound Phase 2 dispatch
- When the deadline fires, undispatched requests are drained to `error.jsonl` as `batch_expired`; completed requests are preserved in `output.jsonl` — per [OpenAI batch expiration spec](https://platform.openai.com/docs/guides/batch#batch-expiration)
- Job transitions directly `in_progress → expired` (no `finalizing` state); partial output and error files are uploaded before status update
- Two new reason labels: `expired_dequeue` (SLO exceeded at dequeue time) and `expired_execution` (SLO fired mid-Phase 2)

## Key Design Decisions

- SLO check uses `sloCtx.Err()` (not `execCtx.Err()`) to avoid a race where `execCancel()` from a model goroutine sets `execCtx.Err = Canceled` before the deadline propagates
- File uploads in `handleExpired` use parent `ctx` (not `sloCtx`), so uploads succeed even after the SLO fires
- `handleExpired` is best-effort for uploads — failures are logged but do not block the `expired` status update

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -52,6 +52,10 @@ import (
 // "batch_expired") are preserved. The caller is responsible for finalizing with expired status.
 var ErrExpired = errors.New("batch SLO expired")
 
+// errCodeBatchExpired is the error code written to the error file for requests that could not
+// be executed before the job's completion window expired, per the OpenAI Batch API spec.
+const errCodeBatchExpired = "batch_expired"
+
 // outputWriters holds the buffered writers and their mutexes for the output and error JSONL files.
 // A single instance is created per job and shared across model goroutines.
 type outputWriters struct {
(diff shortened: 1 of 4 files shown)


internal/database/api/database.go
```diff
@@ -114,10 +114,15 @@ type BatchPriorityQueueClient interface {
...
-	// PQDequeue returns the job priority objects at the head of the queue,
+	// PQDequeue atomically removes and returns the job priority objects at the head of the queue,
 	// up to the maximum number of objects specified in maxItems.
 	// The function blocks up to the timeout value for a job priority object to be available.
 	// If the timeout value is zero, the function returns immediately.
+	//
+	// Implementations MUST atomically remove dequeued items from the queue. The processor
+	// assumes exclusive dequeue semantics: a dequeued job will not be returned by any
+	// subsequent PQDequeue call. Non-atomic (peek/lease) implementations will cause the
+	// same job to be processed multiple times.
```
j-mok-dev: added some comments to clarify PQDequeue atomicity contract to prevent duplicate job processing in worker

internal/processor/worker/executor.go
```diff
@@ -287,13 +313,14 @@ func (p *Processor) processModel(
...
+		wg              sync.WaitGroup
+		errOnce         sync.Once
+		firstErr        error
+		dispatchedCount int
 	)
 
 dispatch:
-	for _, entry := range entries {
+	for i, entry := range entries {
 		if err := checkCancellation(ctx, cancelRequested); err != nil {
```
yizhaodev: do you need to run `sloCtx.Err() == context.DeadlineExceeded` for early fail check?

PR #121, opened 2026-03-11
feat: preserve partial output on cancelled and failed batch jobs

j-mok-dev:
**Summary**
Preserve partial inference output when batch jobs are cancelled or fail mid-execution (Phase 2). Previously, only expired jobs retained partial results — cancelled and failed jobs discarded all output.

- Refactor `drainUnprocessedRequests` and `uploadPartialResults` as reusable helpers across all terminal paths
- Flush writers and return partial counts on cancel and system error paths in `executeJob`
- Add `handleFailedWithPartial` for Phase 2 failures; extend `handleCancelled` to upload partial output
- Add `UpdateFailedStatus`, `UpdateCancelledStatus` to `StatusUpdater` with shared `withFileIDs` helper
- Introduce `batch_cancelled` and `batch_failed` error codes (extensions to the OpenAI Batch API spec)
- Record OTel request count attributes on all terminal states after Phase 2
- Rename `firstErr` to `modelErr` in `processModel` to disambiguate from executeJob-level `firstErr`
- Update design doc: Mermaid lifecycle diagram, partial output preservation section, observability table

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -415,6 +415,7 @@ dispatch:
 	// Priority: SLO expiry > user cancel > system error.
 	// Use sloCtx.Err() rather than ctx.Err(): ctx (execCtx) may report Canceled if execCancel()
 	// was called by another goroutine before the sloCtx deadline propagated.
+	// (Same rationale as the sloCtx check in executeJob.)
 	switch {
 	case sloCtx.Err() == context.DeadlineExceeded && !cancelRequested.Load():
 		// SLO deadline fired during dispatch — record remaining requests as expired.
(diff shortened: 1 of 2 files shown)


internal/processor/worker/executor.go
```diff
@@ -382,41 +390,69 @@ dispatch:
...
+	switch {
+	case sloCtx.Err() == context.DeadlineExceeded && !cancelRequested.Load():
+		// SLO deadline fired during dispatch — record remaining requests as expired.
 		undispatched := entries[dispatchedCount:]
 		if len(undispatched) > 0 {
 			logger.V(logging.INFO).Info("SLO expired: draining undispatched entries", "count", len(undispatched))
-			p.drainUndispatchedAsExpired(ctx, inputFile, undispatched, writers, progress)
+			p.drainUnprocessedRequests(ctx, inputFile, undispatched, writers, progress,
+				batch_types.ErrCodeBatchExpired,
+				"This request could not be executed before the completion window expired.")
```
lioraron: "This request": Re terminology we have `batch` representing the job and `request` representing individual inference request. Just to confirm that this message and the following similar messages refer to individual inferences?

PR #127, opened 2026-03-12
feat: abort in-flight inference requests on batch cancellation

j-mok-dev:
## Summary
When a batch cancellation event arrives, in-flight inference HTTP requests are now aborted immediately by cancelling the inference context (`inferCtx`). Previously, cancellation only prevented new requests from being dispatched / already in-flight requests ran to completion, wasting GPU slots and EPP capacity.

### Changes
- **`cancel.go`**: `watchCancel` accepts `inferCancelFn` and calls it on cancel event, propagating cancellation to all in-flight HTTP requests
- **`job_runner.go`**: Creates `inferCtx` (derived from `sloCtx`) and passes it to `watchCancel` and `executeJob`
- **`executor.go`**: `execCtx` is now derived from `inferCtx` instead of `sloCtx` directly, so user cancellation propagates to all dispatch loops and inference calls
- **`batches_test.go`**: E2E cancel test verifies `"Request cancelled for request_id"` appears in processor logs, confirming in-flight HTTP requests were actually aborted

### Context hierarchy
```
ctx (parent)
 │
 │ ── sloCtx = context.WithDeadline(ctx, SLO)
 │    │
 │    └── inferCtx = context.WithCancel(sloCtx)  ← cancelled on user cancellation (cancel event)
 │         │
 │         └── execCtx = context.WithCancel(inferCtx)
 │              │
 │              └── inferClient.Generate(execCtx, ...) ← stops HTTP request
 │
 └── watchCancel goroutine → inferCancelFn() → inferCtx cancel
```

### Follow-up
- Refactor `checkAbortCondition`: now that `inferCtx` propagates cancellation through the context tree, the `cancelRequested` polling in `checkAbortCondition` is partially redundant for stopping dispatch. Refactor so that context cancellation is the sole dispatch-abort mechanism and `cancelRequested` is only used to distinguish the cancellation reason (user cancel vs SLO vs pod shutdown).

### DIFF
--- a/test/e2e/batches_test.go
+++ b/test/e2e/batches_test.go
@@ -73,6 +73,13 @@ func doTestBatchCancel(t *testing.T) {
 		finalBatch.RequestCounts.Failed,
 		finalBatch.RequestCounts.Total)
 
+	if finalBatch.RequestCounts.Total != int64(len(strings.Split(strings.TrimSpace(slowJSONL), "\n"))) {
+		t.Errorf("Total = %d, want %d", finalBatch.RequestCounts.Total, len(strings.Split(strings.TrimSpace(slowJSONL), "\n")))
+	}
+	if finalBatch.RequestCounts.Completed+finalBatch.RequestCounts.Failed != finalBatch.RequestCounts.Total {
+		t.Errorf("Completed(%d) + Failed(%d) != Total(%d)",
+			finalBatch.RequestCounts.Completed, finalBatch.RequestCounts.Failed, finalBatch.RequestCounts.Total)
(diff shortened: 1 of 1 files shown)


j-mok-dev:
@yizhaodev exact count validation is done in unit tests where we can control timing with mocks. see `TestExecuteJob_InferCtxCancel_AbortsInflightRequests` which asserts Total=1, Completed=0, Failed=1. In E2E, timing is non-deterministic so we only check consistency: Total == N, Completed + Failed == Total, Completed < Total.

PR #162, opened 2026-03-18
fix: cancel race condition fix

### LINKED ISSUE #146: Flaky E2E test: CancelBatch/InProgress — race between processor completion and cancel event
## Summary

The `TestE2E/Batches/CancelBatch/InProgress` E2E test is flaky — it passes ~50% of the time. The root cause is a race condition between the processor finishing execution and the cancel event being delivered.

## Symptoms

- After calling `POST /v1/batches/{id}/cancel`, the batch transitions to `cancelling` but never reaches `cancelled`.
- The test times out waiting for the batch to reach a terminal state (`completed` or `cancelled`).
- The batch remains stuck in `cancelling` forever.

## Root Cause

When a batch has very few requests (as in the E2E test), the processor can finish executing all requests **before** the cancel event arrives via Redis pub/sub. The sequence is:

1. API server sets batch status to `cancelling` and publishes a cancel event.
2. Processor's `watchCancel` goroutine is listening for the event, but the processor has already finished all inference requests.
3. Processor enters finalization: it closes the `watchCancel` channel and proceeds to write `completed` status.
4. The cancel event arrives but has no consumer — it's dropped.
5. The batch is finalized as `completed`, but the DB status is already `cancelling`, so the status update to `completed` is rejected (or succeeds but the apiserver already sees `cancelling`).
6. Result: batch stuck in `cancelling` with no mechanism to transition it to `cancelled`.

## Suggested Fix

In the processor's **finalization path** (after all requests are processed), before writing the terminal status:

1. Re-read the batch status from the database.
2. If the status is `cancelling`, finalize the batch as `cancelled` instead of `completed`.

This makes the processor's finalization idempotent with respect to concurrent cancellation, regardless of whether the cancel event was received in time.

## Reproduction

```bash
# Run multiple times — fails intermittently
go test -v -tags=e2e -count=1 -run "^TestE2E/Batches/CancelBatch/InProgress$" ./test/e2e/...
```

## Labels

flaky-test, race-condition, processor

j-mok-dev:
## Why is this PR needed?

Issue #146 

## What does this PR do?
- tighten processor readiness so `/ready` is exposed only after processor preparation completes, and update dev deployment scripts to check processor readiness instead of health
- stabilize repeated E2E runs by routing non-lifecycle test file creation through a unique-filename helper, avoiding `file already exists` failures
- reduce the cancel/finalization race by persisting `cancelling` before publishing the cancel event from the API server
- during finalization, honor an already-observed cancel request after artifact upload so we do not overwrite cancellation with a later terminal status

## Notes

- cancellation remains best-effort
- if a cancel request arrives after execution and artifact upload have already completed, the batch may still finalize as `completed`

## Test plan

- `pre-commit run -a`
- `go test ./...`
- `make test-e2e`
- run `make test-e2e` 15 times in a row against the current deployment## How was this tested?

## Checklist

## Related Issues
Fixes #146

### DIFF
--- a/scripts/dev-deploy.sh
+++ b/scripts/dev-deploy.sh
@@ -661,7 +661,7 @@ wait_for_deployment() {
 start_apiserver_port_forward() {
     local svc="svc/${HELM_RELEASE}-apiserver"
     local max_retries=3
-    local health_check_attempts=30
+    local readiness_check_attempts=30
 
     kill_ports "${LOCAL_PORT}" "${LOCAL_OBS_PORT}"
 
@@ -675,7 +675,7 @@ start_apiserver_port_forward() {
         log "Waiting for http://localhost:${LOCAL_OBS_PORT}/health ..."
 
         local success=false
-        for i in $(seq 1 ${health_check_attempts}); do
+        for i in $(seq 1 ${readiness_check_attempts}); do
             if curl -sf "http://localhost:${LOCAL_OBS_PORT}/health" >/dev/null 2>&1; then
                 log "API server is ready at https://localhost:${LOCAL_PORT}"
                 success=true
@@ -688,8 +688,8 @@ start_apiserver_port_forward() {
             return 0
         fi
 
-        # Health check failed - kill the port-forward and retry
-        warn "Port-forward health check failed on attempt ${retry}/${max_retries}"
+        # Readiness check failed - kill the port-forward and retry
+        warn "Port-forward readiness check failed on attempt ${retry}/${max_retries}"
         kill "${pf_pid}" 2>/dev/null || true
         kill_ports "${LOCAL_PORT}" "${LOCAL_OBS_PORT}"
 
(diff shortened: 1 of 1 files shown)


scripts/dev-deploy.sh
```diff
@@ -716,11 +716,11 @@ start_processor_port_forward() {
         disown "${pf_pid}"
         log "Processor port-forward PID: ${pf_pid}  (stop with: kill ${pf_pid})"
 
-        log "Waiting for http://localhost:${LOCAL_PROCESSOR_PORT}/health ..."
+        log "Waiting for http://localhost:${LOCAL_PROCESSOR_PORT}/ready ..."
 
         local success=false
         for i in $(seq 1 ${health_check_attempts}); do
```
yizhaodev: should reanme to ready_check_attempts then

yizhaodev (approved):

PR #231, opened 2026-03-26
chore: use gcr mirror for local development

yizhaodev:
## Why is this PR needed?

We got too many rate limit with `docker.io`, I am changing the repo to google mirror

## What does this PR do?

Update `scripts/dev-deploy.sh` to use image from google mirror.

## How was this tested?

## Checklist


j-mok-dev:
One question on scope: if the goal is to reduce Docker Hub rate-limit failures for fresh `make dev-deploy`, do we also need to look at the earlier pulls in the flow?
We still install Redis/PostgreSQL via Bitnami charts first, and `MINIO_IMAGE` still defaults to `minio/minio:latest`, so a fresh setup may still hit Docker Hub before reaching the mirrored Jaeger/Prometheus/Grafana images.
If the rate-limit issue was specifically observed on these three images, then this change makes sense as-is.

PR #253, opened 2026-03-27
fix: add panic recovery guards to child goroutines in job processing

### LINKED ISSUE #242: Child goroutine panics crash the process — no recovery in watchCancel or per-request goroutines
## Problem

Several goroutines spawned during job processing have no panic recovery. A panic in any of them will crash the entire batch-gateway process, taking down all in-flight jobs — not just the affected one.

The code explicitly acknowledges this limitation at `internal/processor/worker/job_runner.go` lines 71-73:

```go
// Note: recover() only catches panics on this goroutine. Panics in child
// goroutines (watchCancel, per-model/per-request goroutines in executeJob)
// are not caught here and will crash the process.
```

### Affected goroutines

1. **`watchCancel`** — spawned at `job_runner.go:133` via `go p.watchCancel(ctx, params)`. Monitors Redis event channel for cancel requests.

2. **Per-model goroutines** — spawned at `executor.go:224` inside `executeJob`. Each model in `modelMap.SafeToModel` gets its own goroutine calling `processModel`.

3. **Per-request goroutines** — spawned at `executor.go:369` inside `processModel`. Each individual inference request runs in its own goroutine calling `executeOneRequest`.

### Impact

- A single malformed inference response causing a nil-pointer dereference in `executeOneRequest` would crash the entire process.
- All other in-flight jobs are lost with no graceful cleanup.
- The `runJob` panic handler (which correctly does DB cleanup) never fires for child goroutine panics.

## Suggested fix

Add `defer func() { if r := recover(); r != nil { ... } }()` guards to each of the affected goroutine entry points:

- **`watchCancel`**: Log the panic and set `cancelRequested` to trigger graceful shutdown of the parent job.
- **Per-model goroutines in `executeJob`**: Recover, send an error on `errCh`, and let the existing error-collection logic handle it.
- **Per-request goroutines in `processModel`**: Recover, write an error response to the error file, update atomic counters, and release the semaphore.

This keeps the blast radius of a panic to a single request/model/job rather than the entire process.

## References

- `internal/processor/worker/job_runner.go:71-73` (acknowledgment comment)
- `internal/processor/worker/job_runner.go:133` (`watchCancel` spawn)
- `internal/processor/worker/executor.go:224` (per-model spawn)
- `internal/processor/worker/executor.go:369` (per-request spawn)

lioraron:
## Why is this PR needed?

Several goroutines spawned during job processing (`watchCancel`, per-model, per-request) have no panic recovery. A panic in any of them crashes the entire batch-gateway process, taking down all in-flight jobs — not just the affected one. The code explicitly acknowledged this limitation in a comment at `job_runner.go:78-80`.

## What does this PR do?

Adds `defer func() { recover() }()` guards to each child goroutine entry point:

- **`watchCancel`** (`cancel.go`): Recovers, logs with stack trace, sets `cancelRequested` and calls `abortInferFn()` to trigger graceful shutdown of the parent job.
- **Per-model goroutines** (`executor.go:224`): Recovers, logs, calls `execCancel()` to cancel sibling models, and sends the panic as an error on `errCh` for the existing error-collection logic.
- **Per-request goroutines** (`executor.go:369`): Recovers, logs, captures the error via `errOnce` for the parent model. Semaphore releases and `wg.Done()` are already deferred above the recovery guard.

This keeps the blast radius of a panic to a single request/model/job rather than the entire process.

## How was this tested?

- [x] Verified compilation and existing tests still pass
- [x] Code review — each recovery handler uses the same error propagation path as non-panic errors

## Checklist

## Related Issues

Fixes #242

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -77,23 +77,51 @@ type outputError struct {
 	Message string `json:"message"`
 }
 
+// progressUpdateInterval is the minimum time between Redis progress updates.
+// Updates within this window are skipped — the next update after the interval
+// will include all accumulated progress. Declared as var so tests can override.
+var progressUpdateInterval = time.Second
+
 // executionProgress tracks per-request progress across goroutines
-// and pushes lightweight updates to the status store after every request.
+// and pushes throttled updates to the status store.
 type executionProgress struct {
-	completed atomic.Int64
-	failed    atomic.Int64
-	total     int64
-	updater   *StatusUpdater
-	jobID     string
+	completed  atomic.Int64
+	failed     atomic.Int64
+	total      int64
+	updater    *StatusUpdater
+	jobID      string
+	lastUpdate atomic.Int64 // unix nanoseconds of last Redis push
 }
 
+// record increments the appropriate counter and pushes a throttled progress
+// update to Redis. Updates are skipped if less than progressUpdateInterval
+// has elapsed since the last push, reducing Redis writes from O(requests)
+// to O(job_duration / interval).
 func (ep *executionProgress) record(ctx context.Context, success bool) {
 	if success {
 		ep.completed.Add(1)
 	} else {
 		ep.failed.Add(1)
 	}
-	// best-effort: status store failure should not block request processing
+	now := time.Now().UnixNano()
+	last := ep.lastUpdate.Load()
+	if now-last < int64(progressUpdateInterval) {
+		return
+	}
+	// Best-effort CAS: if another goroutine raced us, skip this update.
+	if !ep.lastUpdate.CompareAndSwap(last, now) {
+		return
+	}
+	ep.push(ctx)
+}
+
+// flush pushes the final progress to Redis unconditionally, ensuring the
+// last update reflects the true counts regardless of throttling.
+func (ep *executionProgress) flush(ctx context.Context) {
+	ep.push(ctx)
+}
+
+func (ep *executionProgress) push(ctx context.Context) {
 	if err := ep.updater.UpdateProgressCounts(ctx, ep.jobID, &openai.BatchRequestCounts{
 		Total:     ep.total,
 		Completed: ep.completed.Load(),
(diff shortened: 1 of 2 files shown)


internal/processor/worker/job_runner.go
```diff
@@ -75,9 +75,10 @@ func (p *Processor) runJob(ctx context.Context, params *jobExecutionParams) {
 		transitionedToInProgress bool
 		requestCounts            *openai.BatchRequestCounts
 	)
-	// Note: recover() only catches panics on this goroutine. Panics in child
-	// goroutines (watchCancel, per-model/per-request goroutines in executeJob)
-	// are not caught here and will crash the process.
+	// Note: recover() only catches panics on this goroutine. Child goroutines
```
j-mok-dev: This was an intentional design decision. initially we didn't put the panic recovery in child goroutines because a panic in a child goroutine means process state is untrustworthy, and letting Kubernetes restart the pod is safer than continuing in a potentially corrupted state.

If we want to revisit this, it should be a deliberate design discussion rather than a mechanical fix. The tradeoff is:
- **Current (crash)**: blast radius is all in-flight jobs on this pod, but state is guaranteed clean after restart
- **This PR (recover)**: blast radius is one job, but the process continues with potentially corrupted shared state (semaphores, writers, cancel channels).

lioraron:
After reviewing j-mok-dev's feedback, I agree with the assessment. Here's the reasoning:

**The crash-on-panic behavior is the safer default.** When a child goroutine panics mid-operation, it can leave shared state half-mutated — `globalSem` tokens unreleased (silently reducing concurrency), `outputWriters` with partially-written JSONL lines, cancel channels that never fire leaving sibling goroutines hanging, or `WaitGroup` counters that never decrement causing the parent to block indefinitely.

A `recover()` in the child goroutine can *attempt* cleanup, but it's best-effort — the panic could have occurred between acquiring a resource and deferring its release, so the recovery handler can't reliably know what to clean up.

**The pod-crash approach has a built-in safety net.** Kubernetes restarts the pod with clean state, and `recovery.go` already handles stale `in_progress` jobs on startup. The blast radius (all in-flight jobs) is wider, but the recovery is complete and deterministic.

**Panics should be rare.** They indicate bugs, not expected failures. The right response is to fix the bug that caused the panic rather than building elaborate recovery machinery around it.

Closing this PR per the reviewer's recommendation. If a specific child goroutine is later identified as panic-prone (e.g. due to third-party library calls), that one goroutine could get targeted recovery with carefully scoped cleanup — but not as a blanket pattern.

PR #261, opened 2026-03-29
fix: preserve HTTP status code and response body in batch output

yizhaodev:
## Why is this PR needed?

The batch processor currently treats all inference errors the same way — populating the `error` field and writing to `error.jsonl`. However, the [OpenAI batch spec](https://platform.openai.com/docs/api-reference/batch) distinguishes between HTTP error responses (4xx/5xx) and non-HTTP errors (network failures, timeouts). HTTP errors should carry the original status code and response body in the `response` field and be written to `output.jsonl`, not `error.jsonl`.

## What does this PR do?

- **Preserves HTTP status code and response body**: `ClientError` now carries `StatusCode` and `ResponseBody` fields, populated by `HandleErrorResponse`.
- **Routes HTTP errors to output file**: HTTP error responses (4xx/5xx) populate the `response` field with `status_code` and original `body`, and are written to `output.jsonl`. Non-HTTP errors (network, timeout) continue to use the `error` field and go to `error.jsonl`.
- **Adds `OpenAIErrorType()` helper**: Maps internal error categories to OpenAI-style error type strings (e.g., `invalid_request_error`, `server_error`).
- **Handles edge cases**: Empty response body produces an empty JSON object `{}`; non-JSON response body (e.g., HTML) is wrapped in a synthetic error structure.

## How was this tested?

## Checklist

## Related Issues
https://github.com/llm-d-incubation/batch-gateway/issues/205

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -76,6 +76,18 @@ type outputError struct {
 	Message string `json:"message"`
 }
 
+// isSuccess returns true when the output line represents a fully successful request
+// (no non-HTTP error and a 200 HTTP status). HTTP error responses (4xx/5xx) are not
+// considered successful even though they populate the Response field.
+//
+// NOTE: because HTTP errors are written to the output file (not the error file),
+// request_counts.failed may be greater than the number of lines in the error file.
+// This diverges from OpenAI's documented behavior but aligns with the OpenAPI schema
+// (see executeOneRequest for rationale).
+func (o *outputLine) isSuccess() bool {
+	return o.Error == nil && o.Response != nil && o.Response.StatusCode == 200
+}
+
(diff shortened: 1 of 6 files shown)


j-mok-dev (approved):
The OpenAPI schema's error field description ("for requests that failed with a non-HTTP error") supports routing HTTP errors to the output file. But the guides describe output_file_id as "successfully executed requests", and I couldn't find a public example of OpenAI placing a non-200 status_code in the output file. Might be worth adding a short comment in executeOneRequest noting this is our interpretation of the schema.

Also, with this change request_counts.failed now includes HTTP errors that go to the output file (not the error file). So failed count will be greater than the number of lines in the error file. This might confuse users.. a comment on isSuccess() noting this divergence would help future readers.

PR #295, opened 2026-04-01
chore: increase timeout in demo script for slow env

yizhaodev:
## Why is this PR needed?

Some env are slow, need more time for deployment etc to become ready

## What does this PR do?

Increase the timeout waiting for deployment resoureces etc.

## How was this tested?

## Checklist

- [ ] Pre-commit checks pass (`make pre-commit`)
- [ ] Unit tests pass (`make test`)

### DIFF
--- a/examples/deploy-demo/common.sh
+++ b/examples/deploy-demo/common.sh
@@ -64,17 +64,18 @@ gen_id() {
 wait_for_deployment() {
     local deploy_name="$1"
     local namespace="$2"
-    local timeout="${3:-120s}"
+    local timeout="${3:-180s}"
 
     step "Waiting for deployment '${deploy_name}' to be ready..."
 
     local retries=0
+    local max_retries=30
     while ! kubectl get deploy "${deploy_name}" -n "${namespace}" &>/dev/null; do
         retries=$((retries + 1))
-        if [ "$retries" -ge 5 ]; then
-            die "Deployment '${deploy_name}' did not become ready"
+        if [ "$retries" -ge "$max_retries" ]; then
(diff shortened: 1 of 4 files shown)


lioraron (commented):
## Review Summary

The intent of this PR is clear and reasonable: increase timeouts in demo scripts to accommodate slow environments. However, there is one bug that should be fixed before merging.

### Bug: `wait_for_deployment` timeout parameter silently ignored

The most significant issue is that `wait_for_deployment()` in `common.sh` now **hardcodes** `--timeout=180s` for `kubectl rollout status`, making the 3rd parameter (`$timeout`) a dead parameter. This is a regression because several callers pass `300s` (e.g., for Istio gateway, Kuadrant, and llm-d deployments in `deploy-k8s.sh`). After this PR, those deployments will fail at 180s instead of the intended 300s.

**Fix:** Change the default from `120s` to `180s` but keep using the parameter:
```bash
local timeout="${3:-180s}"
...
kubectl rollout status deploy/"${deploy_name}" -n "${namespace}" --timeout="${timeout}"
```

### Minor: Visibility-wait loop is disproportionate

The retry count for the deployment visibility check went from 5 (10s) to 90 (180s). This loop only checks if the Deployment API object exists -- waiting 3 minutes for an API object to appear seems excessive. A value of 15-30 retries (30-60s) would be more appropriate, with the actual rollout wait handled by `kubectl rollout status`.

### Overall

The changes to `deploy-k8s.sh` and `deploy-rhoai.sh` (bumping explicit timeout values from 120s to 180s) are fine. The `common.sh` changes need a fix to preserve the timeout parameter.

lioraron (approved):

j-mok-dev (approved):
lgtm!

PR #297, opened 2026-04-01
feature: skip inference request early if SLO deadline expired

### LINKED ISSUE #296: Skip inference request when SLO deadline has already passed
In executeOneRequest, we should check sloCtx.Err() before calling inferClient.Generate(). If the SLO deadline has already passed, skip the request and return a timeout error directly.

Currently (after PR #256), expired deadlines are clamped to sloMs = 0 and the request is still sent with x-slo-ttft-ms: 0, which makes it the most urgent — stealing priority from requests that still have time left.

Ref: https://github.com/llm-d-incubation/batch-gateway/pull/256#discussion_r3023411238

loicmarchal:
## Why is this PR needed?

Return early from the request processing if SLO deadline is exceeded.

## What does this PR do?

Return an error result as soon as the deadline on sloCtx is exceeded. Also avoid adding SLO headers to the request if the deadline is exceeded.

## How was this tested?

## Checklist

## Related Issues

Fixes #296

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -684,8 +684,7 @@ func (p *Processor) executeOneRequest(
 		Headers:   headers,
 	}
 
-	switch sloCtx.Err() {
-	case context.DeadlineExceeded:
+	if sloCtx.Err() == context.DeadlineExceeded {
 		logger.V(logging.INFO).Info("SLO expired during execution, skipping request", "error", sloCtx.Err())
 		result := &outputLine{
 			ID:       newBatchRequestID(requestID),
@@ -693,18 +692,6 @@ func (p *Processor) executeOneRequest(
 			Error: &outputError{
 				Code:    batch_types.ErrCodeBatchExpired,
 				Message: "This request could not be executed before the completion window expired.",
-			},
-		}
-		metrics.RecordRequestError(modelID)
-		return result, nil
-	case context.Canceled:
-		logger.V(logging.INFO).Info("Execution context cancelled, skipping request", "error", sloCtx.Err())
-		result := &outputLine{
-			ID:       newBatchRequestID(requestID),
-			CustomID: req.CustomID,
-			Error: &outputError{
-				Code:    batch_types.ErrCodeBatchCancelled,
-				Message: "This request was not executed because the batch was cancelled.",
 			},
 		}
 		metrics.RecordRequestError(modelID)
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -591,11 +591,11 @@ func (p *Processor) drainUnprocessedRequests(
 
 const sloTTFTMSHeader = "x-slo-ttft-ms"
 
-// mergeSLOTTFTIntoHeaders sets sloTTFTMSHeader to the remaining time until
-// sloCtx's deadline in whole milliseconds, clamped to >= 0. If sloCtx has no
-// deadline or is cancelled, the headers map is returned unchanged.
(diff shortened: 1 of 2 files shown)


internal/processor/worker/executor.go
```diff
@@ -681,6 +683,20 @@ func (p *Processor) executeOneRequest(
 		Headers:   headers,
 	}
 
+	if sloCtx.Err() != nil {
+		logger.V(logging.INFO).Info("SLO expired during execution or execution context cancelled, skipping request", "error", sloCtx.Err())
+		result := &outputLine{
+			ID:       newBatchRequestID(requestID),
+			CustomID: req.CustomID,
+			Error: &outputError{
+				Code:    string(httpclient.ErrCategoryServer),
```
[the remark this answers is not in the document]
yizhaodev: We should define types for the code

yizhaodev (approved):

internal/processor/worker/executor.go
```diff
@@ -681,6 +683,20 @@ func (p *Processor) executeOneRequest(
 		Headers:   headers,
 	}
 
+	if sloCtx.Err() != nil {
```
lioraron: Using `sloCtx.Err() != nil` here catches both `context.DeadlineExceeded` (SLO expiry) and `context.Canceled`. However, the error response always uses `ErrCodeBatchExpired` with the message "This request could not be executed before the completion window expired." This is misleading when the context was cancelled rather than expired.

The rest of the codebase carefully distinguishes these two cases. For example, `processModel` (line ~491) checks `sloCtx.Err() == context.DeadlineExceeded && !cancelRequested.Load()` separately from `cancelRequested.Load()`, and uses different error codes (`batch_expired` vs `batch_cancelled`).

Consider either:
1. Only checking `sloCtx.Err() == context.DeadlineExceeded` here (consistent with the existing pattern), or
2. Adding a branch that distinguishes cancellation from deadline expiry.

Note: the caller in `processModel` (lines 431-456) does overwrite the result to `batch_cancelled` when `cancelRequested.Load()` is true, so the practical impact may be limited. But if `sloCtx` is cancelled for a reason other than user-initiated cancel (e.g., parent context cancellation from pod shutdown), the error code would be wrong — and the caller would treat it as a successful result (`execErr == nil`) rather than propagating the cancellation.

loicmarchal: That's fair. That's what I initially thought and why the logger line mentions the context cancelled: `logger.V(logging.INFO).Info("SLO expired during execution or execution context cancelled, skipping request", "error", sloCtx.Err())`.

We can have the situation where we cancel the batch before the deadline, in which case the sloCtx would be cancelled. I will add the branch to cover all cases.

lioraron (approved):

internal/processor/worker/executor.go
```diff
@@ -683,8 +684,9 @@ func (p *Processor) executeOneRequest(
 		Headers:   headers,
 	}
 
-	if sloCtx.Err() != nil {
-		logger.V(logging.INFO).Info("SLO expired during execution or execution context cancelled, skipping request", "error", sloCtx.Err())
+	switch sloCtx.Err() {
```
j-mok-dev: I think this branch should only handle `context.DeadlineExceeded`.

Parent context cancellation is already handled separately, and including `context.Canceled` here can mislabel graceful shutdown as `ErrCodeBatchCancelled`, which is not the right semantic result.

Something like this seems more appropriate:

if sloCtx.Err() == context.DeadlineExceeded {
    // handle the expiry case here
}

and leave parent-context cancellation to the existing outer handling.

PR #318, opened 2026-04-07
fix: default TLS verification on in guides, demos, and processor Conf…

### LINKED ISSUE #307: [Release][Bug]: docs + demo: safer TLS defaults; Helm template default false for tls_insecure_skip_verify
CodeRabbit flagged our deployment guides and demo scripts for using `tlsInsecureSkipVerify` (processor → gateway) and Istio `DestinationRule` `insecureSkipVerify: true` in copy-paste examples. Those settings disable TLS certificate verification on paths that can carry Authorization and inference payloads, which is appropriate for lab/demo (e.g. self-signed certs) but is easy to carry into production without noticing (CWE-295: Improper Certificate Validation). It also noted the Helm processor ConfigMap template can render `tls_insecure_skip_verify: <nil>` when the value is unset, which is sloppy YAML and unclear behavior.

We are not adopting CodeRabbit’s stronger suggestion to reject `TLSInsecureSkipVerify == true` in `Validate()`, because the flag is an intentional escape hatch for environments that still need it; banning it at startup would conflict with that and invite ad-hoc workarounds.

Scope (adopted): We will address the CodeRabbit TLS findings like this: update deployment guides and demo scripts so `tlsInsecureSkipVerify` / `Istio insecureSkipVerify` are not enabled in copy-paste examples by default, use false or omit (chart default) where possible, and where a demo truly needs verification disabled, add a prominent demo-only WARNING and an explicit opt-in (e.g. env var) rather than hardcoding insecure defaults. We will also fix the Helm template so `tls_insecure_skip_verify` renders a boolean (e.g. default false) and never emits a `nil` in the generated YAML.

Rejected: CodeRabbit also suggested rejecting TLSInsecureSkipVerify == true in Validate() so the processor refuses to start if skip-verify is set. We are not implementing that. The flag exists for legitimate cases (e.g. lab/CI, self-signed clusters, environments without a proper CA bundle yet). Banning it at validation time would fight that intended escape hatch and push operators toward worse workarounds. We treat documentation, safer defaults, and template hygiene as the right balance for this project.

j-mok-dev:
## Why is this PR needed?

CodeRabbit finding: Copy-paste examples and demo scripts previously enabled **TLS certificate verification bypass** by default (`tlsInsecureSkipVerify` on the processor and `insecureSkipVerify` on Istio `DestinationRule`). That is reasonable for some self-signed lab setups but is easy to carry into production ([CWE-295](https://cwe.mitre.org/data/definitions/295.html)). The Helm processor ConfigMap could also render `tls_insecure_skip_verify: <nil>` when the value was unset, which is unclear for operators.

We are **not** rejecting `TLSInsecureSkipVerify == true` in processor `Validate()` at startup — that remains an intentional escape hatch per issue discussion. (Issue #307)

## What does this PR do?

- **Helm** (`processor-configmap.yaml`): `tls_insecure_skip_verify` uses `| default false` for `globalInferenceGateway` and each `modelGateways` entry so rendered YAML is always a boolean.
- **Helm tests**: Assert `false` when `tlsInsecureSkipVerify` is omitted.
- **Demo scripts** (`common.sh`, `deploy-k8s.sh`, `deploy-rhoai.sh`, `deploy-maas.sh`): Default path verifies certificates; **`DEMO_TLS_INSECURE_SKIP_VERIFY=1`** opts into processor skip-verify and Istio `insecureSkipVerify: true`, with **`warn`** when enabled.
- **Guides** (`deploy-k8s.md`, `deploy-rhoai.md`, `deploy-maas.md`, `maas-integration.md`): Main snippets omit processor skip-verify and set **`insecureSkipVerify: false`** on `DestinationRule`; add short **lab vs production** notes and the insecure escape hatch.
- **README** (`examples/deploy-demo/README.md`): Document **`DEMO_TLS_INSECURE_SKIP_VERIFY`**.

## How was this tested?

- [x] `helm unittest charts/batch-gateway`
- [x] Unit tests (`make test`) — unchanged Go code paths; optional full `make pre-commit`
- [x] Manual cluster run — recommend spot-check: default install vs `DEMO_TLS_INSECURE_SKIP_VERIFY=1` if x509 issues appear

## Checklist

## Related Issues

Fixes #307

### DIFF
--- a/charts/batch-gateway/templates/processor-configmap.yaml
+++ b/charts/batch-gateway/templates/processor-configmap.yaml
@@ -42,7 +42,7 @@ data:
       max_retries: {{ .maxRetries }}
       initial_backoff: {{ .initialBackoff | quote }}
       max_backoff: {{ .maxBackoff | quote }}
-      tls_insecure_skip_verify: {{ .tlsInsecureSkipVerify }}
+      tls_insecure_skip_verify: {{ .tlsInsecureSkipVerify | default false }}
       {{- if .apiKeyName }}
       api_key_name: {{ .apiKeyName | quote }}
       {{- end }}
@@ -69,7 +69,7 @@ data:
         max_retries: {{ $cfg.maxRetries }}
         initial_backoff: {{ $cfg.initialBackoff | quote }}
         max_backoff: {{ $cfg.maxBackoff | quote }}
-        tls_insecure_skip_verify: {{ $cfg.tlsInsecureSkipVerify }}
+        tls_insecure_skip_verify: {{ $cfg.tlsInsecureSkipVerify | default false }}
         {{- if $cfg.apiKeyName }}
         api_key_name: {{ $cfg.apiKeyName | quote }}
         {{- end }}
--- a/charts/batch-gateway/tests/processor-configmap_test.yaml
+++ b/charts/batch-gateway/tests/processor-configmap_test.yaml
@@ -13,6 +13,34 @@ tests:
           path: data["config.yaml"]
           pattern: 'model_gateways:'
 
+  - it: should default tls_insecure_skip_verify to false when omitted from global_inference_gateway
+    set:
+      processor.config.globalInferenceGateway:
+        url: "http://global-gw:8000"
+        requestTimeout: "5m"
+        maxRetries: 3
+        initialBackoff: "1s"
+        maxBackoff: "60s"
+      processor.config.modelGateways: null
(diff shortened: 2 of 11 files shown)


j-mok-dev:
@yizhaodev please test this change with the previous one if you can. it tightens TLS verification in the deploy guides and examples/deploy-demo scripts by default, which might impact PoC environments

PR #331, opened 2026-04-08
Add grace period when GC context is canceled

### LINKED ISSUE #181: Graceful shutdown improvements: server order, shutdown vs cancel distinction, client cleanup
## Summary

Comprehensive review of graceful shutdown across all three components (apiserver, processor, gc). The overall implementation is solid — signal handling, context propagation, defer cleanup, and WaitGroup-based worker draining are all in place. However, several issues were found that could cause problems during rolling updates or pod eviction.

### 1. Apiserver: shutdown order is wrong (medium)

In `internal/apiserver/server/server.go` (lines 227-235):

```go
// Current: obs server shuts down FIRST
if err := obsServer.Shutdown(shutdownCtx); err != nil { ... }
if err := httpserver.Shutdown(shutdownCtx); err != nil { ... }
```

The observability server (which serves `/ready` for Kubernetes probes and `/metrics` for Prometheus) shuts down before the API server finishes draining. This means:
- Kubernetes readiness probe becomes unavailable while the API server is still processing requests
- Prometheus cannot scrape final metrics during the drain period

**Correct order:**
1. Mark `serverReady = false` (already done)
2. Shut down API server first (stop accepting new requests, drain existing ones)
3. Shut down obs server last (allow final probe checks and metrics scrape)

### 2. Apiserver: shared shutdown timeout for two servers (low)

Both servers share a single 60s `shutdownCtx`. If the API server takes 55s to drain, the obs server only gets 5s. Each server should have its own timeout, or the order fix (issue #1) naturally mitigates this since the obs server shutdown becomes near-instant.

### 3. Processor: no distinction between pod shutdown and user cancel (medium)

When SIGTERM arrives, the parent `ctx` is cancelled. This is the same context passed to worker goroutines. Inside `job_runner.go`, context cancellation triggers the same error paths as user-initiated batch cancellation.

The semantics should differ:
- **User cancel**: save partial results, mark batch as `cancelled`
- **Pod shutdown**: re-enqueue the job so another pod picks it up

Currently, re-enqueue only happens in some error paths (`job_runner.go` lines 104, 222), not consistently for all shutdown scenarios. A job that was mid-execution during pod shutdown may be marked as `failed` instead of being re-enqueued.

**Suggested fix:** Introduce a shutdown-specific signal (e.g. `shutdownRequested atomic.Bool` or a separate `shutdownCtx`) so workers can distinguish pod shutdown from user cancel and always re-enqueue during shutdown.

### 4. HTTP client has no Close method (low)

`pkg/clients/http/http_client.go` does not expose a `Close()` method. The underlying `http.Transport` idle connections are never explicitly closed during shutdown.

**Fix:** Add a `Close()` method that calls `transport.CloseIdleConnections()`, and call it from `Clientset.Close()`.

### 5. GC: no shutdown grace period (low)

`cmd/batch-gc/main.go` exits immediately when the context is cancelled. If the GC is mid-cycle deleting files, a DB record may be deleted but the physical file deletion interrupted (or vice versa). This is mitigated by the fact that GC operations are idempotent — the next cycle will retry. But a short grace period (e.g. 5-10s) would reduce orphaned state.

### 6. signal.Notify channel not cleaned up (very low)

`internal/util/interrupt/interrupt.go` creates a signal channel but never calls `signal.Stop(signalChan)` when the context is cancelled. This causes a minor goroutine leak (the signal listener goroutine lives until process exit). Negligible impact for long-running processes.

## What's already done well

- Unified signal handling via `interrupt.ContextWithSignal` with force-exit on second signal
- Processor `Stop()` uses a fresh `context.WithTimeout(context.Background(), ShutdownTimeout)` instead of the already-cancelled ctx
- `job_runner.go` uses `context.Background()` for re-enqueue operations that must succeed after shutdown
- Tracer, clients, and file handles all have proper defer-based cleanup
- `terminationGracePeriodSeconds` is configurable in the Helm chart (default 60s)

## Affected Files

- `internal/apiserver/server/server.go` — shutdown order, shared timeout
- `internal/processor/worker/job_runner.go` — shutdown vs cancel distinction
- `internal/processor/worker/worker.go` — `Stop()` method
- `pkg/clients/http/http_client.go` — missing `Close()`
- `cmd/batch-gc/main.go` — no grace period
- `internal/util/interrupt/interrupt.go` — signal.Stop cleanup

loicmarchal:
## Why is this PR needed?

Currently, if the context is cancelled, the garbage collector exits immediately. If the GC is mid-cycle deleting files, a DB record may be deleted but the physical file deletion interrupted (or vice versa).

## What does this PR do?

It adds a grace period of 5 seconds when the GC context is canceled before propagating the cancellation.

## How was this tested?

## Checklist

## Related Issues

Fixes #181 (issue 4.)


loicmarchal:
> Hi @loicmarchal I think `processFile` already handles this fine. The deletion order is physical file first, then DB metadata, and on the next GC cycle, if the physical file is already gone, it's silently skipped and the DB record gets cleaned up.

Thanks @acardace . I agree in terms of order and how it's handled in cycles. I guess the suggestion was to add some grace time to hope to finish the cleaning during the grace period in case of a cancel?

> The 5-second grace period doesn't add meaningful safety here. It just delays the inevitable cancellation, after which you'd still have the exact same partial-deletion possibility if a `processFile` call goes over the grace deadline. You can't eliminate the window entirely without a real transaction (or a two-phase approach), and the code already doesn't need to because it's designed to converge on the next cycle.

Indeed. There's no way to guarantee the processing won't take longer than the grace period. So it becomes difficult to have a clean solution (short grace period and take the risk to end up in the same partial deletion state, or longer grace period and risk blocking the cancellation for too long?)

> So I'd be inclined to close this.

That's fine by me. @yizhaodev since you're the reporter, what do you think? Can we just close this one and leave the solution as it is?

PS: by the way, now I understand the issues in #181 are Claude generated. Maybe we could think of the possibility to add a label in such cases (if we don't have a Claude user to create issue in our GH) to distinguish between an AI suggestion from a thought through requirement?

acardace:
> PS: by the way, now I understand the issues in #181 are Claude generated. Maybe we could think of the possibility to add a label in such cases (if we don't have a Claude user to create issue in our GH) to distinguish between an AI suggestion from a thought through requirement?

Yup, def agree with this.

loicmarchal:
@yizhaodev I'm closing this PR for now following @acardace feedback. If you think we shouldn't close it and still find a solution, I'll reopen it.

PR #338, opened 2026-04-09
ci: add nightly scheduled integration test workflow

lioraron:
## Summary

Add a GitHub Actions workflow (`ci-integration-tests.yml`) that runs `make dev-deploy` and `make test-e2e` on a Kind cluster.

**Triggers:**
- **Schedule:** Daily at 3:00 AM UTC (runs on `main` only)
- **Push:** On merges to `main` and `release-*` branches
- **Manual:** `workflow_dispatch` from any branch that has the workflow file

**Concurrency:** Per-branch groups (`integration-tests-<branch>`) — concurrent pushes to different branches run in parallel; a new push to the same branch cancels the in-progress run.

**On failure:** Automatically opens a GitHub issue labeled `bug` + `ci`, assigned to all CODEOWNERS, with a link to the run logs. Deduplicates by checking for existing open issues with the same labels and title prefix — won't create a new issue if one is already open.

**Also includes:**
- 30-minute job timeout to prevent runaway jobs from holding runners
- Add concurrency groups to `pre-commit.yml` and `ci-dco-signoff.yml` — cancels in-progress runs when a new commit is pushed to the same PR, avoiding wasted runner time on superseded commits
- Bump `actions/setup-go` v5 → v6, `azure/setup-helm` v4 → v5, `actions/github-script` v7 → v8 (Node.js 24 compatibility)
- Skip merge commits in DCO sign-off check (`scripts/check-dco.sh`) — merge commits created by GitHub's "Update branch" button don't include `Signed-off-by` trailers

## Test plan

- [x] Workflow runs successfully on push (verified via temporary branch trigger)
- [x] On failure: issue is created with correct labels (`bug`, `ci`), assignees (CODEOWNERS), and run link (tested with forced failure, issue #335)
- [x] Deduplication: skips issue creation when matching open issue exists
- [x] Node.js 24 deprecation warning resolved after action version bumps
- [x] All actions at latest versions
- [x] DCO check passes with merge commits present

Fixes #333


.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Install Helm
+        uses: azure/setup-helm@v5
+
+      - name: Deploy dev environment
+        run: make dev-deploy
+
+      - name: Run integration tests
+        run: make test-e2e
+
+      - name: Open issue on failure
```
yizhaodev, line 44: so if there is a issue not resolved for days
this action will keep opening new bugs for same issue every day?

.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          go-version-file: go.mod
+
+      - name: Install Kind
+        run: go install sigs.k8s.io/kind@latest
+
+      - name: Install Helm
+        uses: azure/setup-helm@v5
+
+      - name: Deploy dev environment
+        run: make dev-deploy
```
yizhaodev, line 39: to cover all deploy scenarios, can we deploy with S3 and filesystem separately and test both?

.github/workflows/ci-integration-tests.yml
lioraron: Good catch! Added deduplication in ffe9a5d — the workflow now checks for existing open issues with labels `bug` + `ci` and title prefix `Integration tests failed` before creating a new one. If a matching issue is already open, it skips creation and logs a message.

.github/workflows/ci-integration-tests.yml
lioraron: Good idea! This PR focuses on getting the workflow infrastructure in place with a single backend. Opened #345 to add a job matrix that tests both S3 (via MinIO in Kind) and filesystem backends as a follow-up.

.github/workflows/ci-integration-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+              console.log('Skipping: open integration test failure issue already exists');
+              return;
+            }
+            const content = await github.rest.repos.getContent({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              path: '.github/CODEOWNERS'
+            });
+            const text = Buffer.from(content.data.content, 'base64').toString();
+            const owners = [...new Set(text.match(/@[\w-]+/g).map(h => h.slice(1)))];
```
j-mok-dev: nit: `text.match(/@[\w-]+/g)` can return `null` if no matches. it won't really happen as our CODEOWNERS is filled, but adding a guard can be better.

j-mok-dev:
consider adding 'timeout-minutes'?
for now, everything should be done in 15 mins, so we can set it for 30 mins giving it enough buffer. without an explicit timeout, it can stuck and hold the runner for default(6 hrs)

yizhaodev (approved):

.github/workflows/ci-integration-tests.yml
lioraron: Good catch — added the guard in 625e6e3: `(text.match(/@[\w-]+/g) || [])`.

j-mok-dev (approved):

lioraron:
> consider adding 'timeout-minutes'?

Good call — added `timeout-minutes: 30` in 0ddb54b.

PR #352, opened 2026-04-13
fix(security): restrict GW listerner to only required namespace

zdtsw:
## Why is this PR needed?

- to prevenet any namespace in the cluster could attach an HTTPRoute to this GW. This is a security risk, a rogue namespace could attach routes and intercept or inject  traffic. 
- similar practice in https://github.com/opendatahub-io/opendatahub-operator/pull/2807

## What does this PR do?
- replace from: All to use selector on the namespace where api-server deployment is
- fix lint when run `"make ci"`

## How was this tested?

## Checklist

### DIFF
--- a/docs/guides/deploy-k8s.md
+++ b/docs/guides/deploy-k8s.md
@@ -343,7 +343,10 @@ spec:
     port: 80
     allowedRoutes:
       namespaces:
-        from: All
+        from: Selector
+        selector:
+          matchLabels:
+            batch-gateway.llm-d.ai/apiserver-route: "true"
   - name: https
     protocol: HTTPS
     port: 443
@@ -353,7 +356,10 @@ spec:
       - name: ${GATEWAY_NAME}-tls
     allowedRoutes:
       namespaces:
-        from: All
+        from: Selector
+        selector:
+          matchLabels:
+            batch-gateway.llm-d.ai/apiserver-route: "true"
 EOF
 
 # wait for gateway instance is ready
@@ -362,6 +368,8 @@ kubectl wait --for=condition=Programmed --timeout=300s \
 ```
 
 > **Note**: The Gateway uses a self-signed certificate from cert-manager (not OpenShift router certs). Access via `kubectl port-forward` with `-k` (insecure) flag on curl.
+
+> **Security**: The Gateway uses `allowedRoutes.namespaces.from: Selector` to restrict HTTPRoute attachment. Only the namespace where the batch-gateway apiserver is deployed needs the label `batch-gateway.llm-d.ai/apiserver-route: "true"`. This is applied automatically in the batch-gateway installation step below.
 
 </details>
 
@@ -605,6 +613,7 @@ kubectl wait tokenratelimitpolicy/inference-token-limit \
 ```bash
 BATCH_NS=batch-api
 kubectl create namespace "${BATCH_NS}" 2>/dev/null || true
+kubectl label namespace "${BATCH_NS}" batch-gateway.llm-d.ai/apiserver-route=true
 
 # Install Redis
 helm install redis oci://registry-1.docker.io/bitnamicharts/redis \
--- a/docs/guides/deploy-k8s.md
+++ b/docs/guides/deploy-k8s.md
@@ -346,7 +346,7 @@ spec:
         from: Selector
         selector:
           matchLabels:
-            batch-gateway.llm-d.ai/apiserver-route: "true"
+            llm-d.ai/gateway-route: "true"
   - name: https
     protocol: HTTPS
     port: 443
@@ -359,7 +359,7 @@ spec:
         from: Selector
         selector:
           matchLabels:
-            batch-gateway.llm-d.ai/apiserver-route: "true"
+            llm-d.ai/gateway-route: "true"
 EOF
 
 # wait for gateway instance is ready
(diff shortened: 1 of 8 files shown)


lioraron (commented):
@zdtsw Thanks for the PR.
Can you address the copilot comments?

@yizhaodev Pls review as well?

yizhaodev:
@zdtsw @lioraron
in current doc/scripts, batch-gateway and LLM shared the same gateway CR
so you need to add LLM namespace label to gateway CR allow list
otherwise this will break as LLM traffic will be blocked.

yizhaodev:
another issue is

In RHOAI case, it ask user to follow official doc to create gateway CR
https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/ingress_and_load_balancing/configuring-ingress-cluster-traffic#ingress-gateway-api

The gateway CR is global shared. probably should not restricted by batch-api or LLM only

zdtsw:
> another issue is
> 
> In RHOAI case, it ask user to follow official doc to create gateway CR https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/ingress_and_load_balancing/configuring-ingress-cluster-traffic#ingress-gateway-api
> 
> The gateway CR is global shared. probably should not restricted by batch-api or LLM only

yes, thanks
after reading the review i realized this was missed, i've done the updated to include both and updated label name to make it more flexible

zdtsw:
> > another issue is
> > In RHOAI case, it ask user to follow official doc to create gateway CR https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/ingress_and_load_balancing/configuring-ingress-cluster-traffic#ingress-gateway-api
> > The gateway CR is global shared. probably should not restricted by batch-api or LLM only
> 

As for the GW CR, for Kserve it requires one for itself.
i would expect if batch ships out for llm-d and llm-d in turn is enabled by kserve, then eventually it is either using the same GW which kserve is using or create another one just for batch.

yizhaodev:
> > > another issue is
> > > In RHOAI case, it ask user to follow official doc to create gateway CR https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/ingress_and_load_balancing/configuring-ingress-cluster-traffic#ingress-gateway-api
> > > The gateway CR is global shared. probably should not restricted by batch-api or LLM only
> 
> As for the GW CR, for Kserve it requires one for itself. i would expect if batch ships out for llm-d and llm-d in turn is enabled by kserve, then eventually it is either using the same GW which kserve is using or create another one just for batch.

@zdtsw 

`As for the GW CR, for Kserve it requires one for itself`
do you mean LLMInferenceService CR need a gateway for inferencing

the current flow is:
1) user create gatewat CR `openshift-ai-inference`
2) user create LLMInferenceService CR and refer to gateway `openshift-ai-inference`
```
apiVersion: serving.kserve.io/v1alpha1
kind: LLMInferenceService
spec:
  router:
    route: {}
    gateway:
      refs:
        - name: ${GATEWAY_NAME}
          namespace: ${GATEWAY_NAMESPACE}
```
3) user create batch-api and refer to same gateway

Is this good enough?

zdtsw:
Probably my previous comments are not clear enough.

so here in batch we actually share the same GW CR for two different parts:
- /v1/batch => api-server is using one HTTPRoute CR
- /v1/chat/completions => processor is pointing to differnt HTTPRoute CRs

for the existing flow in Kserve, it creates a dedicated GW CR `openshift-ai-inference` , which is used for different HTTPRoute CRs, due to this reason it is doing a "All" case (IMHO: this should be more work on that part with selector etc in the future)

so for batch to work with Kserve:
"openshift-ai-inference" can be used in different namespaces
and for simplicity, api-serve can share the same GW openshift-ai-inference as well. (or create a dedicated one)

and for batch to work indepedenctly(without llmisvc but purely with llm-d) in the example:
it is good to not full open for all namespaces in GW CR.
- deploy-k8s.sh here is a pure llm-d case, script creates  GW CR + label both namespace
- deploy-rhoai.sh only one llmisvc CR gets created and the namespaces need labels

The docs/examples do not change how, in production, batch work with kserve.

yizhaodev (approved):

lioraron:
@zdtsw can you update this branch with the latest `main`?

zdtsw:
> apiVersion: serving.kserve.io/v1alpha1
> kind: LLMInferenceService
> spec:
>   router:
>     route: {}
>     gateway:
>       refs:
>         - name: ${GATEWAY_NAME}
>           namespace: ${GATEWAY_NAMESPACE}

i dont think you will need to set this " spec.router.gateway.refs"
unless you want to create another GW CR than "openshift-ai-inference"
then it will be something like this

- user create gatewat CR `ABC`
- user create namespace `user-namespace`  
- user create LLMInferenceService CR `batch-inf `and refer to gateway `ABC`
```
apiVersion: serving.kserve.io/v1alpha1
kind: LLMInferenceService
name: batch-inf
namespace: user-namespace
spec:
  router:
    route: {} => user could set it to create HTTPRoute for you instead of doing it yourself
    gateway:
      refs:
        - name: ABC
          namespace: user-namespace
...
```
- user create batch-api and refer to same gateway `ABC`

zdtsw:
> @zdtsw can you update this branch with the latest `main`?

updated

PR #118, opened 2026-03-11
docs: add crash recovery design document

j-mok-dev:
## Summary

Design document for crash recovery of orphaned batch jobs.

When a processor pod crashes, dequeued jobs become stuck in non-terminal status indefinitely. This document proposes a detection and recovery mechanism using:

- **Redis processor registry + health check** for crash detection (two-phase: suspected → confirmed dead)
- **Version-based optimistic locking** for fencing between processor and reconciler
- **Phase-aware startup recovery** for container restarts where local artifacts survive
- **batch-reconciler** (renamed from batch-gc) as the scanner process

## Reading guide

The document is ~400 lines. If time is limited, start with:
1. **Overview** — problem statement
2. **Required Changes** — the four additions at a glance
3. **Scanner Flow** — concrete example with named processors

The rest fills in the details: fencing mechanics, pseudocode, interface changes, observability, and implementation plan (4 PRs).

PR #359, opened 2026-04-15
docs: Proposal/orphan job reconciliation

j-mok-dev:
## Why is this PR needed?

When a processor pod is replaced after dequeuing a job, the job disappears from Redis while PostgreSQL still shows a non-terminal status. Startup recovery cannot close this gap because there is no surviving workdir. Without an explicit reconciler, such jobs remain stuck indefinitely with no component aware of them.

This proposal also covers a second failure mode: a worker that fails to re-enqueue or transition a job to a terminal DB state after a best-effort attempt — logs the error and moves on. The pod is healthy, startup recovery will not run, and the job is silently orphaned.

## What does this PR do?

Adds a proposal document for orphan batch job detection and reconciliation:
`docs/proposals/orphan-job-detection-reconciliation.md`

The proposal defines:
- A precise definition of **orphan job** (narrow scope: diverged, not startup-recovery eligible)
- A prerequisite hardening step to reduce orphan creation before adding new infrastructure
- **Layer 1**: per-job in-flight tracking via a no-TTL Redis hash (`pod` + `ts`)
- **Layer 2**: DB-driven orphan detection (PostgreSQL scan + ZSET membership + heartbeat staleness)
- **Layer 3**: triage rules (`cancelling` → cancelled, expired → expired, others → re-enqueue as validating)
- **Layer 4**: safe clear with startup recovery non-interference via K8s API pod status check

Compared with earlier recovery designs (PR #118), this proposal removes processor registries, version-based fencing, and per-processor Redis sets. The hot-path change is two Redis O(1) operations (`HSET` after dequeue, `HDEL` on completion); all heavier logic runs in a separate background reconciler.

Two open questions are deferred for team discussion: where the reconciler lives (`batch-gc`, processor with lease, or new component), and whether the K8s API check is required or a skip-and-wait fallback is acceptable.

## How was this tested?

- [x] Manual testing performed — proposal-only PR, no code changes

## Checklist

## Related Issues

Related to #135


j-mok-dev:
proposal context was carried forward into #434 / #437

PR #360, opened 2026-04-15
fix: staticcheck SA5011 nil pointer dereference in tests

lioraron:
## Summary

- Fix all `staticcheck SA5011` warnings across the repo by adding explicit `return` statements after `t.Fatal`/`t.Fatalf` calls that guard nil pointer dereferences.
- Although `t.Fatal` terminates the test, staticcheck cannot prove it is noreturn, so it flags the subsequent dereference as a potential nil pointer access.
- These warnings were not caught by CI because the `golangci-lint-action` in the pre-commit workflow only lints files changed in the PR (`--new-from-rev`), while `make pre-commit` locally runs against the entire codebase.

### Files fixed (24 `return` statements added)

| File | Fixes |
|------|-------|
| `internal/database/redis/redis_ds_client_test.go` | 3 |
| `pkg/clients/http/http_client_test.go` | 8 |
| `pkg/clients/inference/inference_client_test.go` | 9 |
| `internal/processor/config/config_test.go` | 1 |
| `internal/processor/worker/executor_test.go` | 1 |
| `internal/processor/worker/poller_test.go` | 2 |

## Test plan

- [x] `golangci-lint run ./...` passes with 0 issues
- [x] `go test ./...` still passes
- [x] `make pre-commit` passes

### DIFF
--- a/internal/database/redis/redis_ds_client_test.go
+++ b/internal/database/redis/redis_ds_client_test.go
@@ -609,6 +609,7 @@ func TestRedisDSClient(t *testing.T) {
 		}
 		if ec == nil {
 			t.Fatalf("Invalid event consumer channel")
+			return
 		}
 		if ec.ID != ID {
 			t.Fatalf("Mismatch ID %s != %s", ec.ID, ID)
@@ -1849,6 +1850,7 @@ func isEqualBatchItem(t *testing.T, a, b *db_api.BatchItem) bool {
 	t.Helper()
 	if a == nil || b == nil {
 		t.Fatalf("Invalid items to compare")
+		return false
 	}
 	if a.ID != b.ID {
 		t.Fatalf("Mismatch id %s != %s", a.ID, b.ID)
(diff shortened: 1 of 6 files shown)


internal/database/redis/redis_ds_client_test.go
```diff
@@ -1849,6 +1850,7 @@ func isEqualBatchItem(t *testing.T, a, b *db_api.BatchItem) bool {
 	t.Helper()
 	if a == nil || b == nil {
 		t.Fatalf("Invalid items to compare")
+		return false
```
yizhaodev, line 1853: I recall `t.Fatal` will exit go routine directly. no need to return after it.

lioraron: You're right that `t.Fatal` calls `runtime.Goexit()` under the hood, so the test does stop executing. The issue is that `staticcheck` (SA5011) performs static analysis and cannot prove that `t.Fatal` is a terminating statement — Go's type system doesn't have a `noreturn` concept. So when it sees:

```go
if foo == nil {
    t.Fatalf("expected non-nil")
}
foo.Bar() // SA5011: possible nil pointer dereference
```

it flags the dereference as a potential nil pointer access.

Adding `return` after `t.Fatal`/`t.Fatalf` is the standard workaround — it gives the static analyzer the proof it needs without changing runtime behavior. This is also recommended in the [staticcheck documentation](https://staticcheck.dev/docs/checks/#SA5011) and is a common pattern in the Go ecosystem (e.g., [golang/go#34494](https://github.com/golang/go/issues/34494)).

yizhaodev (approved):

PR #361, opened 2026-04-15
doc: remove unnecessary gateway.refs in LLMInferenceService

### LINKED ISSUE #357: deploy-rhoai.sh: unnecessary gateway.refs in LLMInferenceService when using default GW
## Problem

In `deploy-rhoai.sh`, the LLMInferenceService CR explicitly sets `spec.router.gateway.refs` to reference the `openshift-ai-inference` gateway. According to [this comment](https://github.com/llm-d-incubation/batch-gateway/pull/352#issuecomment-4237956186) and [this comment](https://github.com/llm-d-incubation/batch-gateway/pull/352#issuecomment-4238088998), this is unnecessary when using the default gateway — KServe will automatically associate with `openshift-ai-inference` without explicit `gateway.refs`.

The `gateway.refs` field is only needed when using a **custom** (non-default) gateway.

Additionally, `spec.router.route: {}` can let KServe auto-create the HTTPRoute, so the script may not need to create it manually.

## Expected

- Remove explicit `gateway.refs` when using the default `openshift-ai-inference` gateway
- Evaluate whether the manual HTTPRoute creation can be replaced by `spec.router.route: {}`

## Reference

- PR #352 discussion: https://github.com/llm-d-incubation/batch-gateway/pull/352#issuecomment-4238088998

yizhaodev:
## Why is this PR needed?

`spec.router.gateway.refs` in LLMInferenceService is unnecessary when using the default gateway. KServe resolves the gateway from `inferenceservice-config` ConfigMap (`kserveIngressGateway: openshift-ingress/openshift-ai-inference`), so explicit refs are redundant and misleading.

## What does this PR do?

- Remove `spec.router.gateway.refs` from LLMInferenceService CR in deploy script and docs

## How was this tested?

- [x] Full install + 13/13 tests passed on clean OpenShift cluster (RHOAI beta channel)
- [x] Verified LLMInferenceService finds gateway without `gateway.refs`
- [x] Verified clean uninstall/reinstall cycle

## Related Issues

Fixes #357

### DIFF
--- a/examples/deploy-demo/deploy-rhoai.sh
+++ b/examples/deploy-demo/deploy-rhoai.sh
@@ -159,6 +159,8 @@ EOF
     log "Cluster domain: ${domain}, Gateway hostname: ${hostname}"
 
     # Gateway CR
+    # KServe does NOT create the Gateway CR itself — it only creates HTTPRoutes
+    # that reference this Gateway. The Gateway must be pre-provisioned.
     if kubectl get gateway ${GATEWAY_NAME} -n "${GATEWAY_NAMESPACE}" &>/dev/null; then
         log "Gateway already exists. Skipping."
     else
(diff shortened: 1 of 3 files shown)


examples/deploy-demo/deploy-rhoai.sh
```diff
@@ -505,10 +507,12 @@ spec:
   replicas: ${MODEL_REPLICAS}
   router:
     route: {}
-    gateway:
-      refs:
-        - name: ${GATEWAY_NAME}
-          namespace: ${GATEWAY_NAMESPACE}
+    #gateway:
```
j-mok-dev, line 510: nit: the doc removes this block silently while the script keeps it with an explanation comment. maybe it's better adding a similar note in the doc so readers understand why the field is absent.

j-mok-dev (approved):

PR #367, opened 2026-04-16
feat: support install RHOAI via custom catalog image

### LINKED ISSUE #343: feat: support installing RHOAI from custom CatalogSource in deploy scripts
## Problem

The deploy script (`deploy-rhoai.sh`) currently installs RHOAI from the default `redhat-operators` CatalogSource. For testing pre-release or custom builds of RHOAI, users need to install from a custom CatalogSource (e.g. a custom operator index image), but the script does not support this.

## Proposal

Add environment variables to allow users to specify a custom CatalogSource for RHOAI installation:

```bash
# Defaults (can be overridden before running the script)
RHOAI_CATALOG_SOURCE="${RHOAI_CATALOG_SOURCE:-redhat-operators}"
RHOAI_CATALOG_SOURCE_NS="${RHOAI_CATALOG_SOURCE_NS:-openshift-marketplace}"
RHOAI_CHANNEL="${RHOAI_CHANNEL:-fast-3.x}"
```

When a custom catalog image is provided, the script should create a custom `CatalogSource` CR before creating the `Subscription`:

```bash
# Optional: create custom CatalogSource if image is specified
RHOAI_CATALOG_IMAGE="${RHOAI_CATALOG_IMAGE:-}"  # e.g. quay.io/my-org/rhoai-operator-index:latest
```

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: rhoai-custom-catalog
  namespace: openshift-marketplace
spec:
  sourceType: grpc
  image: ${RHOAI_CATALOG_IMAGE}
  displayName: RHOAI Custom Catalog
  updateStrategy:
    registryPoll:
      interval: 30m
```

Then the `Subscription` references the custom catalog:

```yaml
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: rhods-operator
  namespace: redhat-ods-operator
spec:
  channel: ${RHOAI_CHANNEL}
  installPlanApproval: Automatic
  name: rhods-operator
  source: ${RHOAI_CATALOG_SOURCE}
  sourceNamespace: ${RHOAI_CATALOG_SOURCE_NS}
```

## Use case

```bash
# Install RHOAI from a custom catalog
RHOAI_CATALOG_IMAGE=quay.io/my-org/rhoai-operator-index:v3.5-rc1     ./scripts/demo/deploy-rhoai.sh
```

## Related

- #342

yizhaodev:
## Why is this PR needed?

The RHOAI operator is normally installed from the default `redhat-operators` catalog. However, for testing early-access (EA) builds or pre-release versions, we need to install from a custom FBC (File-Based Catalog) fragment image. Previously there was no way to do this with `deploy-rhoai.sh`.

## What does this PR do?

Adds `CUSTOM_CATALOG` environment variable support to `deploy-rhoai.sh`:

1. **Custom CatalogSource creation** — When `CUSTOM_CATALOG` is set, creates a `CatalogSource` CR in `openshift-marketplace` pointing to the specified catalog image, waits for it to become READY, and configures the Subscription to use it instead of `redhat-operators`.

2. **Post-install verification** — After the operator installs, verifies via the InstallPlan that the operator was actually installed from the custom catalog (not a default one).

3. **Cleanup on uninstall** — Removes the custom `CatalogSource` during `uninstall`.

4. **Minor fix** — Replaces `kubectl wait pod` with `wait_for_deployment` for Authorino readiness check (consistent with the rest of the script).

### Usage

```bash
# Install from a custom catalog image (e.g. EA build)
CUSTOM_CATALOG=quay.io/rhoai/rhoai-fbc-fragment:rhoai-3.4@sha256:abc123... \
RHOAI_VERSION=3.4 \
RHOAI_CHANNEL=beta \
./deploy-rhoai.sh install

# Install from default redhat-operators catalog (unchanged behavior)
./deploy-rhoai.sh install
```

**Note:** Custom catalog images from private registries (e.g. quay.io) require pull secrets configured in `openshift-marketplace` namespace and the global pull secret. For EA bundles from `brew.registry.redhat.io`, an `ImageDigestMirrorSet` is also needed.

## How was this tested?

  - Tested on OpenShift CI cluster with EA build `rhods-operator.3.4.0-ea.1`
  - Verified operator installed from `rhoai-custom-catalog` (not default catalog)
  - All 13 e2e tests passed (LLM auth/authz/rate-limit, batch auth/authz/lifecycle/rate-limit)

## Checklist

## Related Issues

Fixes #343

### DIFF
--- a/examples/deploy-demo/deploy-rhoai.sh
+++ b/examples/deploy-demo/deploy-rhoai.sh
@@ -28,6 +28,7 @@ LLM_NAMESPACE="${LLM_NAMESPACE:-llm}"
 KUADRANT_NAMESPACE="${KUADRANT_NAMESPACE:-kuadrant-system}"
 
 OPERATOR_TYPE="${OPERATOR_TYPE:-rhoai}"    # rhoai or odh
+CUSTOM_CATALOG="${CUSTOM_CATALOG:-}"       # custom catalog image (e.g. quay.io/rhoai/rhoai-fbc-fragment:...)
 RHOAI_VERSION="${RHOAI_VERSION:-3.4}"
 RHOAI_CHANNEL="${RHOAI_CHANNEL:-stable-${RHOAI_VERSION}}"
 ODH_CHANNEL="${ODH_CHANNEL:-fast-3}"
@@ -317,9 +318,8 @@ spec:
       enabled: false
 EOF
 
-    step "Waiting for Authorino pods to be ready..."
-    kubectl wait --for=condition=ready pod -l authorino-resource=authorino \
-        -n "${ns}" --timeout=180s
+    step "Waiting for Authorino deployment to be ready..."
+    wait_for_deployment "authorino" "${ns}" 180s
 
     # If RHOAI was already installed before Connectivity Link, restart controllers
     if kubectl get deployment odh-model-controller -n redhat-ods-applications &>/dev/null; then
@@ -332,6 +332,48 @@ EOF
 }
 
 # ── 5. RHOAI / ODH operator ─────────────────────────────────────────────────
+
+create_custom_catalogsource() {
+    local name="$1"
+    local namespace="$2"
+    local image="$3"
+    local timeout="${4:-120}"
+
+    step "Creating custom CatalogSource '${name}' from image: ${image}"
+
+    # Delete existing CatalogSource to force image refresh
+    if kubectl get catalogsource "${name}" -n "${namespace}" &>/dev/null; then
+        log "CatalogSource '${name}' already exists. Updating..."
+        kubectl delete catalogsource "${name}" -n "${namespace}" --ignore-not-found
+    fi
+
+    kubectl apply -f - <<EOF
+apiVersion: operators.coreos.com/v1alpha1
+kind: CatalogSource
+metadata:
+  name: ${name}
+  namespace: ${namespace}
+spec:
+  sourceType: grpc
+  image: ${image}
+  displayName: "Custom ${name} Catalog"
+  publisher: "Custom"
+  updateStrategy:
+    registryPoll:
+      interval: 10m
+EOF
+
+    step "Waiting for CatalogSource '${name}' to be ready..."
+    if ! kubectl wait catalogsource "${name}" -n "${namespace}" \
+        --for=jsonpath='{.status.connectionState.lastObservedState}'=READY \
+        --timeout="${timeout}s" 2>/dev/null; then
+        local state
+        state=$(kubectl get catalogsource "${name}" -n "${namespace}" \
+            -o jsonpath='{.status.connectionState.lastObservedState}' 2>/dev/null || echo "unknown")
+        die "CatalogSource '${name}' not ready after ${timeout}s (state: ${state})"
+    fi
+    log "CatalogSource '${name}' is ready."
+}
 
 install_rhoai_operator() {
     step "Installing ${OPERATOR_TYPE} operator (OLM)..."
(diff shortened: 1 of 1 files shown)


examples/deploy-demo/deploy-rhoai.sh
```diff
@@ -355,19 +398,30 @@ install_rhoai_operator() {
             ;;
     esac
```
lioraron, line 399: **Note**: Skipping channel validation for custom catalogs is the right call (PackageManifest sync lag), but it means a typo in `RHOAI_CHANNEL` (e.g. `stabble-3.4`) won't be caught early — it'll fail at Subscription creation time with a less obvious error.

Consider adding a delayed validation: after `wait_for_subscription` succeeds (which implies OLM resolved the channel), you could check the subscription's `status.currentCSV` isn't empty. But this is low priority since the existing version check on line 407-409 (`rhods-operator.${RHOAI_VERSION}.*`) already catches most misconfigurations.

examples/deploy-demo/deploy-rhoai.sh
```diff
@@ -333,6 +333,49 @@ EOF
...
+    local name="$1"
+    local namespace="$2"
+    local image="$3"
+    local timeout="${4:-120}"
+
+    step "Creating custom CatalogSource '${name}' from image: ${image}"
+
+    # Delete existing CatalogSource to force image refresh
+    if kubectl get catalogsource "${name}" -n "${namespace}" &>/dev/null; then
+        log "CatalogSource '${name}' already exists. Updating..."
```
lioraron, line 346: **Nit**: The `sleep 5` after deleting an existing CatalogSource is a somewhat fragile wait for the delete to propagate. It works in practice but could be replaced with a more deterministic check:
```bash
kubectl delete catalogsource "${name}" -n "${namespace}" --wait=true
```
`--wait=true` (the default for `kubectl delete`) blocks until the resource is fully removed. If it's already the default, the explicit `sleep 5` is unnecessary.

examples/deploy-demo/deploy-rhoai.sh
```diff
@@ -28,6 +28,7 @@ LLM_NAMESPACE="${LLM_NAMESPACE:-llm}"
 KUADRANT_NAMESPACE="${KUADRANT_NAMESPACE:-kuadrant-system}"
 
 OPERATOR_TYPE="${OPERATOR_TYPE:-rhoai}"    # rhoai or odh
+RHOAI_CATALOG="${RHOAI_CATALOG:-}"        # custom catalog image (e.g. quay.io/rhoai/rhoai-fbc-fragment:...)
```
lioraron: **Minor**: The variable is named `RHOAI_CATALOG` but the custom catalog logic in `install_rhoai_operator` runs for both `rhoai` and `odh` operator types (the catalog name is `${OPERATOR_TYPE}-custom-catalog`). If this is intentionally RHOAI-only, consider guarding the custom catalog block with `[ "${OPERATOR_TYPE}" = "rhoai" ]` to avoid confusion. If it's meant to support ODH too, consider renaming to `CUSTOM_CATALOG` or documenting that it works for both.

examples/deploy-demo/deploy-rhoai.sh
```diff
@@ -409,6 +463,24 @@ EOF
...
+    # Verify operator was installed from the expected catalog source
+    if [ -n "${RHOAI_CATALOG}" ]; then
+        local expected_catalog="${OPERATOR_TYPE}-custom-catalog"
+        local install_plan
+        install_plan=$(kubectl get subscription.operators.coreos.com "${operator_name}" \
+            -n "${namespace}" -o jsonpath='{.status.installPlanRef.name}' 2>/dev/null || echo "")
+        if [ -z "${install_plan}" ]; then
+            die "No InstallPlan found for subscription '${operator_name}'."
+        fi
+        local actual_catalog
```
lioraron, line 475: **Suggestion**: The `bundleLookups[0].catalogSourceRef.name` jsonpath may not be populated on all OLM versions (the `bundleLookups` field was added in OLM 0.18+). On older OLM versions, `actual_catalog` would be empty and the check would `die` even though the operator installed correctly from the custom catalog.

The `2>/dev/null || echo ""` handles the error case, but consider logging a warning instead of dying if `actual_catalog` is empty:
```bash
if [ -z "${actual_catalog}" ]; then
    log "WARNING: Could not verify catalog source from InstallPlan (bundleLookups not populated)"
elif [ "${actual_catalog}" != "${expected_catalog}" ]; then
    die "Operator installed from catalog '${actual_catalog}' instead of '${expected_catalog}'."
fi
```
Not blocking since you've tested on OCP 4.19 where this works, but worth considering for portability.

lioraron (commented):
Clean addition — the `create_custom_catalogsource` function is well-structured with proper wait/error handling, and the post-install verification via InstallPlan is a nice safety check. The Authorino fix (`wait_for_deployment` instead of `kubectl wait pod`) is a good consistency improvement. A few minor observations below.

examples/deploy-demo/deploy-rhoai.sh
yizhaodev: fixed, thanks

examples/deploy-demo/deploy-rhoai.sh
yizhaodev: Agreed — the existing version check (`rhods-operator.${RHOAI_VERSION}.*`) already catches this. No change needed.

examples/deploy-demo/deploy-rhoai.sh
yizhaodev: renamed to `CUSTOM_CATALOG `

examples/deploy-demo/deploy-rhoai.sh
yizhaodev: LLM-D requires OCP >= 4.20, so we don't need to support older versions.

lioraron (approved):
Previous comments have been addressed — `CUSTOM_CATALOG` rename, `sleep 5` removal, and the responses to the other two items are all reasonable.

Two minor nits remaining:

1. **Commit message typo**: "feat: support install RHOAI vai custom catalog image" — "vai" should be "via".
2. **PR body usage example**: Still references `RHOAI_CATALOG=quay.io/...` but the variable was renamed to `CUSTOM_CATALOG`. Worth updating for consistency.

Otherwise LGTM.

yizhaodev:
> Previous comments have been addressed — `CUSTOM_CATALOG` rename, `sleep 5` removal, and the responses to the other two items are all reasonable.
> 
> Two minor nits remaining:
> 
> 1. **Commit message typo**: "feat: support install RHOAI vai custom catalog image" — "vai" should be "via".
> 2. **PR body usage example**: Still references `RHOAI_CATALOG=quay.io/...` but the variable was renamed to `CUSTOM_CATALOG`. Worth updating for consistency.
> 
> Otherwise LGTM.

fixed, thanks

PR #369, opened 2026-04-17
chore: remove init() function from code

yizhaodev:
## Why is this PR needed?

`func init()` makes execution order implicit and harder to test. Replacing it with an explicit `Register()` call improves clarity and aligns with the project coding conventions (CLAUDE.md).

## What does this PR do?

- Converts `func init()` in `internal/apiserver/metrics/metrics.go` to an explicit `Register()` function
- Calls `metrics.Register()` at apiserver startup in `cmd/apiserver/main.go`

## How was this tested?

- [x] `make build` — compiles successfully
- [x] Manual review — no other callers rely on the `init()` side effect

## Checklist

### DIFF
--- a/internal/apiserver/metrics/metrics.go
+++ b/internal/apiserver/metrics/metrics.go
@@ -46,7 +46,8 @@ var (
 	)
 )
 
-func init() {
+// Register registers all HTTP metrics with the default Prometheus registry.
+func InitMetrics() {
 	prometheus.MustRegister(httpRequestsTotal)
 	prometheus.MustRegister(httpRequestDuration)
 	prometheus.MustRegister(httpRequestsInFlight)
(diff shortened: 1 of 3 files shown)


internal/apiserver/metrics/metrics.go
```diff
@@ -46,7 +46,8 @@ var (
 	)
 )
 
-func init() {
+// Register registers all HTTP metrics with the default Prometheus registry.
+func Register() {
```
j-mok-dev: 2 non blocking issues.

nit(naming):  `processor` and `retryclient` packages use `InitMetrics` for metric registration functions. worth keeping them consistent. either rename this to `InitMetrics`, or align the other packages to `Register`.

`init()` was called exactly once by the go runtime, so double-registration was impossible before. but with an explicit `Register()`, another call will panic. It's not a problem since it's only called from `main()` now, but we can consider adding a guard with `sync.Once` (like `retryclient.InitMetrics` does), or add a comment: "must be called exactly once".

j-mok-dev (approved):

internal/apiserver/metrics/metrics.go
yizhaodev: renamed to `InitMetrics `, thanks

PR #370, opened 2026-04-17
feat: use different retry backoff by HTTP status code

yizhaodev:
## Why is this PR needed?

The inference HTTP client currently uses the same exponential backoff for all retryable errors. This is suboptimal: a 429 (rate limit) means the server explicitly asks the client to slow down, while a 502/503/504 is typically a transient infrastructure issue that resolves quickly. 

## What does this PR do?

Adds a status-code-aware `RetryAfterFunc` to the resty client that differentiates retry delays by HTTP status:

- **429 with `Retry-After` header**: honors the server-specified delay
- **429 without header**: uses 3× longer initial backoff than default
- **502/503/504**: retries faster with `maxBackoff/2` ceiling
- **Other 5xx**: standard exponential backoff

Uses `cenkalti/backoff/v5` for exponential backoff computation with jitter. No changes to Config, retry conditions, or upper-layer callers.

## How was this tested?

## Checklist

### DIFF
--- a/pkg/clients/http/http_client.go
+++ b/pkg/clients/http/http_client.go
@@ -24,9 +24,11 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
+	cbackoff "github.com/cenkalti/backoff/v5"
 	"github.com/go-logr/logr"
 	"github.com/go-resty/resty/v2"
 	"github.com/llm-d-incubation/batch-gateway/internal/util/logging"
@@ -36,6 +38,11 @@ import (
 const (
 	HeaderNameReqID       string = "X-Request-ID"
 	HeaderNameContentType string = "Content-Type"
+
+	// rateLimitBackoffMultiplier scales the initial backoff for 429 responses.
+	// A 429 means the server explicitly asks the client to slow down, so we
+	// back off harder than for transient 5xx errors.
+	rateLimitBackoffMultiplier = 3
 )
 
 // HTTPClient implements HTTP client with retry, TLS, and observability support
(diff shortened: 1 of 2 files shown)


pkg/clients/http/http_client.go
```diff
@@ -267,6 +274,85 @@ func MapStatusCodeToCategory(statusCode int) ErrorCategory {
 	}
 }
 
+// newRetryAfterFunc returns a resty RetryAfterFunc that differentiates retry
```
j-mok-dev, line 277: resty's sleepDuration clamps the value returned by `RetryAfterFunc` to `SetRetryMaxWaitTime` (= maxBackoff from config. let's say that it's set to 60s.). so if a server sends Retry-After: 120, the actual wait is 60s, not 120s. This means the client may retry before the rate limit window resets, potentially causing repeated 429s.

pkg/clients/http/http_client.go
```diff
@@ -267,6 +274,85 @@ func MapStatusCodeToCategory(statusCode int) ErrorCategory {
...
+
+		default:
+			return computeBackoff(resp.Request.Attempt, initialBackoff, maxBackoff), nil
+		}
+	}
+}
+
+// computeBackoff uses cenkalti/backoff to calculate exponential backoff with
+// jitter for the given attempt number.
+func computeBackoff(attempt int, base, max time.Duration) time.Duration {
```
j-mok-dev, line 317: this func creates a newly calculated `ExponentialBackOff` on every call. maybe it's better to use simple `min(max, base & 2^(attemp-1)) + jitter`? it's not a blocker but for readability

pkg/clients/http/http_client.go
yizhaodev: I think the previous implmentation has same issue. exponential delay will aslo be clamped by `RetryMaxWaitTime` ?

That is resry's behavior unless we set `RetryMaxWaitTime` to large number, but it could introduce anohter issue

pkg/clients/http/http_client.go
yizhaodev: I prefer to use cenkalti/backoff so I don't have to handle edge cases like overflow or less than 0 etc...

j-mok-dev: make sense. :)

pkg/clients/http/http_client.go
j-mok-dev: fair point. the difference is that expoenential backoff clamping just retries a bit early, while `retry-after` clamping ignores the server instruction which was my concern. but I agree this isn't something to fix in this pr.

j-mok-dev (approved):

PR #371, opened 2026-04-20
docs: update dispatcher formula

evacchi:
Signed-off-by: Edoardo Vacchi <evacchi@users.noreply.github.com>

## Why is this PR needed?

update docs, following up to https://github.com/llm-d-incubation/llm-d-async/pull/106 and @yizhaodev suggestions

## What does this PR do?

we update the docs with an alternative version of the formula of the dispatch budget, where instead we multiply F_SYS by F_EPP 

> $\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ measure **independent capacity dimensions**; the inference backend and the gateway layer respectively. Their denominators may differ (e.g., the saturation detector normalizes by the dynamic pool capacity, while the EPP load normalizes by a configured constant). Because they are not fractions of the same pool, we compute their product:
> 
> $$D = (1 - \mathrm{F}\_\mathrm{SYS}) \times (1 - \mathrm{F}\_\mathrm{EPP}) \times (1 - B)$$ 
> 

## How was this tested?

the new proposal is better informed about the EPP codebase (GAIE), llm-d-async, and the shape of the available metrics

## Checklist

## Related Issues

https://github.com/llm-d-incubation/llm-d-async/pull/106

### DIFF
--- a/docs/design/batch-dispatcher.md
+++ b/docs/design/batch-dispatcher.md
@@ -5,10 +5,12 @@ Related:
 - [\[PUBLIC\] EPP Flow Controller for Priority, Fairness, and Queuing](https://docs.google.com/document/d/1VZL7opFWuwgWquvgiOzLlXAJ633qZ9U-A0ZixGjBgaI/edit?tab=t.0#heading=h.hfyow92z2d0t)
 - [\[PUBLIC\] Improved Flow Control Request Management](https://docs.google.com/document/d/1JxzJc8gNv2wKK5-a8ohb0btn78ymVKw9XMIb4-S-ncA/edit?tab=t.0#heading=h.rutawybt03nl)
 - [https://gateway-api-inference-extension.sigs.k8s.io/api-types/inferencepool/](https://gateway-api-inference-extension.sigs.k8s.io/api-types/inferencepool/)
+- [Async Processor: Dispatch Gates](https://github.com/llm-d-incubation/llm-d-async/blob/main/README.md#dispatch-gates) (llm-d-async)
+- [Dispatch Budget](https://github.com/llm-d-incubation/llm-d-async/blob/main/docs/dispatch-budget.md) (llm-d-async)
 
 ## Summary
 
-This document proposes the introduction of a **Batch Dispatcher** to extend the existing "online batch processing agent" architecture (see [\[Public Doc\] Serving Online Batch via Inference Gateway](https://docs.google.com/document/d/1notkq9s0qOmWmUNonZ8CfI-5jtGtHA4PGMI-xz8sGRE/edit?tab=t.0#heading=h.i76kzr3j3swj)). While the **Inference Gateway (IGW)** acts as the primary component for scheduling and flow control, the Batch Dispatcher serves as a system-load aware gatekeeper. It ensures that batch workloads (low-priority and sheddable) are pulled from message queues and forwarded to the IGW only when the inference pool has sufficient capacity. This prevents low-priority traffic from flooding the system and competing with realtime requests.
+This document details the design of a **Batch Dispatcher** (sometimes also referred to as the [**Async Processor**](https://github.com/llm-d-incubation/llm-d-async)) to extend the existing "online batch processing agent" architecture (see [\[Public Doc\] Serving Online Batch via Inference Gateway](https://docs.google.com/document/d/1notkq9s0qOmWmUNonZ8CfI-5jtGtHA4PGMI-xz8sGRE/edit?tab=t.0#heading=h.i76kzr3j3swj)). While the **Inference Gateway (IGW)** acts as the primary component for scheduling and flow control, the Batch Dispatcher serves as a system-load aware gatekeeper. It ensures that batch workloads (low-priority and sheddable) are pulled from message queues and forwarded to the IGW only when the inference pool has sufficient capacity. This prevents low-priority traffic from flooding the system and competing with realtime requests.
 
 ## Problem statement
 
@@ -41,82 +43,18 @@ Because there is one InferencePool and one EPP per (base) model (see [InferenceP
 #### **Key Components**
 
 * **Batch Processing Agent:** The component described in [\[Public Doc\] Serving Online Batch via Inference Gateway](https://docs.google.com/document/d/1notkq9s0qOmWmUNonZ8CfI-5jtGtHA4PGMI-xz8sGRE/edit?tab=t.0#heading=h.i76kzr3j3swj)
-* **Batch Dispatcher:** A component that reads flow-control metrics, determines a "Dispatch Budget", and forwards sheddable traffic to the Inference Gateway (it could be realized as a standalone container, possibly an EPP pod sidecar; or it can be thought of as part of the **Batch Processing Agent**)
+* **Batch Dispatcher:** A component that reads flow-control metrics, determines a "Dispatch Budget", and forwards sheddable traffic to the Inference Gateway
 * **Message Queue:** A persistent store (e.g., Redis, Pub/Sub, Kafka) that holds asynchronous requests. The queue is a priority queue, sorted according to some policy (e.g. an SLO, tenancy, etc: out of scope here)
-* **Metrics Store:** Provides real-time data on **Inference Pool usage**.
+* **Metrics Store:** Provides real-time data on **Inference Pool usage**. Such data drives the **Batch Dispatcher** logic
 * **Inference Gateway (IGW):** L7 Proxy \+ Endpoint Picker (EPP) \+ any other accessory service, it handles the final routing to model servers.
 
-## Detailed Approach
+## Deployment Model and Lifecycle
 
-The **Batch Dispatcher** pulls requests from the Message Queue, reads metrics from a **Metrics Store** and decides whether to forward them to the IGW: when the metrics are within a certain interval the "gates" are open, and a certain number of requests may go through. Otherwise, the Batch Dispatcher waits until the metrics return within the acceptable range.
+The dispatcher corresponds to the **batch processing agent** described [\[Public Doc\] Serving Online Batch via Inference Gateway](https://docs.google.com/document/d/1notkq9s0qOmWmUNonZ8CfI-5jtGtHA4PGMI-xz8sGRE/edit?tab=t.0#heading=h.i76kzr3j3swj), effectively implementing a "strategy" to pull and forward requests from Message Queue to the IGW.
 
-### Dispatch Budget
-
-In addition to forwarding the requests, we also propose a simple aggregated metric that we call a "**Dispatch Budget"**, to determine **the number of concurrent batch requests** allowed to enter the IGW at a given time.
-
-Using the following definitions:
-
-* $\mathrm{cur}_\mathrm{SYS}$: a measure of currently used system resources
-  * at the time of writing: number of requests
-  * in the future: number of bytes, number of tokens etc
-
-* $\mathrm{max}_\mathrm{SYS}$: a measure of available resources
-  * at the time of writing: total number of possible requests
-  * in the future: total number of bytes, total number of tokens etc
-
-* $\mathrm{cur}_\mathrm{EPP}$: a measure of currently used EPP resources
-  * e.g.: number of requests in the EPP internal queues
-  * possibly: number of bytes, number of tokens etc; as long as consistent with
-    $\mathrm{cur}\_\mathrm{SYS}$ and $\mathrm{max}_\mathrm{SYS}$
-
-  *Note:* given the variability in size of the requests, $\mathrm{max}_\mathrm{SYS}$ is currently meant to be a statically-configured parameter. However this could be turned into readable metrics as well
-
-We define the following formulas:
-
-* **Saturation**: a measure of "fullness" of the system, slightly reformulated from the metric described in [\[PUBLIC\] Improved Flow Control Request Management](https://docs.google.com/document/d/1JxzJc8gNv2wKK5-a8ohb0btn78ymVKw9XMIb4-S-ncA/edit?tab=t.0)
-  $\mathrm{F}\_\mathrm{SYS} = \frac{\mathrm{cur}\_\mathrm{SYS}}{\mathrm{max}\_\mathrm{SYS}}$
-* **Virtual Load** of the EPP: a measure of **fullness** as F, but relative to the internal queues of the EPP
-  $\mathrm{F}\_\mathrm{EPP} = \frac{\mathrm{cur}\_\mathrm{EPP}}{\mathrm{max}\_\mathrm{SYS}}$
-* A configurable **Reserved Baseline** (e.g., *B \= 0.05*) reserved for unexpected bursts in high-priority realtime traffic.
-
-Then, **Dispatch Budget** is a combined view of pool **Saturation,** **EPP Virtual Load**, and the **Reserved Baseline**. Because the denominator for   $\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ is always $\mathrm{max}\_\mathrm{SYS}$, then we can write:
-
-$D = 1 -(\mathrm{F}\_\mathrm{SYS}+\mathrm{F}\_\mathrm{EPP}+B)$
-
-
-For a given Dispatch Budget, the expected remaining capacity of the system is just
-
-$\mathrm{max}_\mathrm{SYS}\times D$
-#### Example
-
-$\mathrm{F}\_\mathrm{SYS} = 0.5,   \quad    \mathrm{F}\_\mathrm{EPP} = 0.1, \quad     B=0.05$
-
-- Dispatch Budget \= *1 \- (0.5 \+ 0.1 \+ 0.05) \= 0.35*
-- $\mathrm{max}\_\mathrm{SYS} = 50$ requests
-- Dispatchable Requests \= *floor(50 \* 0.35) \= 17* requests in parallel
-
-### Request Lifecycle and Flow Control
-
-The **Batch Dispatcher** monitors the metrics and computes a **Dispatch Budget**.
-
-* When the Dispatch Budget *D* is **within a given range** \[0,u\], where **1** indicates **100%** of system capacity, and u**\<1,** then the Batch Dispatcher may forward ($\mathrm{max}\_\mathrm{SYS} \times D$) requests at once.
-* When $D>u$, then the Batch Dispatcher should stop forwarding requests until again $D\leq u$
-
-#### Failure Modes
-
-* **Queue Consumer Failures:** If the Batch Dispatcher pod crashes, the messages remain in the persistent Message Queue (Redis/PubSub), ensuring no data loss.
-* **IGW Backpressure:** If the IGW returns an HTTP error indicating overload (e.g. HTTP 429\) despite the computed budget, then the **saturation is assumed to be close to 1**, and therefore a **dispatch budget close to 0\.** The Batch Dispatcher will not retry until a subsequent update from the metrics store will bring it back within the acceptable limits.
-  * This might happen if a sudden spike of traffic enters the gateway and the metrics did not update on-time.
-* **Unreadable Metrics:** In case of unreadable metrics, the Batch Processor cannot take informed decisions, and therefore will assume a **dispatch budget of 0\.**
-
-### Deployment Model
-
-We propose two alternative strategies for deployment
-
-1. **Self-Contained Deployment:** A component in the **batch processing agent** described [\[Public Doc\] Serving Online Batch via Inference Gateway](https://docs.google.com/document/d/1notkq9s0qOmWmUNonZ8CfI-5jtGtHA4PGMI-xz8sGRE/edit?tab=t.0#heading=h.i76kzr3j3swj), effectively implementing a "strategy" to pull and forward requests from Message Queue to the IGW.
-2. **EPP Extension/Sidecar:** The scheduler lives within the same Pod as the EPP. This allows the scheduler to potentially access shared memory or local IPC for fast updates on model server saturation, bypassing metrics scraping delays.
-
-For the first iteration, we propose to implement **1**.
+Currently, the dispatcher is meant to be deployed stand-alone, with its own lifecycle. The dispatcher
+can read from multiple queues and dispatch to multiple inference pools. It can be also configured to dispatch
+to a single pool; in this case, you would configure one dispatcher for each pool.
 
 ## Open Questions & Thinking Points
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ The system is designed to facilitate efficient processing of batch workloads in 
      - Priority queue, event channels, and status updates: `Redis` or `Valkey`.
      - File storage: `S3`, `filesystem`.
 
-4. **Batch Dispatcher**
+4. **Batch Dispatcher** (also called **Async Processor**, implemented in [llm-d-incubation/llm-d-async](https://github.com/llm-d-incubation/llm-d-async))
    - Implements intelligent flow control to balance batch and interactive workloads.
    - Monitors downstream inference system metrics (e.g. queue depth, latency, utilization).
    - Dynamically adjusts dispatch flow of batch requests based on downstream system load, to minimize interference with interactive requests while meeting batch jobs SLOs.


docs/design/batch-dispatcher.md
```diff
@@ -56,44 +56,35 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
-* **Virtual Load** of the EPP: a measure of **fullness** as F, but relative to the internal queues of the EPP
-  $\mathrm{F}\_\mathrm{EPP} = \frac{\mathrm{cur}\_\mathrm{EPP}}{\mathrm{max}\_\mathrm{SYS}}$
-* A configurable **Reserved Baseline** (e.g., *B \= 0.05*) reserved for unexpected bursts in high-priority realtime traffic.
+Each factor represents the fraction of available capacity along its dimension. If any single dimension is fully saturated, the budget goes to zero. The multiplicative model avoids double-counting when signals are correlated and does not require shared units or denominators.
 
-Then, **Dispatch Budget** is a combined view of pool **Saturation,** **EPP Virtual Load**, and the **Reserved Baseline**. Because the denominator for   $\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ is always $\mathrm{max}\_\mathrm{SYS}$, then we can write:
+For a given Dispatch Budget, the expected remaining capacity can be estimated using $D$. We estimate total system capacity as:
 
-$D = 1 -(\mathrm{F}\_\mathrm{SYS}+\mathrm{F}\_\mathrm{EPP}+B)$
+$$\mathrm{max}\_\mathrm{SYS} = \texttt{ready\\_pods} \times \texttt{max\\_concurrency}$$
```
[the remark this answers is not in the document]
evacchi: incorrect for github flavored markdown

docs/design/batch-dispatcher.md
```diff
@@ -56,44 +56,35 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
 
+where `ready_pods` is the number of ready endpoints in the inference pool (available as the `inference_pool_ready_pods` metric) and `max_concurrency` is the per-endpoint request capacity configured in the EPP's saturation detector (default 100). The number of dispatchable requests is then:
 
-For a given Dispatch Budget, the expected remaining capacity of the system is just
+$$N = \lfloor \mathrm{max}\_\mathrm{SYS} \times D \rfloor$$
+
```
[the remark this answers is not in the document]
evacchi: actually I removed floor, added instead a note about leaving to the impl the rounding strategy, especially at the edges

docs/design/batch-dispatcher.md
```diff
@@ -56,44 +56,35 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
+Each factor represents the fraction of available capacity along its dimension. If any single dimension is fully saturated, the budget goes to zero. The multiplicative model avoids double-counting when signals are correlated and does not require shared units or denominators.
 
-Then, **Dispatch Budget** is a combined view of pool **Saturation,** **EPP Virtual Load**, and the **Reserved Baseline**. Because the denominator for   $\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ is always $\mathrm{max}\_\mathrm{SYS}$, then we can write:
+For a given Dispatch Budget, the expected remaining capacity can be estimated using $D$. We estimate total system capacity as:
 
-$D = 1 -(\mathrm{F}\_\mathrm{SYS}+\mathrm{F}\_\mathrm{EPP}+B)$
+$$\mathrm{max}\_\mathrm{SYS} = \texttt{ready\\_pods} \times \texttt{max\\_concurrency}$$
 
+where `ready_pods` is the number of ready endpoints in the inference pool (available as the `inference_pool_ready_pods` metric) and `max_concurrency` is the per-endpoint request capacity configured in the EPP's saturation detector (default 100). The number of dispatchable requests is then:
```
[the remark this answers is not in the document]
evacchi: added links to the relevant code/docs in GAIE

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,44 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
+For a given Dispatch Budget, the expected remaining capacity can be estimated using $D$. We estimate total system capacity as:
 
-$D = 1 -(\mathrm{F}\_\mathrm{SYS}+\mathrm{F}\_\mathrm{EPP}+B)$
+$$\mathrm{max}\_\mathrm{SYS} = \texttt{ready\\_pods} \times \texttt{max\\_concurrency}$$
 
+where `ready_pods` is the number of ready endpoints in the inference pool (available as the [`inference_pool_ready_pods` metric](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/365a1c0b61fab425e5e63141057bc83558a6c26e/site-src/guides/metrics-and-observability.md#L25-L58)) and [`max_concurrency` is the per-endpoint request capacity configured in the EPP's saturation detector (default 100)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/e7dca1b3673bf0d104595c39afcad7411e61ca0d/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/config.go#L32-L58). The number of dispatchable requests is then:
```
yizhaodev: question: is `max_concurrency` also available in EPP metics?

yizhaodev (approved):

docs/design/batch-dispatcher.md
evacchi: I don't think it is, it should be just a config parameter

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,44 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
 
-For a given Dispatch Budget, the expected remaining capacity of the system is just
+$$N = \mathrm{max}\_\mathrm{SYS} \times D$$
+
+#### Notes
```
j-mok-dev: Just to confirm my understanding of the metric mapping:

F_SYS: `flow_control_pool_saturation` from EPP
max_SYS: `inference_pool_ready_pods × per-endpoint capacity`  — where does this come from? Is it a Prometheus metric or a static config?
F_EPP: sum(`inference_extension_flow_control_queue_size`) / max_SYS? Or `flow_control_queue_bytes-based`? which metric is intended here?

j-mok-dev: checked the answer above: per point capacity is in config. got it. :)

evacchi: > inference_pool_ready_pods

this is a prometheus metric

> per-endpoint capacity

this is a config parameter in the EPP. This constant value will be a config parameter of the dispatcher, and will have to match with the corresponding EPP

> F_EPP: sum(inference_extension_flow_control_queue_size) / max_SYS? Or flow_control_queue_bytes-based? which metric is intended here

Good point, I should clarify. In the abstract form of the formula, F_SYS, F_EPP and B are all values in [0,1], regardless of their meaning, and we multiply their complement (1-x) together.

max_SYS no longer plays a role in the computation of the dispatch budget; it is only used when we estimate the number of requests we want to forward. So, say that budget is `0.4275`:

- If we expect to be able to forward a maximum of 50 requests, then we can forward 50 * 0.4275 ~= 21 requests
- if we expect to be able to forward a maximum of 1024 KiB of data, then we can forward max 1024 * 0.4275 = 437.76 KiB (truncated at the boundary of a request; e.g. if req 1 = 300 KiB + req 2 = 200 KiB, you'll only send req 1) 

I'll update the doc

j-mok-dev (approved):
lgtm. thanks for detailed rationales and clear definition!

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
 
-$\mathrm{F}\_\mathrm{SYS} = 0.5,   \quad    \mathrm{F}\_\mathrm{EPP} = 0.1, \quad     B=0.05$
+##### Notes
 
-- Dispatch Budget \= *1 \- (0.5 \+ 0.1 \+ 0.05) \= 0.35*
-- $\mathrm{max}\_\mathrm{SYS} = 50$ requests
-- Dispatchable Requests \= *floor(50 \* 0.35) \= 17* requests in parallel
+1. $N$ is a real-valued estimate. The rounding strategy (e.g., floor, ceiling, or ensuring a minimum of 1 when $D > 0$) is left to the implementation, as edge cases near zero require care to avoid prematurely stopping dispatch.
+2. `max_concurrency` is expressed in request units here, but the same approach generalizes to other capacity units (e.g., tokens, bytes) given an equivalent per-endpoint capacity constant.
+3. We may also define system capacity in terms of number of bytes; in this case $D$ would be truncated at the boundary of a valid request; for instance, if we expect to be able to forward a maximum of 1024 KiB of data, then we may forward at most $1024 \times 0.4275 = 437.76$ KiB; which means if we have $r_1, r_2, ...$ enqueued; if $r_1$ is 470 KiB and $r_2$ is 100 KiB, then we would only dispatch $r_1$
```
lioraron: The budget computes to 437.76 KiB, but `r_1` is 470 KiB which *exceeds* the budget. The text says "we would only dispatch r_1" — but under the budget constraint, `r_1` doesn't fit.

Either make `r_1` smaller than 437.76 (e.g., `r_1 = 400 KiB, r_2 = 100 KiB` → dispatch only `r_1` because `r_1 + r_2 = 500 > 437.76`) or change the conclusion to say zero requests can be dispatched.

lioraron (commented):
One numerical issue in Note 3.

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
```
lioraron: Re "Because there is one InferencePool and one EPP per (base) model (see [InferencePool - Kubernetes Gateway API Inference Extension](https://gateway-api-inference-extension.sigs.k8s.io/api-types/inferencepool/)), there should be one Batch Dispatcher per InferencePool."

1. Is this 1-1 mandatory? Can it be more efficient to have a single dispatcher deployment handle multiple inference pools?
2. When we say "dispatcher" it has a consumer side (independent) and producer side (embedded). Does this 1-1 cardinality refer to the consumer side? Namely, there can still be multiple producers each sending to multiple queues?

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
-  * e.g.: number of requests in the EPP internal queues
-  * possibly: number of bytes, number of tokens etc; as long as consistent with
-    $\mathrm{cur}\_\mathrm{SYS}$ and $\mathrm{max}_\mathrm{SYS}$
+* $B \in [0,1]$: A configurable **Reserved Baseline** (e.g., *B \= 0.05*) reserved for unexpected bursts in high-priority realtime traffic.
 
-  *Note:* given the variability in size of the requests, $\mathrm{max}_\mathrm{SYS}$ is currently meant to be a statically-configured parameter. However this could be turned into readable metrics as well
+$\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ measure **independent capacity dimensions**; the inference backend and the gateway layer respectively. Their denominators may differ (e.g., the saturation detector normalizes by the dynamic pool capacity, while the EPP load normalizes by a configured constant). Because they are not fractions of the same pool, we compute their product:
 
-We define the following formulas:
+$$D = (1 - \mathrm{F}\_\mathrm{SYS}) \times (1 - \mathrm{F}\_\mathrm{EPP}) \times (1 - B)$$
```
lioraron: Not sure about the multiplication:
- If EPP metric is available, maybe we should assume that it gives a full picture of the system, because it sees both the inference pool and the external queues, so in this case use only this metric?
- If EPP metric is not available, maybe in this case use the inference pool metric, again standalone?

Regarding `×(1−B)`: What's the conceptual difference between this and simply deducting B from (1-metric)? Namely we say B percent should be reserved, so we deduct this percent?

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
-  $\mathrm{F}\_\mathrm{SYS} = \frac{\mathrm{cur}\_\mathrm{SYS}}{\mathrm{max}\_\mathrm{SYS}}$
-* **Virtual Load** of the EPP: a measure of **fullness** as F, but relative to the internal queues of the EPP
-  $\mathrm{F}\_\mathrm{EPP} = \frac{\mathrm{cur}\_\mathrm{EPP}}{\mathrm{max}\_\mathrm{SYS}}$
-* A configurable **Reserved Baseline** (e.g., *B \= 0.05*) reserved for unexpected bursts in high-priority realtime traffic.
+Each factor represents the fraction of available capacity along its dimension. If any single dimension is fully saturated, the budget goes to zero. The multiplicative model avoids double-counting when signals are correlated and does not require shared units or denominators.
 
-Then, **Dispatch Budget** is a combined view of pool **Saturation,** **EPP Virtual Load**, and the **Reserved Baseline**. Because the denominator for   $\mathrm{F}\_\mathrm{SYS}$ and $\mathrm{F}\_\mathrm{EPP}$ is always $\mathrm{max}\_\mathrm{SYS}$, then we can write:
+For a given Dispatch Budget $D$, the remaining capacity $N$ is estimated by scaling the total system capacity $\mathrm{max}\_\mathrm{SYS}$:
```
lioraron: Is the unit of `maxSYS` number of requests or number of tokens?

docs/design/batch-dispatcher.md
evacchi: right! I'll make r1 = 400

docs/design/batch-dispatcher.md
evacchi: I am often confusing this producer/consumer terminology; if consumer is the side that reads from queue(s) and forwards the requests to the IGW, then, yes 1-1 means that for each InferencePool we deploy one consumer component responsible for the request of that Pool

docs/design/batch-dispatcher.md
evacchi: at this time I have always reasoned in terms of requests (max number of requests or max total size of requests)

evacchi: hah, but I have just noticed that it depends on the way the EPP was configured (this is a recent change)

https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/8bd576508de3d1af618606f8d68c2c43bcf47d8b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go#L123


```go
		if d.config.mode == modeTokens {
			totalInflight += load.Tokens
			totalCapacity += d.config.maxTokenConcurrency
		} else {
			totalInflight += load.Requests
			totalCapacity += d.config.maxConcurrency
		}
	}
```

docs/design/batch-dispatcher.md
evacchi: > What's the conceptual difference between this and simply deducting B from (1-metric)? 

fair, indeed we can just do "X*Y-B"

for the other points; first, notice that in general the strategy that we are describing here _does not_ prescribe a specific metric, it just means that we want to use 2 signals, one coming from "the inference pool" itself (F_SYS), the other coming from the EPP specifically; that is, F_SYS describes the load of the inference servers, and F_EPP describes the capacity of the gateway itself.

However, as I am writing this, I'm realizing you do make a valid point about using either of those; because if the gateway is under heavy load, the pool status might be considered irrelevant, because we STILL have to go through the gateway to forward requests.

So, maybe we could use either metric, with F_EPP being primary, and F_SYS "the fallback"?

Now, specifically for the formulas being used for F_SYS and F_EPP, [in the current implementation:](https://github.com/llm-d-incubation/llm-d-async/blob/5c7b796d0b0c8e8b16b01d7bd649031afa0aef00/pkg/async/inference/flowcontrol/promql_metric_source_factory.go#L85-L93)

```go
// NewBudgetPromQL constructs a PromQL expression for the multiplicative dispatch budget:
//
//	(1 - F_SYS) * (1 - F_EPP) * (1 - B)
//
// where:
//   - F_SYS = inference_extension_flow_control_pool_saturation
//   - F_EPP = sum(inference_extension_flow_control_queue_size) / maxSys
//   - B = baseline
func NewBudgetPromQL(promConfig promapi.Config, inferencePool string, maxSys float64, baseline float64) (*PromQLMetricSource, error) {
```

so, F_SYS corresponds to the saturation metric that is _published by the EPP_. 

[Currently the formula for saturation is](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/8bd576508de3d1af618606f8d68c2c43bcf47d8b/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/detector.go#L123):

```
Saturation = Total Inflight Requests / Total MaxConcurrency Capacity.
```

for F_EPP, (and I realize now that the description of the example below is incomplete), we compute the ratio between the current size of the EPP queues and "max_SYS" i.e. `ready_pods * max_concurrency`; where max_concurrency is the max concurrent requests for each pod (so, e.g. ready_pods = 4, max_concurrency = 100; ten max_SYS = 400)

so, F_EPP (at least in this impl!) is only a view of the capacity of the EPP

bottom line, I think both metrics are only available *when flow control is enabled in the EPP*, so we will still need to think about fallbacks.

F_SYS in the EPP is computed by counting requests directly, but vLLM does produce corresponding metrics, so in theory we should be able to compute a similar metric in Prometheus; per instance in the pool:

```
vllm:num_requests_running + vllm:num_requests_waiting / max_concurrency
```

for the entire pool it would be the average across instances

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
 
 Using the following definitions:
 
-* $\mathrm{cur}_\mathrm{SYS}$: a measure of currently used system resources
-  * at the time of writing: number of requests
-  * in the future: number of bytes, number of tokens etc
+* $\mathrm{F}\_\mathrm{SYS} \in [0,1]$: **System Saturation**: a measure of "fullness" of the inference backend (e.g., GPU utilization, request concurrency, KV cache pressure). This is a pluggable signal; the current implementation uses the EPP's saturation detector, but it may change independently.
```
acardace: What about calling this `Inference Pool Saturation`?

docs/design/batch-dispatcher.md
lioraron: What prevents having a single consumer deployment servicing multiple inference pools?

docs/design/batch-dispatcher.md
acardace: As discussed offline, I think it's better to remove the `(1-B)` from the formula and instead change the dispatch budget range from `[0,u]` to `[B, u]` so that when `D<=B` the dispatchers halts and when `D>B` the dispatcher resumes. This effectively removes `(1-B)` as a scaling factor and turns it into a `(D-B)`. I think the policy comes across cleaner this way.

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
 
-$\mathrm{max}_\mathrm{SYS}\times D$
-#### Example
+$$\mathrm{max}\_\mathrm{SYS} = \texttt{ready\\_pods} \times \texttt{max\\_concurrency}$$
+
+where `ready_pods` is the number of ready endpoints in the inference pool (available as the [`inference_pool_ready_pods` metric](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/365a1c0b61fab425e5e63141057bc83558a6c26e/site-src/guides/metrics-and-observability.md#L25-L58)) and [`max_concurrency` is the per-endpoint request capacity configured in the EPP's saturation detector (default 100)](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/e7dca1b3673bf0d104595c39afcad7411e61ca0d/pkg/epp/framework/plugins/flowcontrol/saturationdetector/concurrency/config.go#L32-L58). The number of dispatchable requests is then:
```
acardace: nit: `ready_pods` -> `ready_models_servers` since the context is an InferencePool?

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
+
+$$N = \mathrm{max}\_\mathrm{SYS} \times D = \texttt{ready\\_pods} \times \texttt{max\\_concurrency} \times D $$
+
+For instance:
+
+$$\mathrm{F}\_\mathrm{SYS} = 0.5, \quad \mathrm{F}\_\mathrm{EPP} = 0.1, \quad B = 0.05$$
+
+- $\texttt{ready\\_pods} = 5, \quad \texttt{max\\_concurrency} = 10, \quad \mathrm{max}\_\mathrm{SYS} = 50$
+- Dispatch Budget $D = (1 - 0.5) \times (1 - 0.1) \times (1 - 0.05) = 0.5 \times 0.9 \times 0.95 = 0.4275$
+- Dispatchable Requests $N = 5 \times 10 \times 0.4275 = 21.375$
```
acardace: If you agree on removing the `1-B` from the formula here you have to add an additional step where you deduct `B` and dispatch `N` when positive.

docs/design/batch-dispatcher.md
```diff
@@ -56,50 +56,51 @@ In addition to forwarding the requests, we also propose a simple aggregated metr
...
+2. `max_concurrency` is expressed in request units here, but the same approach generalizes to other capacity units (e.g., tokens, bytes) given an equivalent per-endpoint capacity constant.
+3. We may also define system capacity in terms of number of bytes; in this case $D$ would be truncated at the boundary of a valid request; for instance, if we expect to be able to forward a maximum of 1024 KiB of data, then we may forward at most $1024 \times 0.4275 = 437.76$ KiB; which means if we have $r_1, r_2, ...$ enqueued; if $r_1$ is 470 KiB and $r_2$ is 100 KiB, then we would only dispatch $r_1$
 
 ### Request Lifecycle and Flow Control
 
 The **Batch Dispatcher** monitors the metrics and computes a **Dispatch Budget**.
 
-* When the Dispatch Budget *D* is **within a given range** \[0,u\], where **1** indicates **100%** of system capacity, and u**\<1,** then the Batch Dispatcher may forward ($\mathrm{max}\_\mathrm{SYS} \times D$) requests at once.
+* When the Dispatch Budget *D* is **within a given range** \[0,u\], where **1** indicates **100%** of system capacity, and u**\<1,** then the Batch Dispatcher may forward $\mathrm{max}\_\mathrm{SYS} \times D$ requests at once.
```
acardace: As discussed offline this needs fixing, I think you swapped the meaning of `D` in the next sentence as this range doesn't make sense since `D` is a ratio of available capacity. In short it's not correct for the dispatcher to stop when `D>u`, otherwise per the definition of `D` you'd be stopping when there is capacity to schedule jobs.

docs/design/batch-dispatcher.md
acardace: > * If EPP metric is available, maybe we should assume that it gives a full picture of the system, because it sees both the inference pool and the external queues, so in this case use only this metric?
> * If EPP metric is not available, maybe in this case use the inference pool metric, again standalone?

Not sure about this, I think we need both to be completely decoupled from the EPP and the InferencePool.

If you just use the EPP metric then you're assuming it's doing flow control effectively otherwise the state of the queues might tell you something that it's not the real state of the system. For example the EPP might be trashing the InferencePool, queues are near-empty but no request is succeeding but they're getting immediately scheduled nevertheless. This is an exaggeration but you have to take into account scenarios like these if you don't want to depend on the EPP doing flow control.

If you just use the InferencePool metrics then you're not taking into account the EPP queues, but requests always go through those queues, so you must account for those as you might be dispatching things that could get stuck there (an exaggeration but still possible if we want to decouple ourselves from any EPP/IP logic).

In short I think the formula is correct (but the baseline thing I've mentioned), we should try to take into account both metrics if we don't want to make assumptions on how these metrics relate to each other.

lioraron: Yes, bottom line, suggesting to simply to the following 2 cases:
- There is an EPP metric: D = (1-F_EPP)-B. 
   - In this case we assume that the EPP metric took **everything** into account, namely both the queues status and the inference servers status. If this is not currently the case, we can propose a metrics change in EPP?
- The is no EPP metric: D = (1-F_SYS)-B.

In general, trying to reduce the duplication of logic of calculating metrics, namely - if EPP already calculated we use that?

lioraron: @acardace If every layer in the chain will have to look at all the metrics of all it's preceding layers in the chain - it looks a problematic design, and also means a lot of implementation duplication of calculating metrics and possible conflicts if different layers calculate similar metrics differently.

docs/design/batch-dispatcher.md
lioraron: Yes in watsonx it was also a config option - to do the calculations either based on req or tokens.
Pls see if the dispatcher needs a similar config?

docs/design/batch-dispatcher.md
acardace: I don't think this is a real option, the batch-gateway and the EPP have different lifecycles. Also you'd have to have a single helm-chart/operator deploying both which is not really a possibility for how the components are split (rightfully so).

docs/design/batch-dispatcher.md
acardace: @lioraron I agree, but this depends on how much we decide to be decoupled from the EPP, I think it's a matter of making assumptions rather than duplication. If you don't make assumptions about how the EPP flow control works there's no duplication here.

In terms of formula, I think the one described by @evacchi captures more with less assumptions, but it's a matter of making a decisions. What if the EPP has flow control disabled? It still publishes metrics (I think) but the calculations are less accurate on the batch side.

docs/design/batch-dispatcher.md
acardace: I guess we have to detect this at runtime? Using tokens is much better of course.

docs/design/batch-dispatcher.md
evacchi: moved this section above for clarity

docs/design/batch-dispatcher.md
evacchi: I am reformulating this; TBH the whole "Dispatch Budget" notion might become redundant, if the formula is:

```
D = 1 - X - B
```

then `1-X` might be simply described as a capacity; moreover, I am including @acardace's remarks about D's range, and instead I'm moving B out of the formula for D, making D just:

```
D = 1 - X
```

(then C would become a better symbol); and then "dispatchable amount" is computed as `N = max_SYS * (D - B)` IFF `D > B` i.e. when D is **above** the given threshold (gate is closed otherwise)

evacchi:
As noted in https://github.com/llm-d-incubation/batch-gateway/pull/371#discussion_r3159625271 and in light of all your comments:

> I am reformulating this; TBH the whole "Dispatch Budget" notion might become redundant, if the formula is:
> 
> ```
> D = 1 - X - B
> ```
> 
> then `1-X` might be simply described as a capacity; moreover, I am including @acardace's remarks about D's range, and instead I'm moving B out of the formula for D, making D just:
> 
> ```
> D = 1 - X
> ```
> 
> (then C would become a better symbol); and then "dispatchable amount" is computed as `N = max_SYS * (D - B)` IFF `D > B` i.e. when D is **above** the given threshold (gate is closed otherwise)

docs/design/batch-dispatcher.md
evacchi: I think the comment was originally meant to describe an implementation strategy that would lead to a different deployment model. It does not make sense with respect to the current implementation. It's indeed misleading, let's remove/update it

docs/design/batch-dispatcher.md
evacchi: nothing in theory except "simplicity"; my understanding was that the current implementation is meant to be tied to a specific pool; however, it looks like the dispatcher is not inherently bound to a single inference pool see https://github.com/llm-d-incubation/llm-d-async/blob/9c72a19b11f6e49b71e6ee5d87319a8513d74bb8/README.md#multiple-queues-configuration-file-syntax

lioraron: For batch-gateway the intent is single deployment can handle multiple inference pools / models and multiple tenants.
Can you check if extending the dispatcher to support multiple inference pools requires a code change, and if so we can add an issue for this and consult with Shimi?

evacchi: yeah it supports multiple endpoints, see link

lioraron:
@evacchi Can we split this doc into 2 - in `batch gateway` repo keep a high level doc that focuses on the arch diagram, and then add a PR to the `async` repo - with the details on the gating, formulas, etc?

evacchi:
will do 👍

evacchi:
the doc has now been split into two parts: in this repo there is now only an introduction; the remaining detailed discussion of the formula has been submitted for review at https://github.com/llm-d-incubation/llm-d-async/pull/140

lioraron (approved):

lioraron:
@evacchi pls update this branch with the latest `main`. Thanks.

Issue #298, opened 2026-04-01
[Refactor]: Centralize batch error codes with typed constants and co-located messages


j-mok-dev:
## Issue description

Batch error codes (`batch_expired`, `batch_cancelled`, `batch_failed`) are defined as untyped string constants in `batch_types`, but their associated messages are duplicated as string literals across multiple call sites.

For example, the SLO expiry message `"This request could not be executed before the completion window expired."` appears in both `drainUnprocessedRequests` (line 498) and the new early-return guard in `executeOneRequest` (PR #297).

## Suggested changes

1. Add a named type `BatchErrorCode string` for compile-time safety (prevents mixing with `httpclient.ErrorCategory` or other string types).
2. Co-locate the canonical message with each code constant so callers don't duplicate message strings.

```go
type BatchErrorCode string

const (
    ErrCodeBatchExpired   BatchErrorCode = "batch_expired"
    ErrCodeBatchCancelled BatchErrorCode = "batch_cancelled"
    ErrCodeBatchFailed    BatchErrorCode = "batch_failed"
)

var batchErrorMessages = map[BatchErrorCode]string{
    ErrCodeBatchExpired:   "This request could not be executed before the completion window expired.",
    ErrCodeBatchCancelled: "This request was not executed because the batch was cancelled.",
    ErrCodeBatchFailed:    "This request was not executed because the batch encountered a system error.",
}
```

## Affected code

- `internal/shared/types/job.go` — error code constants (lines 40-44)
- `internal/processor/worker/executor.go` — `drainUnprocessedRequests` and `executeOneRequest` duplicate messages

j-mok-dev:
fix merged

PR #372, opened 2026-04-20
refactor(worker): replace cancel flag with context hierarchy and harden error handling

j-mok-dev:
## Why is this PR needed?

The existing error handling works correctly in normal operation, but has structural properties that make it harder to maintain and extend safely:

1. **Dual cancellation mechanism**: User cancel propagates via `atomic.Bool` flag polled at checkpoints, while SIGTERM and SLO expiry propagate via context cancellation. Both paths work today, but contributors must remember to poll the flag at every new checkpoint. Missing one silently degrades cancel responsiveness rather than failing visibly.

2. **Finalization under parent context**: `handleCancelled`, `handleExpired`, and `handleFailedWithPartial` run uploads and DB writes under the parent context. If SIGTERM arrives during finalization, these writes can be interrupted, potentially leaving a job in a non-terminal state. This is recoverable via startup recovery, but the window exists.

3. **No failover on cancelled-status DB write**: If `UpdateCancelledStatus` fails (e.g. transient DB error), there is no fallback. The job stays non-terminal with no output file IDs recorded, even if uploads already succeeded.

None of these have caused production incidents today. This PR consolidates the two cancellation mechanisms into a single context hierarchy, uses detached contexts for finalization I/O, and adds DB write failover, reducing the maintenance surface before further work on recovery and dispatcher integration.

## What does this PR do?

Replaces the `atomic.Bool` cancel flag with an explicit context hierarchy and restructures error handling for resilience.

**Context hierarchy** (replaces `cancelRequested *atomic.Bool`):
- `ctx` — cancelled by SIGTERM only
- `sloCtx` — cancelled when SLO deadline fires
- `userCancelCtx` — derived from `context.Background()`; cancelled only on user cancel (isolated from SIGTERM/SLO)
- `requestAbortCtx` — derived from `sloCtx`; cancelled by SLO expiry, SIGTERM (via parent chain), user cancel (`watchCancel` calls `requestAbortFn`), or fatal I/O error (sibling model abort). This is the unified stop signal for all dispatch loops and inference calls.

**Error handling improvements**:
- Unexport sentinel errors; add `errShutdown`, `errRequestInputRead`, `errFinalizeFailedOver` for finer routing via `errors.Is`
- `handleCancelled`/`handleExpired`/`handleFailedWithPartial`: use detached context for uploads and DB writes so SIGTERM cannot interrupt finalization
- `handleCancelled`: failover to `UpdateFailedStatus` when cancelled-status write fails, preserving uploaded file IDs
- Move `cleanupJobArtifacts` after terminal DB write so local files survive for startup recovery on DB failure
- `preprocessor`: replace flag polling with dedicated context checks (`sloCtx.Err()`, `userCancelCtx.Err()`) for cancel/expired/shutdown discrimination

**New test coverage** (21 new test functions):
- Context cancellation: cancel after completion, SIGTERM after completion, cancel+SIGTERM race, SLO expiry during ingestion, sibling abort
- Error routing: `errCancelled`, `errShutdown`, `errExpired` with nil task/nil counts edge cases
- DB write failover: cancelled-write falls back to failed, completed-write falls back to failed, file IDs preserved through failover
- Finalization resilience: SIGTERM during finalization still completes, one upload fails while other survives
- End-to-end: `runJob` reaches preprocess, `runJob` completes full lifecycle, `errFinalizeFailedOver` does not trigger double `handleFailed`

### Suggested review order

1. `errors.go` : new sentinel errors (small, sets the vocabulary)
2. `job_execution.go` : new context fields in params struct (small)
3. `worker.go` : `runJob` call site updated (small)
4. `preprocessor.go` :  cancel flag polling → dedicated context checks
5. `job_runner.go` :  context hierarchy setup + `handleJobError` routing
6. `executor.go` :  dispatch loop context usage + sibling abort
7. `cancel.go` :  `watchCancel` + `handleCancelled` failover
8. `finalizer.go` :  detached context for uploads/DB writes
9. Test files :  verify the above

## How was this tested?

- [x] Pre-commit checks pass
- [x] Integration/e2e tests verified

## Checklist

## Follow-up planned

- Recovery path (`recovery.go`) error handling refactor — same patterns, separate PR
- Godoc consistency check across updated functions
- Documentation update for error handling design
- Centralize duplicated batch error code messages (#298)

## Related Issues

- Related to #298 — batch error code messages are still duplicated across call sites in `executor.go`; centralizing them is a follow-up.

### DIFF
--- a/internal/processor/worker/cancel.go
+++ b/internal/processor/worker/cancel.go
@@ -18,12 +18,14 @@ package worker
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/go-logr/logr"
 
 	db "github.com/llm-d-incubation/batch-gateway/internal/database/api"
 	"github.com/llm-d-incubation/batch-gateway/internal/processor/metrics"
 	"github.com/llm-d-incubation/batch-gateway/internal/util/logging"
+	uotel "github.com/llm-d-incubation/batch-gateway/internal/util/otel"
 )
 
 func (p *Processor) watchCancel(ctx context.Context, params *jobExecutionParams) {
@@ -43,8 +45,12 @@ func (p *Processor) watchCancel(ctx context.Context, params *jobExecutionParams)
 			if event.Type == db.BatchEventCancel {
 				logger.V(logging.INFO).Info("watchCancel: cancel event received")
 
-				params.cancelRequested.Store(true)
-				params.abortInferFn()
+				// userCancelFn marks userCancelCtx as cancelled (user-cancel signal).
+				// requestAbortFn stops the dispatch loop immediately via requestAbortCtx.
+				params.userCancelFn()
+				if params.requestAbortFn != nil {
+					params.requestAbortFn()
+				}
 
 				// We don't update the DB status to 'cancelling' here because
 				// the API server already wrote 'cancelling' before sending this event.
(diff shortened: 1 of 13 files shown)


internal/processor/worker/cancel.go
```diff
@@ -54,31 +60,44 @@ func (p *Processor) watchCancel(ctx context.Context, params *jobExecutionParams)
...
-// results are uploaded. When called before executeJob (ingestion), both are nil and only
-// cleanup + status transition is performed.
+// When called after executeJob (execution), requestCounts is non-nil and partial results are
+// uploaded. When called before executeJob (ingestion), requestCounts is nil — only cleanup
+// and status transition are performed. jobInfo is always non-nil on the normal runJob path
+// (it is populated before the job goroutine is launched).
+//
+// Uses a detached context so that a concurrent SIGTERM cannot abort the upload or DB write.
 func (p *Processor) handleCancelled(ctx context.Context, params *jobExecutionParams) error {
+	ioCtx, ioSpan := uotel.DetachedContext(ctx, "handle-cancelled")
```
yizhaodev, line 67: so the `ioSpan` will simply track time elasped?
can we add some attr to it or use it to record error at line 83?

internal/processor/worker/cancel.go
```diff
@@ -43,8 +45,12 @@ func (p *Processor) watchCancel(ctx context.Context, params *jobExecutionParams)
 			if event.Type == db.BatchEventCancel {
 				logger.V(logging.INFO).Info("watchCancel: cancel event received")
 
-				params.cancelRequested.Store(true)
-				params.abortInferFn()
+				// userCancelFn marks userCancelCtx as cancelled (user-cancel signal).
+				// requestAbortFn stops the dispatch loop immediately via requestAbortCtx.
+				params.userCancelFn()
+				if params.requestAbortFn != nil {
```
yizhaodev: it would be better if associate requestAbortCtx with userCancelCtx after declare it in job_runner.go
```
	requestAbortCtx, requestAbortFn := context.WithCancel(sloCtx)
    context.AfterFunc(userCancelCtx, requestAbortFn)
```

then you don't have to explicitly call `params.requestAbortFn()` here.  `requestAbortFn` will be auto cancelled

internal/processor/worker/executor.go
```diff
@@ -360,16 +373,17 @@ func (p *Processor) executeJob(ctx, sloCtx, abortCtx context.Context, params *jo
 //
 // Error strategy in this function: when a goroutine encounters a fatal error, modelErr is captured
 // via errOnce but the context is NOT cancelled within this function. Context cancellation is
-// propagated at the executeJob level (execCancel), which stops dispatch across all models.
+// propagated at the executeJob level (requestAbortFn), which stops dispatch across all models.
 // Already-dispatched goroutines may finish with errors or cancellation rather than successful
-// completion, depending on when execCancel fires.
+// completion, depending on when requestAbortFn fires.
 func (p *Processor) processModel(
 	ctx context.Context,
```
yizhaodev: the name `ctx` is a little confusing here, because it is actually `requestAbortCtx`
can we rename it to `requestAbortCtx` directly?

internal/processor/worker/job_runner.go
```diff
@@ -266,33 +282,63 @@ func (p *Processor) handlePanicRecovery(
...
 			defer bgSpan.End()
 			if enqErr := p.poller.enqueueOne(bgCtx, params.task); enqErr != nil {
 				logger.Error(enqErr, "Failed to re-enqueue the job to the queue")
-				if failErr := p.handleFailed(bgCtx, params.updater, params.jobItem, nil, params.jobInfo); failErr != nil {
-					logger.Error(failErr, "Failed to mark job as failed after re-enqueue failure")
+				// executeJob flushed partial output/error files to disk before returning
+				// errShutdown. Upload them so the user can retrieve whatever completed
+				// before SIGTERM, rather than losing those results silently.
+				if params.requestCounts != nil && params.jobInfo != nil {
+					if failErr := p.handleFailedWithPartial(bgCtx, params.updater, params.jobItem, params.jobInfo, params.requestCounts); failErr != nil {
```
yizhaodev: can we merge `handleFailedWithPartial` and `handleFailed` into one func ?
it seem they always called as pair

and then you can check `jobInfo != nil` in merged `handleFailed`

internal/processor/worker/cancel.go
j-mok-dev: thanks. added.

internal/processor/worker/cancel.go
j-mok-dev: thanks. updated to wire the contexts using afterfunc.

internal/processor/worker/executor.go
j-mok-dev: you're right. updated. thanks.

internal/processor/worker/job_runner.go
j-mok-dev: done. merged into a single `handleFailed` that checks `jobInfo != nil` to decide whether to upload partial results. This eliminated all the if/else branching at call sites.

yizhaodev (approved):

PR #365, opened 2026-04-16
Create CLAUDE.md

yizhaodev:
## Why is this PR needed?

There is no project-level guidance for AI coding assistants, leading to inconsistent code style and conventions.

## What does this PR do?

Adds a `CLAUDE.md` file that defines coding conventions (error handling, logging, goroutines, etc.), build commands, and test workflows for the project.

## How was this tested?

- [x] Manual review — all conventions verified against existing codebase
- N/A for unit/integration/e2e (documentation only)

## Checklist

- [ ] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [x] Code follows project [contributing guidelines](CONTRIBUTING.md)
- N/A — `make ci` / `make test-e2e` (documentation only)

PR #383, opened 2026-04-23
chore: enforce CLAUDE.md error handling and safety conventions

### LINKED ISSUE #366: chore: align codebase with CLAUDE.md conventions
## Context

PR #365 adds `CLAUDE.md` with coding conventions for the project.
This issue tracks existing code that doesn't yet follow those conventions.

## Known gaps

- [ ] Replace `%v` with `%w` in error wrapping (`redis_db.go`, `tls.go`, others)
- [ ] Audit for `return` + `log` both on same error path
- [ ] Audit for `var _ Interface = (*Impl)(nil)` compliance across all implementations
- [ ] Any other gaps found during review

## Notes

Changes should be non-functional (no behavior change).
Block on #365 merge.

j-mok-dev:
## Why is this PR needed?

#365 added `CLAUDE.md` with coding conventions. This PR addresses #366 by auditing the codebase against those conventions and fixing all violations found.

## What does this PR do?

Enforces three CLAUDE.md rules across 11 files:

### 1. Handle errors once — return OR log, never both

Removes `logger.Error` calls from synchronous helper functions that return the error to their caller. Every removed log was verified to have a caller that already logs at the appropriate boundary:

### 2. Use `%w` not `%v` for wrapping errors

Fixes instances where `%v` or `%s` broke the error chain, preventing
callers from using `errors.Is` / `errors.As`:

### 3. Always use comma-ok form for type assertions

Converts four bare type assertions to comma-ok form to prevent panics on
unexpected data

### What was intentionally left unchanged

- **`cmd/*/main.go`**: top-level entrypoints are the final error consumer. log + exit is correct.
- **`recovery.go` fallback chains**: intermediate logging records the primary failure before fallback; if `recoverWithFailed` succeeds it returns `nil`, making the intermediate log the only record.
- **Background goroutines** (`redis_event.go` listener, `redis_queue.go`, `PQDequeue`): no caller to return errors to. logging is the only option.

## How was this tested?

- [x] Manual check: Verified every removed log has a caller that logs the error

## Checklist

## Related Issues

Closes #366

### DIFF
--- a/internal/database/redis/redis_db.go
+++ b/internal/database/redis/redis_db.go
@@ -37,7 +37,6 @@ func (c *BatchDBClientRedis) DBStore(ctx context.Context, item *db_api.BatchItem
 		ctx = context.Background()
 	}
 	if err = item.Validate(); err != nil {
-		logr.FromContextOrDiscard(ctx).Error(err, "DBStore[Batch]: item validation failed")
 		return
 	}
 	return c.dbStore(ctx, &item.BaseIndexes, &item.BaseContents,
@@ -50,7 +49,6 @@ func (c *FileDBClientRedis) DBStore(ctx context.Context, item *db_api.FileItem) 
 		ctx = context.Background()
 	}
 	if err = item.Validate(); err != nil {
-		logr.FromContextOrDiscard(ctx).Error(err, "DBStore[File]: item validation failed")
 		return
 	}
 	return c.dbStore(ctx, &item.BaseIndexes, &item.BaseContents,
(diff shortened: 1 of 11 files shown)


internal/database/redis/redis_db.go
```diff
@@ -68,7 +66,6 @@ func (c *DSClientRedis) dbStore(ctx context.Context,
 
 	ptags, err := packTags(indexes.Tags)
 	if err != nil {
-		logger.Error(err, logPref+": tags packing failed")
```
yizhaodev, line 71: should var `logPref` also be deleted from `dbStore` signature?

j-mok-dev: logpref is still used for successful transaction log so we don't need to delete it.

j-mok-dev: but do we need to log successful transations? hmm.

yizhaodev (approved):

PR #386, opened 2026-04-23
fix: reliable offset-based pagination for Redis database plugin

### LINKED ISSUE #385: Fix unreliable pagination in Redis database plugin
## Problem

The Redis database plugin uses `SCAN` cursors for pagination in `DBGet` queries (by tenant, tags, expiry, purpose). This is unreliable because:

- Redis `SCAN` cursors are opaque pointers into the internal hash table, not stable offsets. Items can be skipped or duplicated if the dataset changes between calls.
- `SCAN`'s `COUNT` parameter is a hint, not a limit — actual page sizes are unpredictable.
- Post-SCAN filtering (by tenant, tags, etc.) further reduces result sets inconsistently.
- The `cursor != 0` end-of-iteration check is fragile with concurrent modifications.

This causes flaky test failures when running pagination tests with the Redis backend.

### Option 1: Sorted set indexes

Maintain a sorted set per query dimension (e.g., `idx:tenant:{id}:jobs`) using `ZADD` on create and `ZREM` on delete. Pagination uses `ZRANGE ... LIMIT offset count` for stable, O(log N + page size) queries.

- **Pros:** Efficient at scale, true offset-based pagination, no Redis blocking.
- **Cons:** Higher code complexity — requires maintaining secondary indexes on every write/delete path, and ensuring atomicity between hash data and sorted set indexes.

### Option 2: Collect-then-slice in Lua

Modify the existing Lua scripts to loop `SCAN` until completion, collect all matching items, sort by key, and return a slice `[start, start+limit]`.

- **Pros:** Low code complexity — only the 4 Lua pagination scripts change, no modifications to Go write/delete paths.
- **Cons:** O(N) per page request (scans everything every time). Lua scripts are atomic in Redis, so a long-running script blocks all other operations.

## Chosen Option

**Option 2 (collect-then-slice)** — Redis is only used as a metadata store for development/test environments; PostgreSQL is the production backend. The dataset sizes in development/test are small enough (hundreds to low thousands) that full-scan pagination is acceptable, and the implementation simplicity outweighs the O(N) cost.

The fix also adds tenant filtering to existing expiry-based tests that were previously passing by luck due to SCAN's incomplete iteration masking cross-test interference in shared miniredis instances.

## Documentation

The batch-gateway docs in the [llm-d repo](https://github.com/llm-d/llm-d) should state that the Redis metadata plugin is for development/test only. This is tracked in [llm-d PR #1187](https://github.com/llm-d/llm-d/pull/1187).

## Changed Files

- `internal/database/redis/redis_get_by_tags.lua`
- `internal/database/redis/redis_get_by_tenant.lua`
- `internal/database/redis/redis_get_by_expiry.lua`
- `internal/database/redis/redis_get_by_purpose.lua`
- `internal/database/redis/redis_ds_client_test.go`

lioraron:
## Summary

- Replace SCAN-cursor pagination with full scan-sort-slice in Lua scripts for reliable, deterministic pagination
- Add pagination tests (by tenant, tags, purpose, expiry) that verify no duplicates, correct page sizes, and stable ordering
- Fix existing expiry tests to scope by tenant ID, avoiding cross-test interference in shared miniredis
- Add `DB_CLIENT_TYPE` matrix option to CI integration tests (postgresql, redis)

> **Note:** The Redis metadata plugin is intended for development and testing environments only, not for production use.

Fixes #385

## Test plan

- [x] All existing unit tests pass (`make pre-commit`)
- [x] Run e2e with `DB_CLIENT_TYPE=redis make dev-deploy && make test-e2e`
- [x] Run e2e with `DB_CLIENT_TYPE=postgresql make dev-deploy && make test-e2e`


internal/database/redis/redis_get_by_expiry.lua
```diff
@@ -31,26 +31,20 @@ end
...
+		local hash = contents_to_hash(contents)
+		local expVal = tonumber(hash["expiry"])
+		if expVal and expVal > 0 and (expVal <= expTime) and (tenantID == nil or tenantID == '' or tenantID == hash["tenantID"]) then
+			table.insert(matched, {key, contents})
 		end
-		-- Add the item to the result.
-		table.insert(result, contents)
 	end
-end
+until scan_cursor == "0"
```
yizhaodev, line 48: will be poor performance, but should be fine for dev/test only

test/e2e/helpers_test.go
```diff
@@ -36,6 +37,38 @@ func getEnvOrDefault(key, def string) string {
...
+		"-n", testNamespace, "-o", "json",
+	).CombinedOutput()
+	if err != nil {
+		t.Logf("helm get values failed, defaulting to postgresql: %v", err)
+		return "postgresql"
+	}
+	// Look for "global":{"dbClient":{"type":"..."}} in the JSON output.
+	// Use simple string search to avoid importing encoding/json in this helper.
+	s := string(out)
+	const marker = `"type":"`
```
yizhaodev: use extractor and unmarshal?
```
        var vals struct {
                Global struct {
                        DBClient struct {
                                Type string `json:"type"`
                        } `json:"dbClient"`
                } `json:"global"`
        }
        if err := json.Unmarshal(out, &vals); err != nil || vals.Global.DBClient.Type == "" {
                return "postgresql"
        }
```

acardace:
@lioraron can you better structure the commits? It's usually better to rebase a branch rather than adding multiple commits changing the same files or have multiple commits with the same names. If you could add better descriptions and logically split the commits it'd be easier to review by commit and the git history would be easier to navigate. 

Also I'd avoid commits like 129ca278bb0cc3769f3cba898d7702d8916ab60e, it's clearer to rebase your branch on top of main and repush.

acardace (approved):

internal/database/redis/redis_get_by_expiry.lua
lioraron: Yes, the `redis` and `valkey` options for metadata storage are designed only for development and testing purpose. They serve only the exchange mechanisms for production purpose.
I modified the docs in llm-d repo to indicate this.

test/e2e/helpers_test.go
lioraron: Good idea, thanks!

yizhaodev (approved):

lioraron:
@acardace Thanks for the feedback. A few points:

This repo is configured with `squash-merge` only, so individual commits are internal to the PR and never appear in the `main` branch history.

For review purposes, keeping separate commits, especially commits with message "Address comments" - lets reviewers see exactly what changed in response to their feedback without having to re-review the entire PR. Rebasing and force-pushing would lose that visibility.

Regarding https://github.com/llm-d-incubation/batch-gateway/commit/129ca278bb0cc3769f3cba898d7702d8916ab60e (merge main into branch): I prefer merge over rebase for syncing with main during an open PR, since it preserves the review timeline and avoids force-pushing over commits that reviewers may have already looked at. Again, with squash-merge this is moot for the final history.

PR #392, opened 2026-04-27
refactor: converge GC client construction to use shared NewClientset

### LINKED ISSUE #391: refactor: converge GC client construction to use shared NewClientset
## Problem

The GC (`cmd/batch-gc/main.go`) manually constructs file-store and database clients using the same factory functions that `NewClientset` wraps, duplicating ~35 lines of switch/wrap logic. This duplication has caused:

1. **Wrapping order inconsistency**: GC wraps retry→tracing, while `NewClientset` wraps tracing→retry
2. **Missing observability**: GC doesn't set `ServiceName` on its Redis client, so metrics/traces are unlabeled
3. **Maintenance burden**: any change to client construction must be replicated in two places

## Proposed solution

Refactor `NewClientset` to:
- Accept a clean `Options` struct instead of 11 exploded parameters (addresses the existing TODO at `clientset.go:147`)
- Make exchange clients (Queue/Event/Status) optional via a `SkipExchangeClient` flag — GC doesn't need them
- Relax the hard requirement that `redisCfg` is non-nil when exchange clients are skipped

Then update all three callers:
- **apiserver** (`internal/apiserver/server/server.go`) — pass `Options` struct
- **processor** (`cmd/batch-processor/main.go`) — pass `Options` struct
- **GC** (`cmd/batch-gc/main.go`) — replace manual construction with `NewClientset(..., SkipExchangeClient: true)`

## Benefits

- Single source of truth for client construction
- Consistent wrapping order (tracing→retry) across all components
- GC picks up `ServiceName` labeling for Redis observability
- Cleaner `NewClientset` API (struct vs 11 params)
- `Clientset.Close()` already nil-checks each client, so no changes needed there

lioraron:
## Summary
- Replaces the 11-parameter `NewClientset` signature with a clean `Options` struct (addresses existing TODO at `clientset.go:147`)
- Adds `SkipExchangeClient` flag so the GC can use the shared constructor without creating unused Queue/Event/Status clients
- Eliminates ~35 lines of duplicated switch/wrap logic in `cmd/batch-gc/main.go`
- Fixes wrapping-order inconsistency (GC had retry→tracing, clientset had tracing→retry)
- Adds `ServiceName="batch-gc"` for Redis observability
- Introduces shared constants (`DBTypeRedis`, `DBTypeValkey`, `DBTypePostgreSQL`, `FileTypeFS`, `FileTypeS3`) to replace scattered string literals

Closes #391

## Test plan
- [x] `make pre-commit` passes (formatting, vet, lint, unit tests, security scan)
- [x] CI e2e tests pass across all matrix combinations (Redis/Valkey × PostgreSQL)

### DIFF
--- a/internal/util/clientset/clientset.go
+++ b/internal/util/clientset/clientset.go
@@ -33,9 +33,9 @@ import (
 	"github.com/llm-d-incubation/batch-gateway/internal/files_store/retryclient"
 	s3client "github.com/llm-d-incubation/batch-gateway/internal/files_store/s3"
 	fstracing "github.com/llm-d-incubation/batch-gateway/internal/files_store/tracing"
+	sharedcfg "github.com/llm-d-incubation/batch-gateway/internal/shared/config"
 	ucom "github.com/llm-d-incubation/batch-gateway/internal/util/com"
 	uredis "github.com/llm-d-incubation/batch-gateway/internal/util/redis"
-	"github.com/llm-d-incubation/batch-gateway/internal/util/retry"
 	"github.com/llm-d-incubation/batch-gateway/pkg/clients/inference"
 )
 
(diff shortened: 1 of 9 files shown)


internal/util/clientset/clientset.go
```diff
@@ -140,114 +140,106 @@ func NewPostgreSQLDBClients(ctx context.Context, cfg *postgresql.PostgreSQLConfi
...
+	FileCfg           sharedcfg.FileClientConfig
+	InferenceGlobal   *inference.GatewayClientConfig
+	InferencePerModel map[string]inference.GatewayClientConfig
+	Component         ucom.Component
 
-	logger := logr.FromContextOrDiscard(ctx)
+	// SkipExchangeClient skips creation of the Redis exchange client
+	// (Queue, Event, Status). Set this for components that only need
+	// database and file-store clients (e.g. the garbage collector).
+	SkipExchangeClient bool
```
yizhaodev: I am thinking about using `WithXXX` pattern instead of a bool flag.

Each caller declares what it needs rather than what to skip,  and adding future client types (e.g. a metrics client) won't require new skip flags.

**For ClientSet**
```
type Option func(*clientsetConfig)

type clientsetConfig struct {
    dbCfg             *sharedcfg.DBClientConfig
    fileCfg           *sharedcfg.FileClientConfig
    exchangeCfg          *sharedcfg.ExchangeConfig
    inferenceGlobal   *inference.GatewayClientConfig
    inferencePerModel map[string]inference.GatewayClientConfig
}

func WithDB(cfg sharedcfg.DBClientConfig) Option {
    return func(c *clientsetConfig) { c.dbCfg = &cfg }
}

func WithFile(cfg sharedcfg.FileClientConfig) Option {
    return func(c *clientsetConfig) { c.fileCfg = &cfg }
}

func WithExchange(cfg *sharedcfg.ExchangeConfig) Option {
    return func(c *clientsetConfig) { c.exchangeCfg = cfg }
}

func WithGlobalInference(cfg inference.GatewayClientConfig) Option {
    return func(c *clientsetConfig) { c.inferenceGlobal = &cfg }
}

func WithPerModelInference(cfgs map[string]inference.GatewayClientConfig) Option {
    return func(c *clientsetConfig) { c.inferencePerModel = cfgs }
}

func NewClientset(ctx context.Context, component ucom.Component, opts ...Option) (*Clientset, error) {
    cfg := &clientsetConfig{}
    for _, o := range opts {
        o(cfg)
    }
}

```

**For callers:**

```
// GC: DB + File
clientset.NewClientset(ctx, ucom.ComponentGC,
    clientset.WithDB(cfg.DBClientCfg),
    clientset.WithFile(cfg.FileClientCfg),
)

// apiserver: DB + File + Exchange
clientset.NewClientset(ctx, ucom.ComponentApiserver,
    clientset.WithDB(config.DBClientCfg),
    clientset.WithFile(config.FileClientCfg),
    clientset.WithExchange(cfg.ExchangeClientCfg),
)

// processor: DB + File + Exchange + Inference
clientset.NewClientset(ctx, ucom.ComponentProcessor,
    clientset.WithDB(cfg.DBClientCfg),
    clientset.WithFile(cfg.FileClientCfg),
    clientset.WithExchange(cfg.ExchangeClientCfg),
    clientset.WithGlobalInference(*resolved.Global),
)
```

lioraron: Good call — done in 715ce00.

Replaced `Options` struct + `SkipExchangeClient` with functional options: `WithDB`, `WithFile`, `WithExchange`, `WithGlobalInference`, `WithPerModelInference`. Each caller now declares exactly what it needs:

- **apiserver**: `WithDB`, `WithFile`, `WithExchange`
- **processor**: `WithDB`, `WithFile`, `WithExchange`, + conditional `WithGlobalInference` / `WithPerModelInference`
- **GC**: `WithDB`, `WithFile` (no exchange — that's the whole point)

All option functions accept values (not pointers) to prevent caller mutation from leaking into `NewClientset` internals.

internal/util/clientset/clientset.go
```diff
@@ -140,114 +140,137 @@ func NewPostgreSQLDBClients(ctx context.Context, cfg *postgresql.PostgreSQLConfi
...
+}
+
+// WithGlobalInference enables creation of a global inference client.
+func WithGlobalInference(cfg inference.GatewayClientConfig) Option {
+	return func(c *clientsetConfig) { c.inferenceGlobal = &cfg }
+}
+
+// WithPerModelInference enables creation of per-model inference clients.
+func WithPerModelInference(cfgs map[string]inference.GatewayClientConfig) Option {
+	return func(c *clientsetConfig) { c.inferencePerModel = cfgs }
```
yizhaodev: `c.inferencePerModel = cfgs` only copy map reference, won't copy values in map, 
since we want to avoid mutation leaking, should copy values.

yizhaodev: `RedisClientConfig.Certificates` is also a pointer, also need to change copy values

yizhaodev: maybe add `DeepCopy` func to each `Config`
to archive the goal: `prevent caller mutation from leaking into NewClientset internals.`

lioraron: Good catch!
Implemented the deep-copy in cec7897.

Added `DeepCopy()` methods to `RedisClientConfig` (clones the `*Certificates` pointer) and `DBClientConfig` (delegates to `RedisCfg.DeepCopy()`). `WithDB` and `WithExchange` now call `DeepCopy()` before capturing. `WithPerModelInference` copies the map into a new allocation.

`WithFile` and `WithGlobalInference` don't need deep copies — `FileClientConfig` and `GatewayClientConfig` are all value types.

yizhaodev (approved):

PR #394, opened 2026-04-28
feat: add AIMD adaptive concurrency control for inference requests

### LINKED ISSUE #302: [flow-control] Adaptive concurrency control based on flow control backpressure
## Summary

Implement an adaptive concurrency controller that dynamically adjusts the processor's `global_concurrency` based on backpressure signals from GIE's flow control layer, replacing the current static concurrency limit.

## Background

The batch processor currently uses a static `global_concurrency` semaphore (default: 100) to limit in-flight inference requests. With GIE flow control managing admission and queuing on the gateway side, this static limit is suboptimal:

- **Too high**: The processor submits more requests than the gateway can queue, resulting in excessive 429 rejections and wasted retry cycles.
- **Too low**: The processor underutilizes available backend capacity when the gateway has room for more requests.

An adaptive controller would use the 429 response rate as a feedback signal to find the optimal concurrency level automatically.

## Proposed Approach

Implement an adaptive concurrency controller (e.g., AIMD -- Additive Increase, Multiplicative Decrease) that wraps the existing semaphore:

- **Additive Increase**: When requests succeed without 429s, gradually increase the concurrency limit (e.g., +1 per N successful requests).
- **Multiplicative Decrease**: When a 429 is received, reduce the concurrency limit by a factor (e.g., halve it).
- **Bounds**: Enforce a configurable minimum and maximum concurrency to prevent collapse or runaway growth.

```yaml
# Example config
adaptive_concurrency:
  enabled: true
  min_concurrency: 10
  max_concurrency: 200
  increase_step: 1
  decrease_ratio: 0.5
  window: "10s"
```

## Scope

- New `internal/processor/worker/adaptive_concurrency.go` implementing the AIMD controller
- Integration with the existing semaphore acquisition in `processModel`
- Configuration fields and validation
- Unit tests for the controller
- Documentation update

## Why This Is Optional

The static concurrency limit works well enough for initial deployments. The retry-with-backoff mechanism in the HTTP client already provides a basic feedback loop. Adaptive concurrency is an optimization that improves efficiency under variable load but is not required for correctness.

## Related

- #301 -- Send priority header for GIE flow control (prerequisite)
- GIE flow control: https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/epp/flowcontrol
- Batch Gateway flow control guide: `docs/guides/flow-control-setup.md`

### LINKED ISSUE #419: Add Prometheus metrics for AIMD concurrency control
## Summary

PR #394 added AIMD adaptive concurrency control but does not expose the current AIMD state as Prometheus metrics. Without these, operators cannot observe AIMD behavior in dashboards or alerts, and e2e tests cannot verify AIMD reactions.

## Proposed metrics

| Metric | Type | Labels | Description |
|--------|------|--------|-------------|
| `batch_processor_aimd_concurrency_limit` | Gauge | `endpoint` | Current effective concurrency limit per endpoint |
| `batch_processor_aimd_decreases_total` | Counter | `endpoint` | Number of multiplicative decreases (429/5xx signals) |
| `batch_processor_aimd_increases_total` | Counter | `endpoint` | Number of additive increases (sustained success windows) |

The gauge is the most important — it enables dashboards showing how AIMD reacts to backpressure over time. The counters provide rate-of-change visibility.

## Scope

- `internal/processor/metrics/metrics.go` — register new metrics
- `internal/processor/worker/executor.go` — update metrics on AIMD state changes
- `pkg/semaphore/aimd.go` — possibly expose a callback or have the caller update metrics after `RecordRateLimit`/`RecordSuccess`
- Grafana dashboard update if one exists

j-mok-dev:
## Why is this PR needed?                                 

  The batch processor uses a fixed `GlobalConcurrency` limit for in-flight inference requests. When the downstream inference gateway is overloaded (returning 429/5xx), the processor keeps dispatching at the same rate, worsening congestion and increasing tail latency for both batch and interactive workloads. Operators must manually tune `GlobalConcurrency` : too high risks overwhelming the gateway, too low wastes capacity.

  ## What does this PR do?

  Adds AIMD (Additive Increase / Multiplicative Decrease) adaptive concurrency  
  control. The global semaphore is now dynamically resizable, and an AIMD
  controller adjusts it based on inference gateway feedback:                    
                                                            
  - **Multiplicative Decrease**: On 429, 5xx, or 200-with-retries signals, the  
  limit is cut by `BackoffFactor` (default 0.5).
  - **Additive Increase**: After a full window of consecutive clean 200s (window
   size = current limit), the limit increases by `AdditiveIncrease` (default 1).
  - **Bounds**: The effective limit stays within `[MinConcurrency,
  GlobalConcurrency]` (default [5, 100]).                                       
  - **Disable**: Set `MinConcurrency = GlobalConcurrency` to pin the limit and
  restore pre-existing behavior.                                                
                                                            
  ### Key design decisions                                                      
                                                            
  - **Why a new `AdaptiveSemaphore` instead of reusing the channel-based one?** 
  Go channels cannot be resized after creation. The adaptive semaphore uses
  `sync.Mutex` + `sync.Cond` to support runtime `SetLimit()`.                   
  - **Why treat 200-with-retries as a rate-limit signal?** If the HTTP retry
  layer absorbed 429s before returning a 200, the gateway was already under     
  pressure. Treating this as a success would mask backpressure and delay AIMD
  response.                                                                     
  - **Why call `setFn` outside the AIMD mutex?** Avoids lock-ordering deadlock
  between the AIMD controller's mutex and the semaphore's mutex. The tradeoff is
   eventual consistency — acceptable for AIMD's approximate control semantics
  (documented in code).                                                         
  - **Non-HTTP errors (network, timeout) skip AIMD entirely.** These carry no
  capacity signal about the gateway — they could be transient infrastructure    
  issues.
                                                            
  ## How was this tested?                                                       
                                                            
    - `adaptive_test.go`: acquire/release, context cancellation, TryAcquire,
  SetLimit wakeup, concurrent load, double-release callback, benchmarks         
    - `aimd_test.go`: initial clamp, additive increase, multiplicative decrease,
   backoff factors, sawtooth pattern, concurrent signals, no-op when at bounds, 
  benchmarks                                                
    - `config_test.go`: AIMD validation (backoff range, min > max, defaults,    
  explicit values preserved)                                                    
    - `executor_test.go`: AIMD signaling in `processModel` — clean 200, 429,
  5xx, 200-with-retries, non-HTTP error skip                                    
    - `http_client_test.go`: retry count propagation on 500, 429, 400 (no retry)
                                                                                
  ## Checklist                                                                  
                                                            
  ## Related Issues                                         

  Fixes #302
  Closes #419


yizhaodev:
2 questions

1) this seems overlap with batch dispatcher (gate mechanism)? which also control sending requests rate.

2) AIMD controller and semaphore are global. 
So If one model gateway returns 429 while others are 200, the limit reduction affects all. 
can we reblance quota among gateways rather than decrease the total number?

j-mok-dev:
> 2 questions
> 
> 1. this seems overlap with batch dispatcher (gate mechanism)? which also control sending requests rate.
> 2. AIMD controller and semaphore are global.
>    So If one model gateway returns 429 while others are 200, the limit reduction affects all.
>    can we reblance quota among gateways rather than decrease the total number?

thanks.
1: the batch dispatcher will be toggleable via config flag. this change is for deployments that don't use the dispatcher. 
2: you're right. this is global, so a 429 from one model unfairly throttles unaffected models. working on the fix.

j-mok-dev:
Update: now refactors global AIMD to **per-endpoint AIMD**.      
                                                                                    
  Models sharing the same inference endpoint (same `InferenceClient`) share one     
  `AdaptiveSemaphore` + `AIMDController` pair. This means a 429 from endpoint A only
   reduces concurrency for models routed to A — endpoint B is unaffected.           
                                                                                    
  Other changes since the initial commit:                                           
  - Capacity retry tracking (`HadCapacityRetry`): distinguishes 429/5xx retries from
   network-error retries for precise AIMD signaling                                 
  - Removed dead `RetryAttempts` field (replaced by `HadCapacityRetry`)             
  - Config: removed `applyConcurrencyDefaults` (silent zero-swallow), rely on       
  `NewConfig()` defaults + `Validate()` rejection                                   
  - AIMD logger now includes `endpoint` URL for operational visibility              
  - Design doc, metrics doc, values.yaml comments updated to reflect per-endpoint   
  semantics                                                                         
  - Tests: endpoint isolation test, capacity vs network retry tests

pkg/clients/http/http_client.go
```diff
@@ -189,9 +213,9 @@ func (c *HTTPClient) Close() error {
 	return nil
 }
 
-// Post makes an HTTP POST request with automatic retry logic
-// Returns the response body, status code, and any error
-func (c *HTTPClient) Post(ctx context.Context, endpoint string, body interface{}, headers map[string]string, requestID string) ([]byte, int, error) {
+// Post makes an HTTP POST request with automatic retry logic.
+// Returns the response body, status code, number of retries performed, and any error.
+func (c *HTTPClient) Post(ctx context.Context, endpoint string, body interface{}, headers map[string]string, requestID string) ([]byte, int, int, error) {
```
lioraron: The third return value (`retries`) is not used by any caller — all call sites assign it to `_`. The capacity retry tracking already works via the context-based mechanism (`NewCapacityRetryContext`), making this redundant. Consider removing it to avoid the unnecessary API surface change and ~20 call-site updates.

internal/processor/worker/worker.go
```diff
@@ -37,24 +37,35 @@ import (
...
 	wg     sync.WaitGroup
 
 	// globalSem limits total in-flight inference requests across all workers.
+	// Fixed capacity — serves as a ceiling only.
 	globalSem semaphore.Semaphore
 
+	// endpointLimits maps each unique InferenceClient to its per-endpoint
+	// adaptive semaphore + AIMD controller. Created in Run() from the
+	// resolver's client set.
+	endpointLimits map[inference.InferenceClient]*endpointLimit
```
lioraron, line 66: Using an interface as map key relies on pointer identity of the concrete `*InferenceHTTPClient`. If someone wraps the client in a decorator/middleware, pointer identity breaks and models get separate AIMD controllers when they should share one. Add a comment here documenting this assumption.

charts/batch-gateway/templates/processor-configmap.yaml
```diff
@@ -14,6 +14,9 @@ data:
     poll_interval: {{ .Values.processor.config.pollInterval | quote }}
     num_workers: {{ .Values.processor.config.numWorkers }}
     global_concurrency: {{ .Values.processor.config.globalConcurrency }}
+    min_concurrency: {{ .Values.processor.config.minConcurrency | default 5 }}
```
lioraron: These new fields use `| default` but existing fields (`globalConcurrency`, `perModelMaxConcurrency`) rely on `values.yaml` defaults without it. For consistency, consider removing `| default` here and relying on the defaults already set in `values.yaml`.

pkg/clients/http/http_client.go
j-mok-dev: Good catch. Thanks. I removed the extra retried return value and reverted call sites.

internal/processor/worker/worker.go
j-mok-dev: Thanks. I added the comment on endpointLimits documenting the identity assumption. (InferenceClient instance identity as map key) This is safe because GatewayResolver dedup/reuses clients for identical endpoint configs.

charts/batch-gateway/templates/processor-configmap.yaml
j-mok-dev: agreed. it's now rely on values.yaml defaults. it's now consistent with existing fields.

lioraron (commented):
The existing Prometheus metrics expose `processor_max_inflight_concurrency` (the fixed global ceiling) and `model_inflight_requests` (current in-flight count per model), but there's no gauge for the current AIMD-adjusted per-endpoint concurrency limit. Operators can observe the *effect* of AIMD (fewer in-flight requests) but not the *cause* (the limit was reduced).

Log lines capture transitions (`AIMD decrease old=10 new=5`), but a gauge would be more operationally useful:
- Plottable in Grafana alongside in-flight and error rate metrics
- Alertable (e.g., "endpoint X AIMD limit dropped below threshold")
- Correlatable at scrape time without log parsing

Suggest adding a gauge like `processor_endpoint_concurrency_limit{endpoint}` that's updated in the `setFn` callback. Not a blocker — the AIMD behavior is correct and observable via logs — but worth a follow-up.

j-mok-dev:
> The existing Prometheus metrics expose `processor_max_inflight_concurrency` (the fixed global ceiling) and `model_inflight_requests` (current in-flight count per model), but there's no gauge for the current AIMD-adjusted per-endpoint concurrency limit. Operators can observe the _effect_ of AIMD (fewer in-flight requests) but not the _cause_ (the limit was reduced).
> 
> Log lines capture transitions (`AIMD decrease old=10 new=5`), but a gauge would be more operationally useful:
> 
> * Plottable in Grafana alongside in-flight and error rate metrics
> * Alertable (e.g., "endpoint X AIMD limit dropped below threshold")
> * Correlatable at scrape time without log parsing
> 
> Suggest adding a gauge like `processor_endpoint_concurrency_limit{endpoint}` that's updated in the `setFn` callback. Not a blocker — the AIMD behavior is correct and observable via logs — but worth a follow-up.


I intentionally did not add a per-endpoint AIMD limit metric in this PR due to cardinality concerns (endpoint/model label explosion risk across dynamic gateway configs). For now, AIMD transitions are observable via structured logs, and this PR keeps the runtime adaptation behavior isolated.

lioraron:
Please resolve the conflicts? Thanks.

charts/batch-gateway/values.yaml
```diff
@@ -304,6 +304,15 @@ processor:
     pollInterval: "5s"
     numWorkers: 20
     globalConcurrency: 100
+    # AIMD floor per endpoint. AIMD adjusts the effective limit within
+    # [minConcurrency, perModelMaxConcurrency] based on 429/5xx backpressure.
+    # Set minConcurrency = perModelMaxConcurrency to disable adaptive behavior.
+    minConcurrency: 5
+    backoffFactor: 0.5
```
lioraron: The flow control guide (`docs/guides/flow-control-setup.md`) should be updated to document AIMD. The "Recommended Processor Settings" section currently shows `global_concurrency` and `per_model_max_concurrency` as static values and doesn't mention adaptive concurrency. After this PR:

- The recommended config block should include `min_concurrency`, `backoff_factor`, and `additive_increase`
- The "Key Considerations" section should explain that AIMD provides a second layer of backpressure response (reducing concurrency) on top of retry backoff, and how it interacts with GIE flow control shedding

Can be a follow-up PR.

j-mok-dev: Thanks for the suggestion. I’ve updated the flow-control guide in this PR.

Changes made in `docs/guides/flow-control-setup.md`:
- Added AIMD-related processor settings to the recommended config block:
  - `min_concurrency`
  - `backoff_factor`
  - `additive_increase`
- Expanded **Key Considerations** to explain:
  - how retry backoff and AIMD play different roles
  - how processor-side AIMD interacts with GIE flow-control shedding
  - that `global_concurrency` remains a hard global ceiling while AIMD adapts per endpoint

Please take another look. Thanks!

internal/processor/worker/executor_test.go
```diff
@@ -3135,3 +3136,296 @@ func TestExecuteOneRequest_PerModelInferenceObjective(t *testing.T) {
...
+			generateFn: func(_ context.Context, _ *inference.GenerateRequest) (*inference.GenerateResponse, *inference.ClientError) {
+				return nil, &inference.ClientError{
+					Category: httpclient.ErrCategoryServer,
+					Message:  "connection refused",
+					RawError: errors.New("dial tcp: connection refused"),
+				}
+			},
+		}
+
+		p := runProcessModel(t, cfg, mock, makeRequests(1))
```
lioraron: This test doesn't verify the documented "skip" semantics — it only verifies that non-HTTP errors don't trigger `RecordRateLimit` (no decrease).

The AIMD starts at `maxLimit` (20) which equals `MaxLimit` (20), so `RecordSuccess` at the ceiling is a no-op: the limit stays at 20 whether AIMD is skipped entirely or incorrectly called with `RecordSuccess`. With only 1 request and a window size of 20, even if the code were buggy and counted non-HTTP errors as successes, the test would still pass.

This matters because false `RecordSuccess` calls would prematurely recover concurrency after a real backoff (e.g., after a 429 reduces the limit from 20 to 10, spurious successes from connection-refused errors would fill the window and push the limit back up).

To test the skip-vs-success distinction, the AIMD needs to start below `MaxLimit`. One approach (similar to `TestProcessModel_AIMDEndpointIsolation`): drive `processModel` directly, call `epLimit.aimd.RecordRateLimit()` first to reduce the limit, then process non-HTTP errors and verify the limit didn't increase back.

lioraron (commented):
## Proposal: group concurrency config into a structured object

The concurrency-related params are currently flat top-level fields in `ProcessorConfig`. Since there are no users yet (no backward-compat concern), this is a good time to restructure them into a dedicated config object with an explicit AIMD toggle.

### Proposed YAML structure

```yaml
concurrency:
  global: 100
  per_endpoint: 10
  recovery: 5
  aimd:
    enabled: true
    min: 5
    backoff_factor: 0.5
    additive_increase: 1
```

### Why

1. **Clear grouping** — all dispatch-rate and concurrency knobs live under one key instead of being scattered across the top level.
2. **Explicit AIMD toggle** — `aimd.enabled: false` is obvious and self-documenting. No need to reason about setting `min = per_endpoint` to disable adaptive behavior.
3. **Rename opportunity** — `per_model_max_concurrency` → `per_endpoint` since this PR already documents that the name is misleading (models sharing the same gateway endpoint share one semaphore).
4. **When `aimd.enabled: false`**, the AIMD params (`min`, `backoff_factor`, `additive_increase`) are ignored entirely — less room for misconfiguration.

### Scope

- `ProcessorConfig` struct: new `ConcurrencyConfig` with nested `AIMDConfig`
- `config.yaml`, `values.yaml`, `processor-configmap.yaml`: restructured YAML
- `config_test.go`: updated validation tests
- `worker.go`, `executor.go`: read from nested config
- Docs: `batch_processor_architecture.md`, `flow-control-setup.md` updated accordingly

pkg/clients/inference/inference_client_resolver.go
```diff
@@ -135,6 +141,35 @@ func (r *GatewayResolver) ClientFor(modelID string) InferenceClient {
 	return nil
 }
 
+// Clients returns the deduplicated set of InferenceClient instances managed by
+// this resolver. In global mode this is a single-element slice; in per-model
+// mode it reflects the pooled clients (models sharing identical gateway configs
+// share a single InferenceClient).
+func (r *GatewayResolver) Clients() []InferenceClient {
```
lioraron, line 148: `Clients()` and `ClientLabel()` are new public methods with no unit tests. `Clients()` has its own deduplication loop that could diverge from `ClientFor`'s pooling. Missing coverage:

- `Clients()` on a global resolver returns exactly one element
- `Clients()` on a per-model resolver with shared configs returns deduplicated set
- `ClientLabel()` returns the URL for a known client
- `ClientLabel()` returns `"unknown"` for an unrecognized client

Also: `NewSingleClientResolver` (line 178) and `NewPerModelClientResolver` (line 185) don't populate `clientURLs`, so `ClientLabel()` always returns `"unknown"` for clients created via these constructors. The latter is test-only, but `NewSingleClientResolver` is also used in production test paths — if it ever appears in a real deployment, all AIMD log messages would show `endpoint=unknown`.

internal/processor/worker/executor.go
```diff
@@ -403,8 +408,20 @@ func (p *Processor) processModel(
...
-	// PerModelMaxConcurrency > 0 is enforced by config validation; this cannot fail.
-	modelSem, _ := semaphore.New(p.cfg.PerModelMaxConcurrency, p.guardCallback)
+	// Resolve the per-endpoint adaptive semaphore and AIMD controller for this
+	// model. Models sharing the same inference endpoint share the same pair.
+	// ClientFor can return nil after gateway config changes between ingestion and
+	// execution, or during recovery when model_map/plan files predate the current
+	// resolver. In that case, drain all entries as model_not_found.
+	client := p.inference.ClientFor(modelID)
+	epLimit := p.endpointLimits[client]
+	if epLimit == nil {
```
lioraron, line 419: This defensive drain path is untested. It handles the case where `ClientFor` returns nil (or a client not in `endpointLimits`) — e.g., during recovery when plan files predate the current resolver config.

A test should exercise this by setting up a processor whose `endpointLimits` map doesn't contain the model's client, then verifying all entries are drained as `model_not_found`.

pkg/clients/http/http_client_test.go
```diff
@@ -1187,6 +1190,107 @@ func containsString(s, substr string) bool {
 	return false
 }
 
+func TestCapacityRetryTracking(t *testing.T) {
```
lioraron, line 1193: Missing test: 5xx retry sets the capacity flag. The retry hook sets the flag for both `StatusTooManyRequests` and `>= StatusInternalServerError`, but only 429 is tested here. A 502→200 sequence would close this gap and catch regressions if someone narrows the condition to 429-only.

internal/processor/config/config_test.go
```diff
@@ -414,6 +414,80 @@ func TestProcessorConfig_Validate_MinimumValueChecks(t *testing.T) {
 	}
 }
 
+func TestProcessorConfig_Validate_ConcurrencyAIMD(t *testing.T) {
```
lioraron, line 423: Missing negative-value validation tests for `MinConcurrency` and `AdditiveIncrease`. The validation guards exist (`MinConcurrency <= 0` and `AdditiveIncrease <= 0`), but unlike `BackoffFactor` (which has an explicit `zero backoff_factor rejected` test), these guards are never exercised. Add subtests for `MinConcurrency = 0` and `AdditiveIncrease = 0`.

internal/processor/worker/executor_test.go
```diff
@@ -3135,3 +3136,296 @@ func TestExecuteOneRequest_PerModelInferenceObjective(t *testing.T) {
 		})
 	}
 }
+
+// =====================================================================
+// Tests: AIMD signaling in processModel
+// =====================================================================
+
+func TestProcessModel_AIMDSignaling(t *testing.T) {
```
lioraron, line 3146: Missing test case: 4xx (not 429) response records `RecordSuccess`. The AIMD signal comment documents `4xx (not 429) → RecordSuccess (gateway had capacity, request was malformed)` but this path is never tested at the executor level. A subtest where the mock returns `StatusCode: 400` would verify the `default` branch handles non-capacity HTTP errors correctly.

internal/processor/worker/executor.go
```diff
@@ -422,25 +439,51 @@ dispatch:
...
+			//   4xx (not 429) → RecordSuccess (gateway had capacity, request was malformed)
+			//   non-HTTP err → skip (no capacity signal — network, timeout, etc.)
+			//   fatal execErr → skip (local I/O, not related to gateway capacity)
+			//
+			// AIMD only affects future dispatch. It does not abort in-flight
+			// requests — those continue until completion or context cancellation.
+			if execErr == nil && result != nil && result.Response != nil {
+				sc := result.Response.StatusCode
+				switch {
+				case sc == 429 || sc >= 500:
```
lioraron: Use `net/http` constants instead of hardcoded numbers — the retry hook in `http_client.go` already uses these:
```go
case sc == http.StatusTooManyRequests || sc >= http.StatusInternalServerError:
```

lioraron:
## ℹ️ Resizable semaphore — third-party library analysis (no action required)

Evaluated existing Go packages for resizable semaphores to see if the custom `AdaptiveSemaphore` could be replaced:

| | `marusama/semaphore/v2` | `tlkamp/go-semaphore` | `sherifabdlnaby/semaphore` |
|---|---|---|---|
| Stars | 178 | 1 | 3 |
| Last commit | 2021 (stable, 0 issues) | Feb 2025 | 2019 (abandoned) |
| Impl | CAS on packed uint64 | Wraps EOL `eapache/channels` | Fork of `x/sync/semaphore` |
| External deps | **Zero** | 2 runtime (one EOL) | Zero (but no go.mod) |
| SetLimit under contention | ✅ Correct | ✅ Correct | ❌ Bug: Resize + ctx cancel leaks waiters permanently |
| Double-release handling | Panics (no callback) | Silent corruption | Panics (no callback) |
| TryAcquire | Non-blocking CAS | TOCTOU race with 1ms sleep | Non-blocking |

**`marusama/semaphore/v2`** is the only viable candidate, but it **panics on double-release** with no callback mechanism. The custom `AdaptiveSemaphore` in this PR uses a `sync.Once`-guarded `onDoubleRelease` callback — a feature none of these libraries support.

The other two are non-starters: `tlkamp` depends on an EOL library with a racy `TryAcquire`, and `sherifabdlnaby` has a [confirmed bug](https://github.com/sherifabdlnaby/semaphore/issues/1) in Resize + context cancellation (the exact feature needed here).

**Conclusion:** The custom `AdaptiveSemaphore` (~150 lines, zero deps) is justified. It's small, handles the double-release callback cleanly, and avoids the gaps in available third-party options.

lioraron:
## ℹ️ AIMD controller — third-party library analysis (no action required)

Evaluated existing Go packages that implement AIMD concurrency control to see if the custom `AIMDController` could be replaced:

| | `go-concurrency-limits` (`AIMDLimit`) | `goresilience` (`NewAIMD`) | `failsafe-go` |
|---|---|---|---|
| Algorithm | AIMD (closest match) | AIMD + TCP slow-start | Latency-trend (**not AIMD**) |
| Limit change callback | ✅ `NotifyOnChange` | ❌ Return value only | ✅ But wrong algorithm |
| Configurable min/max | ❌ Min hardcoded to 1, no max | Min only, no max | Yes, but irrelevant |
| Success trigger | inflight >= limit | inflight × factor > limit | Latency trend |
| Maintained | ✅ Active (May 2026) | ❌ Abandoned (2021) | ✅ Active |
| Dependencies | Heavy (grpc, protobuf, datadog, prometheus) | Light but ancient versions | Medium |

The closest candidate is [`platinummonkey/go-concurrency-limits`](https://pkg.go.dev/github.com/platinummonkey/go-concurrency-limits) (Netflix port, 143 stars, actively maintained). Its `AIMDLimit` has a change callback (`NotifyOnChange`), but:

1. **No configurable min/max limits** — min is hardcoded to 1, no max cap
2. **Different increase trigger** — fires when `inflight >= limit`, not after a window of consecutive successes
3. **Heavy dependency tree** — pulls in grpc, protobuf, datadog, prometheus even if you only use the `limit` package

To use it, you'd need a wrapper enforcing min/max and accepting different trigger semantics — at which point the wrapper would be more code than the ~80-line custom `AIMDController`.

**Conclusion:** Same as the [semaphore analysis](https://github.com/llm-d-incubation/batch-gateway/pull/394#issuecomment-4399354943) — the custom implementation is justified. It's small (~80 lines), zero-dep, and matches the exact semantics needed (consecutive-success window, explicit 429/5xx signals, configurable min/max, `setFn` callback).

j-mok-dev:
> ## Proposal: group concurrency config into a structured object
> The concurrency-related params are currently flat top-level fields in `ProcessorConfig`. Since there are no users yet (no backward-compat concern), this is a good time to restructure them into a dedicated config object with an explicit AIMD toggle.
> 
> ### Proposed YAML structure
> ```yaml
> concurrency:
>   global: 100
>   per_endpoint: 10
>   recovery: 5
>   aimd:
>     enabled: true
>     min: 5
>     backoff_factor: 0.5
>     additive_increase: 1
> ```
> 
> ### Why
> 1. **Clear grouping** — all dispatch-rate and concurrency knobs live under one key instead of being scattered across the top level.
> 2. **Explicit AIMD toggle** — `aimd.enabled: false` is obvious and self-documenting. No need to reason about setting `min = per_endpoint` to disable adaptive behavior.
> 3. **Rename opportunity** — `per_model_max_concurrency` → `per_endpoint` since this PR already documents that the name is misleading (models sharing the same gateway endpoint share one semaphore).
> 4. **When `aimd.enabled: false`**, the AIMD params (`min`, `backoff_factor`, `additive_increase`) are ignored entirely — less room for misconfiguration.
> 
> ### Scope
> * `ProcessorConfig` struct: new `ConcurrencyConfig` with nested `AIMDConfig`
> * `config.yaml`, `values.yaml`, `processor-configmap.yaml`: restructured YAML
> * `config_test.go`: updated validation tests
> * `worker.go`, `executor.go`: read from nested config
> * Docs: `batch_processor_architecture.md`, `flow-control-setup.md` updated accordingly


Restructured as proposed:
- `ProcessorConfig` → `ConcurrencyConfig` with nested `AIMDConfig`
- Renamed `per_model_max_concurrency` → `per_endpoint`
- Added `aimd.enabled` toggle; when `false`, AIMD params are ignored and validation is skipped
- `executor.go` guards all AIMD calls with `if epLimit.aimd != nil`
- Updated: `config.yaml`, `values.yaml`, `processor-configmap.yaml`, `config_test.go`, `worker.go`, `recovery.go`, `test_helpers_test.go`, `flow-control-setup.md`, `dev-deploy.sh`

internal/processor/worker/executor_test.go
j-mok-dev: The test now calls `RecordRateLimit()` first to reduce the limit to `maxLimit/2` (10), then sends exactly that many non-HTTP errors through `processModel`. If the code incorrectly called `RecordSuccess`, the window would fill and push the limit back up — but we assert it stays at the reduced value.
This distinguishes "skip" (limit unchanged) from "false RecordSuccess" (limit recovers), which the original test at ceiling couldn't detect.

pkg/clients/inference/inference_client_resolver.go
j-mok-dev: **Tests added** (`inference_client_resolver_test.go`):
- `TestGatewayResolver_Clients/global_resolver_returns_single_element`
- `TestGatewayResolver_Clients/per-model_resolver_with_shared_configs_returns_deduplicated_set`
- `TestGatewayResolver_ClientLabel/returns_URL_for_known_client`
- `TestGatewayResolver_ClientLabel/returns_unknown_for_unrecognized_client`
- `TestGatewayResolver_ClientLabel/NewSingleClientResolver_populates_label`
- `TestGatewayResolver_ClientLabel/NewPerModelClientResolver_populates_labels`

**Fix**: Both constructors now populate `clientURLs` with synthetic labels (`"test"` / `"test-<model>"`) so `ClientLabel()` returns a distinguishable string instead of `"unknown"`. This is sufficient for test log readability — production resolvers already use the real endpoint URL.

internal/processor/worker/executor.go
j-mok-dev: Added `TestProcessModel_EndpointLimitNil_DrainsAsModelNotFound`

pkg/clients/http/http_client_test.go
j-mok-dev: Thanks, added the subtest `"5xx retry sets capacity flag"`.

internal/processor/config/config_test.go
j-mok-dev: Thanks. added `TestProcessorConfig_Validate_ConcurrencyAIMD`

internal/processor/worker/executor_test.go
j-mok-dev: Added subtest `"4xx (not 429) records success"`

internal/processor/worker/executor.go
j-mok-dev: done. Thanks!

charts/batch-gateway/values.yaml
```diff
@@ -303,10 +303,15 @@ processor:
     taskWaitTime: "1s"
     pollInterval: "5s"
     numWorkers: 20
-    globalConcurrency: 100
-    perModelMaxConcurrency: 10
-    # Maximum number of concurrent job recoveries during startup
-    recoveryMaxConcurrency: 5
+    concurrency:
+      global: 100            # Fixed ceiling for total in-flight requests
```
lioraron: **Suggestion:** When AIMD is enabled, per-endpoint limits self-regulate via backpressure, so the global limit mostly acts as a burst ceiling. The default of 100 is fine for small deployments but could become a bottleneck with many endpoints.

Consider expanding the inline comment to guide sizing:
```yaml
# Fixed ceiling for total in-flight requests across all endpoints.
# When AIMD is enabled, per-endpoint limits self-regulate — set this
# high enough to avoid being the first bottleneck
# (e.g., perEndpoint * expected_endpoint_count * 2).
global: 100
```

And a similar note in the flow control guide's "Key Considerations" section.

charts/batch-gateway/values.yaml
```diff
@@ -303,10 +303,15 @@ processor:
...
+      perEndpoint: 10        # Initial/max concurrency per inference endpoint
+      recovery: 5            # Max concurrent job recoveries during startup
+      aimd:
+        enabled: true        # Set to false to disable adaptive concurrency
+        min: 5               # AIMD floor per endpoint
```
lioraron: The default AIMD range [5, 10] is very narrow — a single 429 halves the limit straight to the floor.

**`perEndpoint`**: The right value depends on what sits behind the endpoint URL:
- **Single vLLM instance** → 20 is reasonable
- **GIE/EPP pool routing to N replicas** → could be much higher (e.g., `20 * N`), since the pool absorbs more concurrency

With AIMD, starting high is safer than starting low — AIMD backs off quickly on 429s but recovers slowly at +1 per window. Starting too low means underutilizing the backend until AIMD crawls up.

Add a doc note explaining how to size `perEndpoint` based on backend topology.

**`min`**: It's not clear what drives the default of 5 — it should reflect the minimum request rate you want to sustain to an endpoint under heavy backpressure, balancing:
- **Too high** → AIMD can't back off enough when the backend is genuinely overloaded
- **Too low** → a few 429s starve the endpoint, and recovery is very slow at `additiveIncrease: 1` per window

Add a comment explaining how operators should tune `min` relative to `perEndpoint` and their expected backend capacity.

docs/guides/flow-control-setup.md
```diff
@@ -196,8 +203,9 @@ model_gateways:
...
+- **`max_retries` and `max_backoff`**: When the system is saturated, GIE sheds batch requests with HTTP 429. Retry backoff slows resubmission pressure, while AIMD separately lowers per-endpoint concurrency (`aimd.backoff_factor`) and later raises it gradually (`aimd.additive_increase`) as successes accumulate.
+- **AIMD with flow control**: Flow control decides queueing/shedding priority in GIE, and AIMD controls how aggressively the processor feeds each endpoint. Together they provide two layers of backpressure response: Router-side admission/shedding plus processor-side concurrency adaptation. Set `aimd.enabled: false` to use fixed concurrency without adaptive behavior.
+- **`concurrency.global`**: This remains a hard global ceiling across endpoints. Keep it high enough to avoid becoming the first bottleneck, while per-endpoint AIMD reacts to localized overload.
```
lioraron: The sizing guidance requested in the `values.yaml` comments should also be added here — specifically:

- How to size `per_endpoint` based on backend topology (single instance vs. GIE pool with N replicas)
- How to size `concurrency.global` relative to endpoint count when AIMD is enabled
- How to choose `aimd.min` relative to `per_endpoint`

This is the guide operators will reference when configuring flow control + batch-gateway together.

internal/processor/metrics/metrics.go
```diff
@@ -159,7 +159,7 @@ func InitMetrics(cfg config.ProcessorConfig) error {
 			Help: "Configured maximum number of concurrent in-flight inference requests (GlobalConcurrency)",
 		},
 	)
-	processorMaxInflightConc.Set(float64(cfg.GlobalConcurrency))
+	processorMaxInflightConc.Set(float64(cfg.Concurrency.Global))
```
lioraron, line 165: AIMD is being introduced without any observability into its behavior. Since you're already revising this PR, consider adding metrics here:

| Metric | Type | Labels | Description |
|--------|------|--------|-------------|
| `batch_processor_aimd_concurrency_limit` | Gauge | `endpoint` | Current effective concurrency limit per endpoint |
| `batch_processor_aimd_decreases_total` | Counter | `endpoint` | Multiplicative decreases (429/5xx signals) |
| `batch_processor_aimd_increases_total` | Counter | `endpoint` | Additive increases (sustained success windows) |

The gauge is the most important — without it, operators have no way to see how AIMD is reacting to backpressure. It also unblocks AIMD e2e tests (#420) which need observable metrics to verify behavior.

See #419 for the full scope.

lioraron (commented):
Thanks!

charts/batch-gateway/values.yaml
j-mok-dev: thanks. added!

charts/batch-gateway/values.yaml
j-mok-dev: Thanks. added!

docs/guides/flow-control-setup.md
j-mok-dev: Thanks. added information about how to set the values in the guide doc.

internal/processor/metrics/metrics.go
j-mok-dev: Added all three metrics from this comment (and #419):
- `batch_processor_aimd_concurrency_limit` (gauge, by endpoint)
- `batch_processor_aimd_decreases_total` (counter, by endpoint + signal)
- `batch_processor_aimd_increases_total` (counter, by endpoint)

I had originally held off on endpoint-labeled metrics due to cardinality concerns. endpoints are config-driven but unbounded in principle. Since these are needed for AIMD e2e observability (#420), I've added them with a startup warning when the number of endpoints exceeds 50 (which would create ~250 time series).

Note: `decreases_total` counts backpressure *signals* (429, 5xx, capacity_retry), not actual limit changes — the counter still increments when AIMD is already at the floor. This is intentional: operators need visibility into sustained backpressure even when the limit can't drop further. The Help text documents this.

Full-fidelity per-event AIMD logging (with reason) is also in place for detailed diagnostics.

lioraron (approved):
Great work!

PR #396, opened 2026-04-28
docs: remove MaaS and RHOAI deployment guides and scripts

yizhaodev:
## Why is this PR needed?

Remove midstream related docs/scripts from upstream

## What does this PR do?

Remove midstream related docs/scripts from upstream

## How was this tested?

- [ ] Unit tests added/updated/verified
- [ ] Integration/e2e tests added/updated/verified
- [ ] Manual testing performed

## Checklist

- [x] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [x] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [ ] CI checks pass (`make ci`)
- [ ] E2E tests pass (`make test-e2e`)

PR #397, opened 2026-04-29
refactor: unify image tag resolution

yizhaodev:
## Why is this PR needed?

Image tag logic was duplicated and `BATCH_DEV_VERSION=local` hardcoded `latest` instead of using chart defaults. RHOAI script had hardcoded image overrides now baked into `values.yaml`.

## What does this PR do?

- Extract `get_batch_gateway_image_tag()` to deduplicate tag resolution
- Fall back to chart `values.yaml` defaults when no explicit tag is set
- Remove RHOAI hardcoded image repo/tag overrides
- Default `MODEL_REPLICAS` to 1

## How was this tested?

- [x] Full uninstall → install → test on OpenShift (RHOAI 3.4 beta), 13/13 tests passed

### DIFF
--- a/examples/deploy-demo/common.sh
+++ b/examples/deploy-demo/common.sh
@@ -556,6 +556,21 @@ EOF
     log "Secret '${BATCH_APP_SECRET_NAME}' applied."
 }
 
+get_batch_gateway_image_tag() {
+    local image_tag=""
+    if [ -n "${BATCH_IMAGE_TAG}" ]; then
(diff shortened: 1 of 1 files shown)


j-mok-dev:
Looks like the RHOAI image overrides and MODEL_REPLICAS changes mentioned in this PR description were previously handled by #396?

j-mok-dev (approved):

lioraron (approved):

PR #403, opened 2026-05-01
feat: send x-gateway-inference-fairness-id header for per-tenant fairness

### LINKED ISSUE #401: [flow-control] Support x-gateway-inference-fairness-id header for per-tenant fairness
The processor doesn't currently send the `x-gateway-inference-fairness-id` header — all batch requests share GIE's default flow (`"default-flow"`). For multi-tenant batch platforms using `round-robin` fairness in the batch priority band, the processor should send this header with the tenant identifier from each job, enabling GIE to enforce per-tenant fair scheduling within the batch band.

**Context:** See the [flow control setup guide](https://github.com/llm-d-incubation/batch-gateway/blob/main/docs/guides/flow-control-setup.md#headers-sent) for background on how Batch Gateway integrates with GIE flow control.

j-mok-dev:
## Why is this PR needed?

The processor doesn't currently send the `x-gateway-inference-fairness-id` header — all batch requests share GIE's default flow (`"default-flow"`). For multi-tenant batch platforms using `round-robin` fairness in the batch priority band, the processor should send this header with the tenant identifier from each job, enabling GIE to enforce per-tenant fair scheduling within the batch band.

## What does this PR do?

Threads `jobInfo.TenantID` through the execution path (`executeJob` → `processModel` → `executeOneRequest` → `mergeInferenceHeaders`) and sets the `x-gateway-inference-fairness-id` header when the tenant ID is non-empty.

- Adds `fairnessIDHeader` constant and extends `mergeInferenceHeaders` with a `fairnessID` parameter.
- No config change required — the header is automatically sent when `TenantID` is present on the job (which is always the case for jobs created through the API, since middleware defaults to `"default"`).
- Header is omitted for empty tenant ID

## How was this tested?

  - Fairness header present when tenantID is non-empty
  - Fairness header absent when tenantID is empty
  - All three processor-managed headers set simultaneously
  - Processor-managed headers override conflicting pass-through headers
- [x] Manual testing performed (pre-commit + dev-deploy + e2e)

## Checklist

## Related Issues

Fixes #401

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -267,6 +267,8 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
 	// from sloCtx (derived from context.Background), so SLO expiry and SIGTERM do
 	// not set it. processModel's drain phase checks sloCtx.Err() vs
 	// userCancelCtx.Err() to choose the right error code (errExpired vs errCancelled).
+	tenantID := params.jobInfo.TenantID
+
 	for safeModelID, modelID := range modelMap.SafeToModel {
 		// Ordering guarantee: processModel returns → requestAbortFn → errCh send.
 		// This ensures the first real error reaches errCh before any context.Canceled
@@ -282,6 +284,7 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
 				writers,
 				progress,
 				passThroughHeaders,
+				tenantID,
 			)
 			// Abort all sibling models when any model hits a fatal I/O error
 			// (e.g. output file write failure). modelErr is only set for local
@@ -387,6 +390,7 @@ func (p *Processor) processModel(
 	writers *outputWriters,
 	progress *executionProgress,
 	passThroughHeaders map[string]string,
+	tenantID string,
 ) error {
 	logger := logr.FromContextOrDiscard(requestAbortCtx).WithValues("model", modelID)
 	requestAbortCtx = logr.NewContext(requestAbortCtx, logger)
@@ -436,7 +440,7 @@ dispatch:
 			defer modelSem.Release()
 			defer p.globalSem.Release()
 
-			result, execErr := p.executeOneRequest(requestAbortCtx, sloCtx, inputFile, entry, modelID, passThroughHeaders)
+			result, execErr := p.executeOneRequest(requestAbortCtx, sloCtx, inputFile, entry, modelID, passThroughHeaders, tenantID)
 			if execErr != nil {
 				// Fatal read failure: the input file is unreadable at this offset
 				// (e.g. disk corruption). We do not know the CustomID, so we cannot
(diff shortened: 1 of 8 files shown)


yizhaodev (commented):
If `passThroughHeaders` already contains `x-gateway-inference-fairness-id`, `mergeInferenceHeaders` silently overwrites it with `tenantID`. We should honor the user-supplied pass-through value — only fall back to `tenantID` when the header is absent.

internal/processor/worker/executor.go
```diff
@@ -621,17 +625,21 @@ func (p *Processor) drainUnprocessedRequests(
...
 //   - x-gateway-inference-objective: name of the InferenceObjective CRD that
 //     determines the priority band for this request.
+//   - x-gateway-inference-fairness-id: tenant identifier for per-tenant fairness
+//     within a priority band.
 //
 // Headers are only added when the relevant value is available/configured.
 // If sloCtx has no deadline, is cancelled, or has an expired deadline, the SLO
 // header is not set. If inferenceObjective is empty, the objective header is not set.
-func mergeInferenceHeaders(headers map[string]string, sloCtx context.Context, inferenceObjective string) map[string]string {
+// If fairnessID is empty, the fairness header is not set.
```
yizhaodev: If `passThroughHeaders` already contains `x-gateway-inference-fairness-id`, this overwrites the user-supplied value. We should honor pass-through — only fall back to `tenantID` when the header is absent in `passThroughHeaders`.

internal/processor/worker/executor_test.go
```diff
@@ -2907,12 +2907,72 @@ func TestMergeInferenceHeaders(t *testing.T) {
...
+	})
+
+	t.Run("fairness header with default tenant", func(t *testing.T) {
+		h := mergeInferenceHeaders(nil, context.Background(), "", "default")
+		if h[fairnessIDHeader] != "default" {
+			t.Fatalf("fairness header: got %q, want %q", h[fairnessIDHeader], "default")
+		}
+	})
+
+	t.Run("processor-managed headers override pass-through conflicts", func(t *testing.T) {
```
yizhaodev: I think we should honor `pass-through`, 
if user want to use apikey or group id as flow key, they can pass it as fair id header.

j-mok-dev:
Thanks @yizhaodev! updated to honor pass-through for x-gateway-inference-fairness-id.

Now mergeInferenceHeaders only sets the fairness header from tenantID as a fallback when the outgoing headers do not already include x-gateway-inference-fairness-id. If pass-through already provides it, we keep that value unchanged.

Also updated unit tests to cover this precedence behavior.

yizhaodev (approved):

lioraron (commented):
The `x-gateway-inference-fairness-id` header will be sent unconditionally on every inference request, because `TenantID` is never empty — it defaults to `"default"` via `common.DefaultTenantID` when no tenant header is configured.

This is inconsistent with the other flow control headers:
- `x-slo-ttft-ms` — only sent when the batch request includes a `ttft_slo_ms` value
- `x-gateway-inference-objective` — only sent when `inference_objective` is configured in processor config

For deployments that don't use multi-tenancy (or GIE fairness policies), this means every request will carry a `x-gateway-inference-fairness-id: default` header with no semantic value.

**Suggestion**: Add a processor config param (e.g., `send_fairness_header: true`) to make this opt-in, consistent with how `inference_objective` works. When disabled (the default), the header is omitted entirely.

j-mok-dev:
> The `x-gateway-inference-fairness-id` header will be sent unconditionally on every inference request, because `TenantID` is never empty — it defaults to `"default"` via `common.DefaultTenantID` when no tenant header is configured.
> 
> This is inconsistent with the other flow control headers:
> 
> * `x-slo-ttft-ms` — only sent when the batch request includes a `ttft_slo_ms` value
> * `x-gateway-inference-objective` — only sent when `inference_objective` is configured in processor config
> 
> For deployments that don't use multi-tenancy (or GIE fairness policies), this means every request will carry a `x-gateway-inference-fairness-id: default` header with no semantic value.
> 
> **Suggestion**: Add a processor config param (e.g., `send_fairness_header: true`) to make this opt-in, consistent with how `inference_objective` works. When disabled (the default), the header is omitted entirely.

Thanks @lioraron 
I changed fairness header emission to be opt-in via processor config:
- `send_fairness_header` (default: false)
Behaviour now:
- default/false: `x-gateway-inference-fairness-id` is omitted entirely
- true: processor sends fairness header using tenant ID as fallback
- if pass-through already provides `x-gateway-inference-fairness-id`, we keep the pass-through value (no overwrite)

lioraron (commented):
## Review

### 1. Inconsistent pass-through precedence — needs a "why" comment

`mergeInferenceHeaders` now has asymmetric behavior: SLO and objective headers always override pass-through values, but fairness honors existing pass-through values (skips if already set). The doc comment describes the *what* but not the *why*:

```go
// If fairnessID is non-empty, the fairness header is only set when the outgoing
// headers do not already include x-gateway-inference-fairness-id.
```

A future contributor could see fairness not overriding as a bug and "fix" it to match SLO/objective behavior. Add a short comment explaining the design intent — e.g., fairness is user-overridable (callers can set tenant identity via pass-through), while SLO and objective are processor-authoritative.

### 2. `SendFairnessHeader` config gating is untested in `executeOneRequest`

Every `executeOneRequest` test passes `""` as tenantID and none set `SendFairnessHeader: true` on the processor config. The gating logic:

```go
fairnessID := ""
if p.cfg.SendFairnessHeader {
    fairnessID = tenantID
}
```

is never exercised by any test. If someone removes or inverts the condition, no test fails. Add at least two `executeOneRequest` test cases:
- `SendFairnessHeader=true` + non-empty tenantID → header present in request
- `SendFairnessHeader=false` + non-empty tenantID → header absent

lioraron (approved):
Thanks!
Pls resolve the conflicts when available, and feel free to merge.

Issue #402, opened 2026-04-30
[flow-control] Add integration test for flow control with GIE priority bands and shedding


lioraron:
## Summary

There is currently no integration test that validates the end-to-end flow control integration between Batch Gateway and GIE. The [flow control setup guide](https://github.com/llm-d-incubation/batch-gateway/blob/main/docs/guides/flow-control-setup.md) describes a recommended configuration with sheddable batch priority, SLO-deadline ordering, and saturation-based shedding, but none of this is tested in CI.

> **Note:** Simpler flow control tests (header propagation, 429 retry) that don't require GIE are tracked separately in #405.

## What to test

1. **Priority band interaction**: Batch requests (priority -1) are dispatched after interactive requests (priority 0 or 100) and are shed during saturation.
2. **SLO-deadline ordering**: Batch requests closer to their SLO deadline are dispatched before newer requests within the batch band.
3. **Shedding under saturation**: When the model server is saturated, GIE sheds batch-priority requests (429) while interactive requests continue to be served.
4. **Batch completion under mixed load**: Batch jobs eventually complete once saturation subsides, with correct retry and backoff behavior.

## Dev-deploy: composable modes

The current `dev-deploy.sh` deploys batch-gateway with direct vLLM simulator access. This issue adds GIE as an opt-in layer. The dev-deploy design uses **composable, independently toggleable flags** so that additional components (like the async processor) can be added the same way in the future.

### Deployment topologies

Each flag adds a component to the pipeline. Flags compose independently:

```
Default:                    Client → Batch Gateway → vLLM Simulator
ENABLE_GIE=true:            Client → Batch Gateway → GIE/EPP → vLLM Simulator
ENABLE_ASYNC=true:          Client → Batch Gateway → Async Processor → vLLM Simulator
Both:                       Client → Batch Gateway → Async Processor → GIE/EPP → vLLM Simulator
```

Each flag gates its own `install_*` functions and adjusts upstream URL wiring — `processor.config.globalInferenceGateway.url` points to whichever component is next in the chain. The default (no flags) remains fast and simple for everyday batch-gateway development.

> **Future:** `ENABLE_ASYNC=true` will be implemented when async processor integration is ready. The pattern established here (flag → install functions → URL rewiring) applies directly.

### 1. `scripts/dev-deploy.sh` — Add GIE deployment functions

Add these functions, gated behind `ENABLE_GIE=true`:

[...]

PR #408, opened 2026-05-04
feat(e2e): add flow control e2e tests for header propagation and 429 …

### LINKED ISSUE #405: [flow-control] Add e2e tests for header propagation and 429 retry behavior
## Summary

Add e2e tests that validate the processor's flow control behavior without requiring a full GIE/EPP deployment. These tests cover batch-gateway's own responsibilities — sending the right headers and handling 429 responses — and can run in the existing Kind dev-deploy setup.

Split from #402, which tracks the full GIE integration tests (priority band interaction, shedding under saturation).

### 1. Header propagation
When `inferenceObjective` is configured, the processor should send:
- `x-gateway-inference-objective` with the configured objective name
- `x-slo-ttft-ms` with the remaining milliseconds until the batch SLO deadline

**Approach:** Add a mode to the vLLM simulator (or use a lightweight HTTP mock) that captures and exposes received headers, so the test can assert on the exact header values sent by the processor.

### 2. 429 retry behavior
When the inference server returns HTTP 429 (shed/rate-limited), the processor should:
- Retry with exponential backoff
- Eventually complete the batch if the server recovers
- Fail gracefully if retries are exhausted before the SLO deadline

**Approach:** Configure the mock inference server to return 429 for a configurable number of requests before succeeding, then verify batch completion and retry metrics.

## Out of scope

- Priority band interaction between batch and interactive requests (requires GIE — see #402)
- Saturation-based shedding with real EPP flow control (requires GIE — see #402)

## Related

- #402 — Full GIE integration tests (parent issue)
- [Flow control setup guide](https://github.com/llm-d-incubation/batch-gateway/blob/main/docs/guides/flow-control-setup.md)

j-mok-dev:
## What does this PR do?

Adds `flow_control_test.go` with two test groups:

**HeaderAndRetry** (runs in all CI matrix combinations):
- `InferenceObjectiveHeader`: validates processor ConfigMap has the correct `inferenceObjective` for the test model, submits a batch and verifies completion
- `SLOHeader`: submits a batch with `completion_window=10m` and verifies it completes within the SLO window
- `RetryOn429`: targets a 50% failure-rate simulator (10 requests), verifies the batch completes after retries with zero failed requests, and asserts retry occurrence via processor logs
- `RetryExhaustion`: targets a 100% failure-rate simulator (`sim-model-always-fail`, `maxRetries=1`), verifies the batch reaches `completed` status with `Completed==0`, `Failed>0`, output file contains 429 responses, and `request_errors_by_model_total` metric is recorded

**GIE** (skipped when GIE is not deployed):
- `HeaderPropagation`: verifies requests route through EPP with evidence in EPP logs
- `BatchCompletionThroughEPP`: verifies multi-model batch completion through separate per-model EPP instances

> **Note:** Direct e2e header value assertion (`x-gateway-inference-objective`, `x-slo-ttft-ms`) is deferred until `llm-d-inference-sim` releases `--log-http` support ([merged to main](https://github.com/llm-d/llm-d-inference-sim/pull/470) on 2026-05-03, not yet in a release tag). TODO comments track this in the test file. Header correctness is covered by unit tests in `executor_test.go`.

## How was this tested?

- [x] Integration/e2e tests added/updated/verified (dev-deploy in progress)

## Checklist

## Related Issues

Closes #405
Ref #402


scripts/dev-deploy.sh
```diff
@@ -42,10 +44,24 @@ GC_IMG="${GC_IMG:-ghcr.io/llm-d-incubation/batch-gateway-gc:${IMAGE_TAG}}"
...
+#   EPP Helm release:      ${GIE_EPP_RELEASE}-${model}          e.g. epp-sim-model
+#   EPP deployment/service: ${GIE_EPP_RELEASE}-${model}-epp     e.g. epp-sim-model-epp
+#   InferencePool name:     ${GIE_EPP_RELEASE}-${model}          (created by Helm chart)
+#   InferenceObjective:     batch-sheddable-${model}             (GIE mode)
+#                           batch-sheddable                      (non-GIE mode)
+#   Managed-by label:       app.kubernetes.io/managed-by=batch-gateway-dev
 ENABLE_GIE="${ENABLE_GIE:-false}"
 GIE_REPO="${GIE_REPO:-}"
 GIE_UPSTREAM_REPO="https://github.com/kubernetes-sigs/gateway-api-inference-extension.git"
+GIE_VERSION="${GIE_VERSION:-v1.4.0}"
```
yizhaodev: 1.4.0 released at Mar, does it has the flow control feature we need?

lioraron (commented):
## Review

### `ensure_gie_repo` silently modifies sibling repo and ignores checkout failures

```bash
if [ -d "${sibling_dir}" ] && git -C "${sibling_dir}" rev-parse "${GIE_VERSION}" &>/dev/null; then
    GIE_REPO="${sibling_dir}"
    git -C "${GIE_REPO}" checkout "${GIE_VERSION}" --quiet 2>/dev/null || true
```

Two problems:

1. **Side effect on sibling repo**: `git checkout v1.4.0` changes the checked-out ref in the user's GIE working tree. A developer working on a GIE feature branch who runs `make dev-deploy-gie` gets silently switched to v1.4.0.

2. **Silent failure path**: if checkout fails (e.g., dirty working tree), `|| true` suppresses the error but `GIE_REPO` is already set to the sibling dir. The script continues using whatever branch the sibling is on — not necessarily v1.4.0. This can produce confusing helm errors from chart incompatibilities.

**Suggestion**: use a detached worktree instead of checking out in-place, or fall through to the clone path when checkout fails:

```bash
if [ -d "${sibling_dir}" ] && git -C "${sibling_dir}" rev-parse "${GIE_VERSION}" &>/dev/null; then
    if git -C "${sibling_dir}" checkout "${GIE_VERSION}" --quiet 2>/dev/null; then
        GIE_REPO="${sibling_dir}"
        log "Using sibling GIE repo at ${GIE_REPO} (checked out ${GIE_VERSION})"
    else
        warn "Sibling GIE repo has uncommitted changes; cloning pinned version instead"
    fi
```

This way: successful checkout → use sibling; failed checkout → fall through to clone, leaving the user's repo untouched.

---

The rest of the PR looks solid — per-model EPP architecture is correct, test structure is well-organized, and the CI matrix addition is sensible.

scripts/dev-deploy.sh
```diff
@@ -887,11 +915,6 @@ inferenceExtension:
       plugins:
         - type: round-robin-fairness-policy
         - type: global-strict-fairness-policy
-        - type: slo-deadline-ordering-policy
```
yizhaodev, line 890: why are they removed?

scripts/dev-deploy.sh
yizhaodev: probably need 1.5.0 or later

scripts/dev-deploy.sh
```diff
@@ -846,14 +865,18 @@ ensure_gie_repo() {
...
         local sibling_dir
         sibling_dir="$(cd "${REPO_ROOT}/.." && pwd)/gateway-api-inference-extension"
-        if [ -d "${sibling_dir}" ]; then
+        if [ -d "${sibling_dir}" ] && git -C "${sibling_dir}" rev-parse "${GIE_VERSION}" &>/dev/null; then
             GIE_REPO="${sibling_dir}"
-            log "Using sibling GIE repo at ${GIE_REPO}"
-        else
-            log "GIE repo not found locally. Cloning from ${GIE_UPSTREAM_REPO}..."
+            log "Using sibling GIE repo at ${GIE_REPO} (checking out ${GIE_VERSION})"
+            git -C "${GIE_REPO}" checkout "${GIE_VERSION}" --quiet 2>/dev/null || true
```
lioraron: **Issue: `git checkout` modifies the user's sibling repo working tree**

When `GIE_REPO` resolves to the sibling directory (`../gateway-api-inference-extension`), `git checkout "${GIE_VERSION}"` detaches HEAD in the user's local clone. This is a side effect on a repo the user may be actively working in.

Additionally, if the checkout fails (wrong tag, dirty worktree), `|| true` suppresses the error but `GIE_REPO` is already set — so the script continues with whatever ref was previously checked out, which could be an incompatible chart version.

Suggestion: only set `GIE_REPO` after confirming the checkout succeeded, or use `git worktree` / a temp clone for the sibling case too.

test/e2e/flow_control_test.go
```diff
@@ -0,0 +1,534 @@
...
+	// The processor records 429 responses as output and marks the batch
+	// as completed (all requests were processed, even if none succeeded).
+	if finalBatch.Status != openai.BatchStatusCompleted {
+		t.Errorf("expected batch status %q (processor finished processing), got %q",
+			openai.BatchStatusCompleted, finalBatch.Status)
+	}
+	if finalBatch.RequestCounts.Completed != 0 {
+		t.Errorf("expected 0 successfully completed requests, got %d", finalBatch.RequestCounts.Completed)
+	}
+	if finalBatch.RequestCounts.Failed == 0 {
```
yizhaodev: failed must be 2?

test/e2e/flow_control_test.go
```diff
@@ -0,0 +1,534 @@
...
+		t.Fatalf("expected batch to complete within SLO window, got status %s", finalBatch.Status)
+	}
+	if finalBatch.RequestCounts.Completed != 2 {
+		t.Fatalf("expected 2 completed, got %d", finalBatch.RequestCounts.Completed)
+	}
+
+	t.Logf("batch completed within 10m SLO window (x-slo-ttft-ms header propagation unit-tested)")
+}
+
+// doTestRetryOn429 submits a batch targeting a simulator with 50% failure
```
yizhaodev, line 131: why not change failure rate to 100% for model 429
then you only need one request for testing.

test/e2e/flow_control_test.go
```diff
@@ -0,0 +1,534 @@
...
+	if batch.RequestCounts.Completed != 2 {
+		t.Fatalf("expected 2 completed requests, got %d", batch.RequestCounts.Completed)
+	}
+	if batch.RequestCounts.Failed != 0 {
+		t.Fatalf("expected 0 failed requests, got %d", batch.RequestCounts.Failed)
+	}
+
+	eppLogs := getEPPLogsSince(t, eppDeployment, sinceTime)
+
+	received := strings.Count(eppLogs, "EPP received request")
```
yizhaodev, line 305: should we also check if it is handled by flow contorl?
does FC has its own special log pattern?

test/e2e/flow_control_test.go
```diff
@@ -0,0 +1,534 @@
...
+			return b
+		}
+		time.Sleep(2 * time.Second)
+	}
+	t.Fatalf("batch %s did not reach terminal status within %v", batchID, timeout)
+	return nil
+}
+
+// fetchOutputFile downloads the output file for a batch and returns its body.
+func fetchOutputFile(t *testing.T, batch *openai.Batch) string {
```
yizhaodev: can you move it to `helpers_test.go`
func `fetchBatchResults` can reuse this as well

scripts/dev-deploy.sh
```diff
@@ -846,14 +865,18 @@ ensure_gie_repo() {
     else
         local sibling_dir
         sibling_dir="$(cd "${REPO_ROOT}/.." && pwd)/gateway-api-inference-extension"
```
yizhaodev: can we not deploy to sibling dir
i do have my GIE working dir at there
how about /tmp
@lioraron ?

scripts/dev-deploy.sh
```diff
@@ -887,11 +915,6 @@ inferenceExtension:
...
         - type: round-robin-fairness-policy
         - type: global-strict-fairness-policy
-        - type: slo-deadline-ordering-policy
-        - type: utilization-detector
-          parameters:
-            queueDepthThreshold: 5
-            kvCacheUtilThreshold: 0.8
       flowControl:
         maxBytes: 4294967296
         defaultRequestTTL: 30s
```
lioraron, line 915: **Keep the full flow-control config from the guide doc**

The removed plugins (`slo-deadline-ordering-policy`, `utilization-detector` + its parameters), the `saturationDetector` block, and the `orderingPolicyRef: slo-deadline-ordering-policy` on the batch priority band should all be kept to match the config documented in `docs/guides/flow-control-setup.md`.

The dev-deploy should validate that EPP actually accepts and parses the full guide config. These plugins are safe to declare against simulators — EPP loads the full plugin config at startup regardless of whether metrics are available at runtime. The `utilization-detector` will simply never fire (thresholds never met), and `slo-deadline-ordering-policy` falls back to FCFS when there's no saturation signal. No errors, no crashes — just inactive code paths.

scripts/dev-deploy.sh
lioraron: Agreed — I raised the same concern [here](https://github.com/llm-d-incubation/batch-gateway/pull/408/changes#r3182664389). The sibling dir should be treated as read-only. When it exists, the script should clone to a temp dir (or use `git worktree`) rather than running `git checkout` on it.

Makefile
```diff
@@ -337,6 +341,10 @@ dev-rm-cluster:
 	@kind delete cluster --name batch-gateway-dev || echo "Cluster not found or already deleted"
 	@echo "✅ Kind cluster deleted"
 
+## test-e2e-gie: Run E2E tests including GIE flow control (requires dev-deploy-gie)
+test-e2e-gie:
+	@ENABLE_GIE=true $(MAKE) test-e2e
```
lioraron: **This target is unnecessary — the e2e tests already auto-detect GIE**

`detectGIEDeployed(t)` in the test code checks for EPP deployments via `kubectl get deployments` and skips the GIE test group when none are found. So plain `make test-e2e` works in both cases: GIE tests run when EPP is present, skip when it's not.

`ENABLE_GIE` is not read by the test binary at all — the env var is only used by `dev-deploy.sh`. This target just passes it through to `make test-e2e` where it's ignored.

Suggest removing `test-e2e-gie` to maintain the convention that `test-e2e` auto-detects the environment.

scripts/dev-deploy.sh
j-mok-dev: thanks. updated to 1.5.0

scripts/dev-deploy.sh
j-mok-dev: restored. I initially removed them thinking simulators wouldn't need the full config, but you're right.

scripts/dev-deploy.sh
j-mok-dev: fixed. removed the sibling repo logic entirely.
if `GIE_REPO` is not explicitly set, the script now always clones the pinned version into a temp directory.

test/e2e/flow_control_test.go
j-mok-dev: thanks. fixed.

test/e2e/flow_control_test.go
j-mok-dev: This test verifies "429 -> retry -> eventual success", so the model needs a partial failure rate. With 100%, retries also fail and it becomes identical to RetryExhaustion. 50% rate with 10 requests keeps P(zero retries) ≈ 0.1% for CI stability.

test/e2e/flow_control_test.go
j-mok-dev: moved to helpers_test.go.

scripts/dev-deploy.sh
j-mok-dev: thanks, fixed. now always clones to a temp dir when GIE_REPO is not explicitly set.

scripts/dev-deploy.sh
j-mok-dev: thanks. good point. restored the full config to match the guide doc.

Makefile
j-mok-dev: thanks. removed.

test/e2e/flow_control_test.go
j-mok-dev: thanks. epp's flow controller logs "Item dispatched." at TRACE level (V=5) for each request it dispatches. I've enabled EPP TRACE logging in dev-deploy and added an assertion for this log entry.

scripts/dev-deploy.sh
```diff
@@ -842,19 +861,12 @@ EOF
...
-            GIE_REPO="${sibling_dir}"
-            log "Using sibling GIE repo at ${GIE_REPO}"
-        else
-            log "GIE repo not found locally. Cloning from ${GIE_UPSTREAM_REPO}..."
-            GIE_REPO="$(mktemp -d)/gateway-api-inference-extension"
-            git clone --depth 1 "${GIE_UPSTREAM_REPO}" "${GIE_REPO}"
-            log "Cloned GIE repo to ${GIE_REPO}"
-        fi
+        log "Cloning ${GIE_UPSTREAM_REPO} at ${GIE_VERSION}..."
+        GIE_REPO="$(mktemp -d)/gateway-api-inference-extension"
```
yizhaodev, line 867: when will dir created by `mktemp -d` be deleted? will they always stay in my laptop?

Makefile
```diff
@@ -327,6 +327,10 @@ KIND_CLUSTER_NAME ?= batch-gateway-dev
 dev-deploy:
 	@bash scripts/dev-deploy.sh
 
+## dev-deploy-gie: Deploy with GIE integration (per-model EPP + InferenceObjectives)
+dev-deploy-gie:
```
yizhaodev, line 331: do we need new target for this?

lioraron (approved):

scripts/dev-deploy.sh
j-mok-dev: thanks. good catch. the temp dir is now removed right after GIE deploy finishes (CRDs + EPP installs). but it doesn't remove it if the GIE_REPO was provided by user.

Makefile
j-mok-dev: yes, it's convenience to have it instead of typing `ENABLE_GIE=true make dev-deploy`, as the deploy script needs the env var to decide whether to install EPP instances.

yizhaodev (approved):
thanks!

PR #410, opened 2026-05-06
feat(ci): publish Helm chart to GHCR OCI registry on release

acardace:
## Summary

- Add a `helm` job to the CI Release workflow that packages and pushes `charts/batch-gateway` to `oci://ghcr.io/llm-d-incubation/charts/batch-gateway` on every push to `main` and on version tags
- Job is skipped for PR/feature branches to avoid polluting the registry
- Chart version is derived from the trigger: semver tag → `1.2.3`, main → `0.0.0-main-<sha>`

## Motivation

The `llm-d-batch-gateway-operator` currently consumes this chart via a git submodule, which requires cloning the entire `batch-gateway` repo just to get the chart. Publishing the chart to GHCR allows the operator to pull it directly by version (OCI), removing the submodule dependency entirely.

Ref: https://github.com/opendatahub-io/llm-d-batch-gateway-operator

### DIFF
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -88,3 +88,52 @@ jobs:
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-processor:${{ steps.meta.outputs.tag }}"
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-gc:${{ steps.meta.outputs.commit_sha }}"
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-gc:${{ steps.meta.outputs.tag }}"
+
+  helm:
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'main' || github.ref_type == 'tag'
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+
+      - name: Log in to GitHub Container Registry
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io --username ${{ github.actor }} --password-stdin
+
+      - name: Determine chart version
+        id: chart
+        run: |
+          COMMIT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
+
+          if [[ "${{ github.ref_type }}" == "tag" ]]; then
+            # Semver tag — use it directly as the chart version (strip leading 'v')
+            CHART_VERSION="${{ github.ref_name }}"
+            CHART_VERSION="${CHART_VERSION#v}"
+            APP_VERSION="${{ github.ref_name }}"
+          else
+            # main branch
+            CHART_VERSION="0.0.0-main-${COMMIT_SHA}"
+            APP_VERSION="main-${COMMIT_SHA}"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-release.yaml
```diff
@@ -88,3 +88,52 @@ jobs:
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-processor:${{ steps.meta.outputs.tag }}"
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-gc:${{ steps.meta.outputs.commit_sha }}"
           echo "  - ghcr.io/llm-d-incubation/batch-gateway-gc:${{ steps.meta.outputs.tag }}"
+
+  helm:
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'main' || github.ref_type == 'tag'
```
yizhaodev, line 94: i recall `create-release.yml` will also publish helm oci image after tagging
@vishbhat please help to confirm

if that is true, we should only publish oci `0.0.0-main-<sha>` upon each commit

yizhaodev:
quick question: should batch gateway operator consumes charts from upstream (`oci://ghcr.io/llm-d-incubation/charts/batch-gateway`) or midstream?

currently the gitbsubmodule points to midstream
https://github.com/opendatahub-io/llm-d-batch-gateway-operator/blob/main/.gitmodules

.github/workflows/ci-release.yaml
vishbhat: @yizhaodev Yes, you're right. `create-release.yml` already publishes the Helm chart on version tags.

acardace:
> quick question: should batch gateway operator consumes charts from upstream (`oci://ghcr.io/llm-d-incubation/charts/batch-gateway`) or midstream?
> 
> currently the gitbsubmodule points to midstream https://github.com/opendatahub-io/llm-d-batch-gateway-operator/blob/main/.gitmodules

I would point to upstream charts to keep things in sync, then downstream we can pin them according to our needs. 
@zdtsw what do you think? In general I don't really like the idea of keeping a git submodule around just for the charts.

.github/workflows/ci-release.yaml
yizhaodev: @acardace 
I think you should remove condition `github.ref_type == 'tag'` because `create-release.yml` will create oci on tagging

zdtsw:
I might be wrong but I do not think mid/downstream can use submodel.

zdtsw:
Btw I am a bit confused, I thought batch -gateway-operator is only for mid/downstream. Is the plan to get it in upstream llm-d( start with incubation first)?

acardace:
> Btw I am a bit confused, I thought batch -gateway-operator is only for mid/downstream. Is the plan to get it in upstream llm-d( start with incubation first)?

No it will only be a midstream component.

lioraron (commented):
@acardace Is this PR still needed?
1. @yizhaodev pointed out that `create-release.yml` already publishes the Helm chart on version tags. This PR's `github.ref_type == 'tag'` condition would create duplicate publishing?
2. The stated motivation is replacing the operator's git submodule, but the operator is midstream and its submodule points to `opendatahub-io/batch-gateway`, not this upstream repo. Who is the actual consumer of these main-branch chart versions (`0.0.0-main-<sha>`)?

acardace:
@yizhaodev @lioraron closing as this is not needed, I missed the `create-release.yaml` comment, thank you!

PR #411, opened 2026-05-06
ci: add helm lint and unittest to PR workflow

yizhaodev:
## Why is this PR needed?

The pre-commit workflow has a `helm-unittest` hook but it silently skips in CI because Helm is not installed before `pre-commit/action` runs. `helm lint` is not configured anywhere. This means broken Helm charts can be merged and published.

## What does this PR do?

- Install Helm and the helm-unittest plugin before pre-commit hooks run, so the existing hook actually executes
- Add an explicit `helm lint` step to catch chart issues before merge

## How was this tested?

## Checklist

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -28,6 +28,9 @@ jobs:
           go-version-file: go.mod
           cache: true
 
+      - name: Install Helm
+        uses: azure/setup-helm@v5
+
       - name: Install Go tools (goimports, gosec, ruleguard)
         run: make install-pre-commit-tools
 
@@ -38,3 +41,6 @@ jobs:
 
       - name: Run pre-commit hooks
         uses: pre-commit/action@v3.0.1
+
+      - name: Lint Helm chart
+        run: helm lint charts/batch-gateway
--- a/Makefile
+++ b/Makefile
@@ -233,8 +233,7 @@ install-pre-commit-tools:
 	$(GO) install github.com/quasilyte/go-ruleguard/cmd/ruleguard@v0.4.5
 	@if command -v helm >/dev/null 2>&1; then \
 		helm plugin list | grep -q unittest || \
-			helm plugin install https://github.com/helm-unittest/helm-unittest.git || \
-			helm plugin install --verify=false https://github.com/helm-unittest/helm-unittest.git; \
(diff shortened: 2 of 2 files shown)


lioraron:
## Review feedback

Good catch — the helm-unittest pre-commit hook was indeed silently skipping in CI due to missing Helm. Two issues:

### 1. `azure/setup-helm@v4` — should be `@v5`

The other workflows in this repo (`ci-integration-tests.yml`, `create-release.yml`) use `azure/setup-helm@v5`. This PR pins to `@v4`. Use `@v5` for consistency.

### 2. helm-unittest plugin not installed in CI

The PR installs Helm but doesn't install the helm-unittest plugin before `pre-commit/action` runs. The pre-commit hook for `helm-unittest` checks `helm plugin list | grep -q unittest` — without the plugin installed, it will still print "helm-unittest not installed, skipping" and pass silently. The only improvement is that `helm lint` now runs as a separate step.

To actually fix the original problem (helm-unittest silently skipping), add the plugin install after the Helm install step:

```yaml
- name: Install Helm
  uses: azure/setup-helm@v5

- name: Install helm-unittest plugin
  run: helm plugin install --verify=false --version v1.0.3 https://github.com/helm-unittest/helm-unittest.git
```

The Makefile change to `install-pre-commit-tools` only runs the plugin install when `helm` is already present (`if command -v helm`), and `install-pre-commit-tools` runs *after* Helm is installed in the workflow — so that path should work too. But it would be clearer to verify: does the `Install Go tools` step (which calls `make install-pre-commit-tools`) run after the new `Install Helm` step? Looking at the ordering in the diff, yes it does. So the Makefile's `helm plugin install` should now actually execute. If that's the intended path, then the plugin install is covered — but you should remove the "skipping" fallback message from the Makefile since it would now be a real failure if the install fails.

### Minor

The Makefile change from try-without-verify-then-fallback-to-verify-false to just `--verify=false --version v1.0.3` is a good simplification. Pinning the version prevents surprise breakage.

yizhaodev:
Thanks for the review.

1. Fixed: updated to `@v5`.
2. The plugin install is covered: `Install Helm` runs before `make install-pre-commit-tools`, so the Makefile's `helm plugin install` now executes (previously skipped because helm wasn't found). Verified in the latest CI run.

lioraron (approved):

Issue #413, opened 2026-05-07
Rate limit should exempt batch processor requests to the LLM route


yizhaodev:
## Problem

The `TokenRateLimitPolicy` on the LLM route (or Gateway) enforces per-user token rate limiting using `auth.identity.user.username` as the counter. The batch processor currently sends inference requests **back through the same Gateway** with the **user's original token** (via `passThroughHeaders`). This means batch (offline) requests consume the same rate limit quota as online (direct user) requests.

Rate limiting should only apply to **online requests** from users directly, not to **batch requests** from the batch processor.

### Current flow

```
Online:  User --> Gateway --> llm-route --> [TokenRateLimitPolicy] --> InferencePool --> EPP --> vLLM
Batch:   User --> Gateway --> batch-route --> Apiserver --> Processor --> Gateway --> llm-route --> [TokenRateLimitPolicy] --> InferencePool --> EPP --> vLLM
```

Both flows share the same per-user token rate limit counter on the LLM route.

### Desired behavior

- Online requests: subject to `TokenRateLimitPolicy`
- Batch processor requests: exempt from `TokenRateLimitPolicy`, but still routed through Gateway / Envoy / EPP / vLLM (to preserve intelligent routing)

## Constraints

- **Must go through Envoy + EPP**: The EPP (Endpoint Picker Plugin) runs as an Envoy ext_proc filter and provides KV-cache aware scheduling. Bypassing the Gateway means bypassing EPP, which degrades to round-robin load balancing. This is especially harmful for batch workloads with high request volume. Direct backend access is ruled out.
- **Must not be spoofable**: Online users should not be able to bypass rate limiting by impersonating batch requests.

## Decided Approach: Internal Gateway + Dedicated HTTPRoute

After evaluating Options A–C below, we chose a fourth approach that combines the best of the alternatives:

The batch processor routes through a **ClusterIP-only Internal Gateway** (not reachable externally) → dedicated `batch-llm-route` (no `TokenRateLimitPolicy`) → same InferencePool/EPP. The user's original token is forwarded, so AuthPolicy and audit logging work as-is. External users cannot bypass rate limiting because they cannot reach the Internal Gateway.

### New flow

```
Online:  User --> Gateway --> llm-route --> [AuthPolicy + TokenRateLimitPolicy] --> InferencePool --> EPP --> vLLM
[...]

PR #416, opened 2026-05-07
feat: send x-batch-request header on inference requests

yizhaodev:
## Why is this PR needed?

Downstream inference gateways currently have no way to distinguish batch traffic from real-time requests. This makes it difficult to apply differentiated policies (e.g. scheduling priority, rate limiting, resource allocation) based on request origin.

## What does this PR do?

Adds an unconditional `x-batch-request: true` header to every outgoing inference request from the batch processor. This allows downstream gateways to identify and handle batch traffic differently from real-time traffic.

## How was this tested?

## Checklist

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -626,6 +626,7 @@ const (
 	sloTTFTMSHeader          = "x-slo-ttft-ms"
 	inferenceObjectiveHeader = "x-gateway-inference-objective"
 	fairnessIDHeader         = "x-gateway-inference-fairness-id"
+	batchRequestHeader       = "x-batch-request"
 )
 
 // mergeInferenceHeaders adds processor-managed headers to the outgoing inference request:
@@ -748,6 +749,10 @@ func (p *Processor) executeOneRequest(
 
 	headers := maps.Clone(passThroughHeaders)
 	headers = mergeInferenceHeaders(headers, sloCtx, p.cfg.InferenceObjectiveFor(modelID), fairnessID)
+	if headers == nil {
+		headers = make(map[string]string)
+	}
+	headers[batchRequestHeader] = "true"
 
 	inferReq := &inference.GenerateRequest{
 		RequestID: newBatchRequestID(requestID),
--- a/internal/processor/worker/executor_test.go
+++ b/internal/processor/worker/executor_test.go
@@ -47,6 +47,9 @@ func TestExecuteOneRequest_Success(t *testing.T) {
 			if _, ok := req.Headers[sloTTFTMSHeader]; !ok {
 				t.Fatalf("expected %s header to be set", sloTTFTMSHeader)
 			}
+			if req.Headers[batchRequestHeader] != "true" {
+				t.Fatalf("expected %s header to be \"true\", got %q", batchRequestHeader, req.Headers[batchRequestHeader])
+			}
 			return &inference.GenerateResponse{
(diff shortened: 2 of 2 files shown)


internal/processor/worker/executor.go
```diff
@@ -626,6 +626,7 @@ const (
 	sloTTFTMSHeader          = "x-slo-ttft-ms"
 	inferenceObjectiveHeader = "x-gateway-inference-objective"
 	fairnessIDHeader         = "x-gateway-inference-fairness-id"
+	batchRequestHeader       = "x-batch-request"
```
lioraron, line 629: How do we prevent another caller, which is not batch-processor, from adding this header and bypass the rate limit?
Is there a more secure way?

lioraron (commented):
Thanks!
Pls add the corresponding issue?

internal/processor/worker/executor.go
yizhaodev: more secure way is to let admin define the header in config

1) admin define the header name in config
2) admin define the policy to decide which header should pass rate limit

since only admin know the header, user can't bypass it.

lioraron:
Regarding the relationship to #413:

The [decision comment](https://github.com/llm-d-incubation/batch-gateway/issues/413#issuecomment-4399170166) on that issue chose **Option B (dedicated HTTPRoute)** and explicitly rejected Option C (header-based) as "fragile to maintain" and "not portable." Yet this PR implements an Option C mechanism — adding a spoofable `x-batch-request` header — without implementing any part of Option B.

With Option B, rate limit exemption comes from the route itself (no `TokenRateLimitPolicy` attached to the batch-specific route), so there's nothing to spoof and no header stripping needed.

If this header is intended as an *additional* signal for downstream scheduling/priority (separate from the rate limit exemption, which would be handled by the dedicated route), that's reasonable — but the PR description should clarify that explicitly and note that this header alone is **not sufficient** for rate limit exemption since any external caller can set it.

Could you clarify how this PR fits into the Option B plan?

internal/processor/worker/executor.go
lioraron: This is security-through-obscurity — a configurable header name doesn't make it secure:

1. **Header names are visible in transit** — anyone with network access or Envoy access logs can see all request headers, including the "secret" one
2. **Header names leak** — through monitoring dashboards, OpenTelemetry traces, debug logs, etc.
3. **The source code is public** — the header name (or its config key) is readable by anyone
4. **No authentication** — there's no cryptographic proof that the header was set by the batch processor vs. an external caller

The [decided approach (Option B)](https://github.com/llm-d-incubation/batch-gateway/issues/413#issuecomment-4399170166) solves this architecturally — a dedicated HTTPRoute without `TokenRateLimitPolicy` doesn't need any header for rate limit exemption. The route itself *is* the authorization boundary.

If this header is useful for other purposes (e.g., EPP scheduling hints), that's fine — but it should not be relied on for security-sensitive decisions like rate limit exemption.

yizhaodev:
@lioraron 
The PR is for **Option C: Header-based exclusion**

For **Option B (dedicated HTTPRoute)**
In RHOAI, kserve auto deploys the LLM route. so user need to manually maintain the new LLM route just for batch
If this is accepatble, we can close this PR and go with option B

internal/processor/worker/executor.go
yizhaodev: make sense. closing this pr

PR #423, opened 2026-05-11
feat: add internal gateway to bypass token rate limit

### LINKED ISSUE #413: Rate limit should exempt batch processor requests to the LLM route

yizhaodev:
## Why is this PR needed?

The batch processor currently sends inference requests through the main gateway, which applies per-user token rate limits.

## What does this PR do?

- Update scripts to add an **Internal Gateway** (ClusterIP, no external exposure) that the batch processor uses to reach inference backends. This gateway enforces authn/authz but skips token rate limiting.
- Updates docs to reflect the two-gateway topology.

## How was this tested?

- [x] `deploy-k8s.sh install && deploy-k8s.sh test` — 13/13 passed on OpenShift CI cluster

Fixes https://github.com/llm-d-incubation/batch-gateway/issues/413


j-mok-dev:
`deploy-maas.sh` is referenced in the test checklist but doesn't exist in this PR or the repo. is it coming in a follow-up PR?

yizhaodev:
> `deploy-maas.sh` is referenced in the test checklist but doesn't exist in this PR or the repo. is it coming in a follow-up PR?

the description was updated, `deploy-rhoai.sh` and `deploy-maas.sh` will be only in midstream
https://github.com/opendatahub-io/batch-gateway/pull/52

examples/deploy-demo/common.sh
```diff
@@ -705,6 +705,53 @@ do_deploy_batch_gateway() {
...
+  gatewayClassName: ${GATEWAY_CLASS_NAME}
+  listeners:
+  - name: http
+    port: 80
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: Selector
+        selector:
+          matchLabels:
```
lioraron, line 739: `GATEWAY_CLASS_NAME` is used here and on line 752 (`wait_for_deployment`) but is only defined in `deploy-k8s.sh`. If another script sources `common.sh` and calls `create_batch_internal_gateway()`, it will crash with `unbound variable` (under `set -u`) or produce an invalid Gateway spec.

Consider either adding a default in `common.sh`:
```bash
GATEWAY_CLASS_NAME="${GATEWAY_CLASS_NAME:-istio}"
```
Or adding a guard at the top of this function:
```bash
[ -z "${GATEWAY_CLASS_NAME:-}" ] && die "GATEWAY_CLASS_NAME is not set."
```

examples/deploy-demo/common.sh
```diff
@@ -715,6 +762,66 @@ do_deploy_batch_gateway() {
...
+    svc_type=$(kubectl get svc -n "${gw_ns}" \
+        -l "gateway.networking.k8s.io/gateway-name=${gw_name}" \
+        -o jsonpath='{.items[0].spec.type}' 2>/dev/null || echo "")
+
+    echo ""
+    echo "── Check: Internal Gateway Service type is ClusterIP ──"
+    if [ "${svc_type}" = "ClusterIP" ]; then
+        echo "  Service type: ${svc_type}"
+        log "PASSED: Internal Gateway Service is ClusterIP (no external IP)"
+        _ig_passed=$((_ig_passed + 1))
```
lioraron, line 793: Two improvements for `check_batch_internal_gateway()`:

**1. Empty `svc_type` gives a misleading error.** If the Internal Gateway service doesn't exist, the user sees `FAILED: Expected ClusterIP, got ''`, which sends them debugging service type annotations rather than checking whether the Gateway was created. Consider:
```bash
if [ -z "${svc_type}" ]; then
    warn "FAILED: No Service found for Internal Gateway in namespace '${gw_ns}'"
    _ig_failed=$((_ig_failed + 1))
elif [ "${svc_type}" = "ClusterIP" ]; then
    # ...existing pass logic...
```

**2. Route check gives false PASSED on non-OpenShift.** `kubectl get route` fails because the resource type doesn't exist on vanilla K8s, but the error is swallowed by `2>/dev/null`, so the check always passes. The existing `is_openshift()` helper (line 108) can gate this:
```bash
if is_openshift; then
    # check for Routes
else
    log "SKIPPED: OpenShift Routes not available on this cluster"
fi
```

examples/deploy-demo/common.sh
```diff
@@ -899,7 +1006,7 @@ JSONL
     assert_http 200 "Unauthorized but authenticated batch request -> 200 (no authz on batch-route)" -H "${unauthorized_header}" "${batch_url}/v1/batches"
 
     # ── 5. Batch Authz (LLM route enforces) ──────────────
```
lioraron, line 1016: nit: These test descriptions still reference "LLM route" but the batch processor now goes through `batch-llm-route` on the Internal Gateway:
```
next_test "Unauthorized user batch -> LLM route rejects with 403"
echo "  Goal: Batch processor forwards unauthorized credentials to LLM route, gets 403"
```
Should be updated for consistency with the rest of the PR's terminology.

examples/deploy-demo/deploy-k8s.sh
```diff
@@ -552,17 +671,25 @@ cmd_install() {
     apply_llm_auth_policy
     apply_llm_token_rate_limit
 
+    create_batch_internal_gateway
+    create_batch_llm_route
+    apply_batch_llm_auth_policy
```
lioraron, line 676: Consider calling `check_batch_internal_gateway` here at the end of `cmd_install` (after `create_batch_internal_gateway` + `create_batch_llm_route`), not just in `cmd_test`. An operator who runs `install` without immediately running `test` won't get any isolation verification, and could run with a misconfigured gateway.

lioraron (commented):
Clean implementation of Option D from #413. The two-gateway topology is sound — Internal Gateway (ClusterIP, HTTP, AuthPolicy) for batch, External Gateway (LoadBalancer, HTTPS, AuthPolicy + TokenRateLimitPolicy) for online. The AuthPolicy on `batch-llm-route` is byte-for-byte identical to `llm-route-auth`, which is correct. Docs and architecture diagram are thorough.

**Scope note**: This PR bundles several changes beyond the Internal Gateway: FS→S3/MinIO storage default, `LLMD_VERSION` default change (`main`→`v0.6.0`), `helm install`→`helm upgrade --install` throughout, PVC `ReadWriteOnce`→`ReadWriteMany`. These are individually fine but worth calling out in the description for reviewers/bisect. The RWX change in particular breaks the FS storage path on clusters with block storage (gp2/gp3) — the docs mention this but the script has no warning.

examples/deploy-demo/common.sh
yizhaodev: fixed, thanks

examples/deploy-demo/common.sh
yizhaodev: fixed, thanks

examples/deploy-demo/common.sh
yizhaodev: fixed, thanks

examples/deploy-demo/deploy-k8s.sh
yizhaodev: fixed, thanks

lioraron (commented):
Re-reviewed the updated diff. All 4 previously flagged issues are confirmed fixed:

1. **`GATEWAY_CLASS_NAME` guard** — now properly checks `${GATEWAY_CLASS_NAME:-}` with `die` on empty
2. **Empty `svc_type` handling** — warns and returns before the ClusterIP assertion
3. **Route check gated on OpenShift** — wrapped in `if is_openshift; then ... fi`
4. **Test descriptions** — updated from "LLM route" to "batch-llm-route"

lioraron (approved):

PR #424, opened 2026-05-11
test(e2e): add AIMD adaptive concurrency e2e tests (#420)

### LINKED ISSUE #420: [flow-control] Add e2e tests for AIMD adaptive concurrency control
## Summary

PR #394 added AIMD adaptive concurrency control with thorough unit/integration tests, but no e2e tests that exercise AIMD through a real deployment.

## Depends on

- #419 (AIMD Prometheus metrics) — e2e tests need observable metrics to verify AIMD behavior without relying on log scraping.

## Proposed test scenarios

1. **429 triggers concurrency decrease**: Submit batch requests, configure the vLLM simulator to return 429s, verify `batch_processor_aimd_concurrency_limit` decreases.

2. **Recovery after backpressure**: After the 429 phase, let the simulator return 200s and verify the AIMD limit gradually increases back toward `per_endpoint`.

3. **Endpoint isolation**: With two model endpoints, trigger 429s on only one. Verify only that endpoint's AIMD limit decreases while the other stays at max.

## Infrastructure

- Uses the existing dev-deploy Kind cluster with vLLM simulator
- Simulator's admin API can be used to toggle 429 responses (or configure via delay/capacity controls)
- Prometheus is already deployed in the e2e environment — query `batch_processor_aimd_concurrency_limit` via PromQL

j-mok-dev:
## Why is this PR needed?

PR #394 added AIMD adaptive concurrency control with unit tests, but no e2e tests that exercise AIMD through a real deployment. Operators need confidence that AIMD reacts correctly to backpressure in a live cluster.

## What does this PR do?

Adds e2e tests for all three AIMD scenarios from #420:

- **Decrease**: 429s from sim-model-429 drive the AIMD concurrency limit below perEndpoint.
- **Isolation**: sim-model (0% failure) stays at perEndpoint while sim-model-429 decreases independently.
- **Recovery**: a dedicated sim-model-aimd starts at 100% failure (driving limit to min), then is patched to 0% failure via `kubectl patch`; subsequent requests drive the limit back toward perEndpoint.

Infrastructure changes:
- New `vllm-sim-aimd` simulator in dev-deploy (100% failure rate, patched to 0% mid-test for recovery).
- `patchSimulatorFailureRate` helper for toggling simulator failure rate via deployment patch.

## How was this tested?

- [x] `make pre-commit` passes
- [x] `make test-e2e` passes

## Checklist

## Related Issues

Fixes #420

---

> ~~**Stacked on #394** (`feat/adaptive-concurrency`). This PR targets `feat/adaptive-concurrency` as its base branch. Review only the top commit; the rest belongs to #394.~~
> **Update**: #394 has been merged. This PR is now rebased on `main`.

### DIFF
--- a/scripts/dev-deploy.sh
+++ b/scripts/dev-deploy.sh
@@ -25,6 +25,7 @@ VLLM_SIM_MODEL="${VLLM_SIM_MODEL:-sim-model}"
 VLLM_SIM_B_MODEL="${VLLM_SIM_B_MODEL:-sim-model-b}"
 VLLM_SIM_429_MODEL="${VLLM_SIM_429_MODEL:-sim-model-429}"
 VLLM_SIM_ALWAYS_FAIL_MODEL="${VLLM_SIM_ALWAYS_FAIL_MODEL:-sim-model-always-fail}"
+VLLM_SIM_AIMD_MODEL="${VLLM_SIM_AIMD_MODEL:-sim-model-aimd}"
 VLLM_SIM_IMAGE="${VLLM_SIM_IMAGE:-ghcr.io/llm-d/llm-d-inference-sim:latest}"
 JAEGER_IMAGE="${JAEGER_IMAGE:-${IMAGE_REGISTRY}/jaegertracing/all-in-one:latest}"
 PROMETHEUS_IMAGE="${PROMETHEUS_IMAGE:-${IMAGE_REGISTRY}/prom/prometheus:latest}"
@@ -1047,6 +1048,12 @@ install_batch_gateway() {
         --set "processor.config.modelGateways.${VLLM_SIM_ALWAYS_FAIL_MODEL}.maxRetries=1"
         --set "processor.config.modelGateways.${VLLM_SIM_ALWAYS_FAIL_MODEL}.initialBackoff=500ms"
         --set "processor.config.modelGateways.${VLLM_SIM_ALWAYS_FAIL_MODEL}.maxBackoff=1s"
+        # AIMD recovery test model: starts at 100% rate_limit, patched to 0% mid-test.
+        --set "processor.config.modelGateways.${VLLM_SIM_AIMD_MODEL}.url=http://${VLLM_SIM_AIMD_NAME}.${NAMESPACE}.svc.cluster.local:8000"
+        --set "processor.config.modelGateways.${VLLM_SIM_AIMD_MODEL}.requestTimeout=2m"
+        --set "processor.config.modelGateways.${VLLM_SIM_AIMD_MODEL}.maxRetries=10"
+        --set "processor.config.modelGateways.${VLLM_SIM_AIMD_MODEL}.initialBackoff=500ms"
+        --set "processor.config.modelGateways.${VLLM_SIM_AIMD_MODEL}.maxBackoff=5s"
         --set "processor.config.modelGateways.${VLLM_SIM_MODEL}.inferenceObjective=${GIE_OBJECTIVE_PREFIX}"
         --set "processor.config.modelGateways.${VLLM_SIM_B_MODEL}.inferenceObjective=${GIE_OBJECTIVE_PREFIX}"
         --set "processor.config.modelGateways.${VLLM_SIM_429_MODEL}.inferenceObjective=${GIE_OBJECTIVE_PREFIX}"
@@ -1301,6 +1308,7 @@ print_usage() {
     echo "       - sim-model-b   (vLLM simulator at ${VLLM_SIM_B_NAME})"
     echo "       - sim-model-429 (vLLM simulator at ${VLLM_SIM_429_NAME}, 50% rate_limit failure injection)"
     echo "       - sim-model-always-fail (vLLM simulator at ${VLLM_SIM_ALWAYS_FAIL_NAME}, 100% rate_limit failure injection)"
+    echo "       - sim-model-aimd (vLLM simulator at ${VLLM_SIM_AIMD_NAME}, AIMD recovery test — starts 100% rate_limit, patched to 0% mid-test)"
     if [ "${ENABLE_GIE}" = "true" ]; then
     echo ""
     echo "     GIE (flow control) is enabled:"
@@ -1397,6 +1405,8 @@ main() {
     install_vllm_sim "${VLLM_SIM_429_NAME}" "${VLLM_SIM_429_MODEL}" "10ms" "10ms" \
         "--failure-injection-rate=50" "--failure-types=rate_limit"
     install_vllm_sim "${VLLM_SIM_ALWAYS_FAIL_NAME}" "${VLLM_SIM_ALWAYS_FAIL_MODEL}" "10ms" "10ms" \
+        "--failure-injection-rate=100" "--failure-types=rate_limit"
+    install_vllm_sim "${VLLM_SIM_AIMD_NAME}" "${VLLM_SIM_AIMD_MODEL}" "10ms" "10ms" \
         "--failure-injection-rate=100" "--failure-types=rate_limit"
     if [ "${ENABLE_GIE}" = "true" ]; then
         ensure_gie_repo
--- a/test/e2e/aimd_test.go
+++ b/test/e2e/aimd_test.go
@@ -147,7 +147,7 @@ func doTestAIMDDecreaseAndIsolation(t *testing.T) {
 // Endpoint labels are gateway URLs like "http://vllm-sim.<ns>.svc.cluster.local:8000",
 // so matching "vllm-sim." (with trailing dot) excludes all hyphenated variants.
 func isHealthySimEndpoint(endpoint string) bool {
-	return strings.Contains(endpoint, "vllm-sim.")  &&
+	return strings.Contains(endpoint, "vllm-sim.") &&
 		!strings.Contains(endpoint, "vllm-sim-")
 }
 
--- a/test/e2e/aimd_test.go
+++ b/test/e2e/aimd_test.go
@@ -143,15 +143,12 @@ func doTestAIMDDecreaseAndIsolation(t *testing.T) {
 }
 
 // isHealthySimEndpoint returns true if the endpoint label refers to the
-// healthy sim-model simulator (vllm-sim) but not the 429, always-fail, aimd,
-// or sim-b variants.
+// primary vllm-sim simulator, not any variant (vllm-sim-429, vllm-sim-b, etc.).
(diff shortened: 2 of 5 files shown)


test/e2e/aimd_test.go
```diff
@@ -0,0 +1,350 @@
...
+			}
+			break
+		}
+	}
+	if aimdEndpoint == "" {
+		t.Fatal("no AIMD metric found for endpoint containing 'vllm-sim-aimd'")
+	}
+
+	// Phase 2: Patch simulator to 0% failure and wait for rollout.
+	t.Log("phase 2: patching simulator to 0% failure rate...")
```
lioraron, line 261: **Missing `t.Cleanup` — simulator stays at 0% failure if the test fails mid-recovery.**

If any `t.Fatalf` fires between here (phase 2 patch to 0%) and the cleanup block at line 316, the simulator remains at 0% failure, corrupting state for subsequent runs.

Register cleanup immediately after the patch succeeds:

```go
t.Log("phase 2: patching simulator to 0% failure rate...")
patchSimulatorFailureRate(t, aimdSimDeployment, 0)
t.Cleanup(func() {
    patchSimulatorFailureRate(t, aimdSimDeployment, 100)
    waitForRollout(t, aimdSimDeployment)
})
waitForRollout(t, aimdSimDeployment)
```

Then remove the manual cleanup at lines 316-319.

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,350 @@
...
+
+// patchSimulatorFailureRate patches the vllm-sim container args in the given
+// deployment to set --failure-injection-rate to the specified value. A rate of
+// 0 removes the failure injection flags entirely.
+func patchSimulatorFailureRate(t *testing.T, deployment string, rate int) {
+	t.Helper()
+
+	var patch string
+	if rate == 0 {
+		patch = `{"spec":{"template":{"spec":{"containers":[{"name":"vllm-sim","args":["--model","sim-model-aimd","--port","8000","--time-to-first-token=10ms","--inter-token-latency=10ms","--v=5"]}]}}}}`
```
lioraron: The patch JSON hardcodes `"sim-model-aimd"` as the `--model` arg. If someone overrides `TEST_MODEL_AIMD`, the simulator serves a model that doesn't match the gateway config.

Since `patchSimulatorFailureRate` already takes the deployment name, consider also taking the model name:

```go
func patchSimulatorFailureRate(t *testing.T, deployment, model string, rate int) {
```

and passing `testModelAIMD` from the caller.

Same applies to the hardcoded timing args (`10ms`, `10ms`) — they couple to `dev-deploy.sh`'s `install_vllm_sim` invocation. If the deploy defaults change, this patch silently overwrites them. At minimum, worth a comment noting the coupling.

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	if !foundNormalEndpoint {
+		t.Error("no AIMD metric found for a healthy (non-429) endpoint")
+	}
+}
+
+// isHealthySimEndpoint returns true if the endpoint label refers to the
+// healthy sim-model simulator (vllm-sim) but not the 429, always-fail, aimd,
+// or sim-b variants.
+func isHealthySimEndpoint(endpoint string) bool {
```
lioraron, line 151: nit: This exclusion-based approach needs updating whenever a new sim variant is added (e.g., `vllm-sim-slow`). A positive match would be more robust:

```go
func isHealthySimEndpoint(endpoint string) bool {
    // Match only the primary vllm-sim, not variants like vllm-sim-429, vllm-sim-b, etc.
    // Endpoint format: "<ip>:<port>" from pod in service vllm-sim.<ns>
    return strings.Contains(endpoint, "vllm-sim.") ||
        strings.HasSuffix(endpoint, "vllm-sim")
}
```

Or match against the service DNS pattern directly if the endpoint label includes it.

lioraron (commented):
## Review: AIMD e2e tests

**Rebase needed**: This branch has 11 commits, but the first 10 are from the already-merged PR #394 (same SHAs: `09d66eb`..`c3a36b2`). The branch diverged from main *before* #394's squash merge (`86b8324`), so GitHub shows all of PR #394's content in this diff. After `git rebase origin/main && git push --force-with-lease`, the diff will shrink from 3056 additions / 32 files to ~363 additions / 4 files.

The actual new content (commit `d324c8c`) is well-structured — clean three-phase recovery test, good reuse of existing helpers (`scrapeProcessorMetrics`, `waitForRetryExhaustion`, `waitForRollout`), and the probability math in the comments is solid.

Three issues to address:

test/e2e/aimd_test.go
j-mok-dev: Thanks. I removed the manual cleanup block and registered t.Cleanup func to recover to 100% failure rate.

test/e2e/aimd_test.go
j-mok-dev: updated to positive match: if it includes `vllm-sim.` and doesn't include `vllm-sim-`, then the func returns true.

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,352 @@
...
+//   - The healthy endpoint's AIMD concurrency limit stayed at perEndpoint.
+//   - The 429 endpoint accumulated AIMD decrease signals.
+//
+// With 50% failure rate and maxRetries=10, all requests eventually succeed,
+// but intermediate 429 responses trigger AIMD multiplicative decreases.
+// Two consecutive 429s are enough to hit the floor (10→5 with backoffFactor=0.5,
+// min=5), and the probability of avoiding two consecutive 429s across 10
+// requests is negligible ((1−0.5²)^9 ≈ 7.5%).
+func doTestAIMDDecreaseAndIsolation(t *testing.T) {
+	t.Helper()
```
yizhaodev: you don't need to call `t.Helper()` here for function invokded by `t.Run()`

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,352 @@
...
+//  2. Patches the simulator to 0% failure rate and waits for rollout.
+//  3. Submits more requests and verifies the limit increased above min.
+//
+// t.Cleanup restores the simulator to 100% failure for subsequent runs.
+//
+// Because AIMD increases by +1 per successful window, full recovery to
+// perEndpoint (10) requires many requests. We only assert limit > min to
+// avoid flakiness from timing variance.
+func doTestAIMDRecovery(t *testing.T) {
+	t.Helper()
```
yizhaodev: you don't need to call `t.Helper()` here for function invokded by `t.Run()`

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,352 @@
...
+	if !foundNormalEndpoint {
+		t.Error("no AIMD metric found for a healthy (non-429) endpoint")
+	}
+}
+
+// isHealthySimEndpoint returns true if the endpoint label refers to the
+// primary vllm-sim simulator, not any variant (vllm-sim-429, vllm-sim-b, etc.).
+// Endpoint labels are gateway URLs like "http://vllm-sim.<ns>.svc.cluster.local:8000",
+// so matching "vllm-sim." (with trailing dot) excludes all hyphenated variants.
+func isHealthySimEndpoint(endpoint string) bool {
```
yizhaodev, line 151: do we have model tag on metrics to filter out endpoint?

test/e2e/aimd_test.go
j-mok-dev: Thanks! removed.

test/e2e/aimd_test.go
j-mok-dev: removed. thanks!

test/e2e/aimd_test.go
j-mok-dev: no we don't. AIMD limits are per endpoint, not per-model. different models can share the same gateway endpoint, and the concurrency limit applies at that endpoint level.

scripts/dev-deploy.sh
```diff
@@ -1398,6 +1406,8 @@ main() {
         "--failure-injection-rate=50" "--failure-types=rate_limit"
     install_vllm_sim "${VLLM_SIM_ALWAYS_FAIL_NAME}" "${VLLM_SIM_ALWAYS_FAIL_MODEL}" "10ms" "10ms" \
         "--failure-injection-rate=100" "--failure-types=rate_limit"
+    install_vllm_sim "${VLLM_SIM_AIMD_NAME}" "${VLLM_SIM_AIMD_MODEL}" "10ms" "10ms" \
```
lioraron, line 1409: `dev-clean.sh` is missing the corresponding cleanup for this new simulator. Every other sim has an explicit `kubectl delete` at lines 43-46 of `dev-clean.sh`; without one for `vllm-sim-aimd`, `make dev-clean` leaves it orphaned.

```bash
kubectl delete deployment,svc "${VLLM_SIM_AIMD_NAME}" -n "${NAMESPACE}" --ignore-not-found=true
```

test/e2e/aimd_test.go
```diff
@@ -0,0 +1,348 @@
...
+		}
+	}
+	if aimdEndpoint == "" {
+		t.Fatal("no AIMD metric found for endpoint containing 'vllm-sim-aimd'")
+	}
+
+	// Phase 2: Patch simulator to 0% failure and wait for rollout.
+	t.Log("phase 2: patching simulator to 0% failure rate...")
+	patchSimulatorFailureRate(t, aimdSimDeployment, testModelAIMD, 0)
+	t.Cleanup(func() {
```
lioraron, line 263: `patchSimulatorFailureRate` calls `t.Fatalf` on failure, which inside `t.Cleanup` aborts the cleanup function — so `waitForRollout` never runs, and the simulator stays at 0% failure rate for the next test run.

`helm_upgrade_test.go:80` handles this by logging a "MANUAL RESTORE REQUIRED" message. Consider the same pattern here: use `t.Errorf` + early return in the cleanup so `waitForRollout` still runs, or inline the patch with `t.Errorf` instead of `t.Fatalf`.

scripts/dev-deploy.sh
j-mok-dev: Thanks. I missed it. added.

test/e2e/aimd_test.go
j-mok-dev: done. thanks.

lioraron (approved):

lioraron:
Thanks @j-mok-dev for implementing this!
Is this ready for merge?

j-mok-dev:
Merged! thanks :D

PR #430, opened 2026-05-12
feat: add async dispatch config

vishbhat:
## Why is this PR needed?

The batch-gateway currently only supports synchronous HTTP dispatch for inference requests. To enable integration with [llm-d-async](https://github.com/llm-d-incubation/llm-d-async) as an alternative asynchronous dispatch backend, the processor needs configuration plumbing before implementation can begin. This PR lays the config groundwork so the follow-up implementation PR can wire up the async dispatch path cleanly.

## What does this PR do?

- **Adds `AsyncDispatchConfig` struct** with fields for Redis connection (`redis_url`), queue names (`request_queue_name`, `result_queue_name`), result poll timeout, and per-request timeout (default: 60 minutes).
- **Adds `DispatchMode` field to `ProcessorConfig`** with constants `sync` (default) and `async` to select the inference dispatch backend.
- **Adds `IsAsync()` helper** on `ProcessorConfig` for convenient dispatch mode checks.
- **Adds validation** for async config fields: required fields when mode is `async`, positive timeouts, and `per_request_timeout > result_poll_timeout`. Sync mode and empty dispatch mode skip async validation.
- **Adds defaults** in `NewConfig()`: `DispatchMode` defaults to `sync`, `ResultPollTimeout` to 5s, `PerRequestTimeout` to 60m.
- **Adds unit tests** for all new validation cases (12 table-driven subtests), default value assertions, and `IsAsync()` behavior.
- **Adds `.cursor` to `.gitignore`**.

No runtime behaviour changes — async dispatch implementation will follow in a separate PR.

## Item for follow up PR

The per-pool queue design (`request:{pool_name}` / `result:{pool_name}`) from #432 would need per model queue routing, and the current single global `request_queue_name`/`result_queue_name` in `AsyncDispatchConfig` won't cover that. Plan to extend `ModelGatewayConfig` with a `pool_name` and adjust AsyncDispatchConfig accordingly. 

## How was this tested?

## Checklist

## Related Issues

Pre-requisite for https://github.com/llm-d-incubation/batch-gateway/issues/431

### DIFF
--- a/internal/processor/config/config.go
+++ b/internal/processor/config/config.go
@@ -358,24 +358,6 @@ func (c *ProcessorConfig) Validate() error {
 		return fmt.Errorf("e2e_latency_bucket must satisfy: start > 0, factor > 1, count > 0")
 	}
 
-	if c.GlobalInferenceGateway == nil && len(c.ModelGateways) == 0 {
-		return fmt.Errorf("either global_inference_gateway or model_gateways must be configured")
-	}
-	if c.GlobalInferenceGateway != nil && len(c.ModelGateways) > 0 {
-		return fmt.Errorf("global_inference_gateway and model_gateways are mutually exclusive")
-	}
-
-	if c.GlobalInferenceGateway != nil {
-		if err := validateGatewayConfig("global_inference_gateway", *c.GlobalInferenceGateway); err != nil {
-			return err
-		}
-	}
-	for model, gw := range c.ModelGateways {
-		if err := validateGatewayConfig(fmt.Sprintf("model_gateways[%s]", model), gw); err != nil {
-			return err
-		}
-	}
-
 	if err := c.FileClientCfg.Retry.Validate(); err != nil {
 		return fmt.Errorf("file_client.retry: %w", err)
 	}
@@ -394,12 +376,34 @@ func (c *ProcessorConfig) Validate() error {
 func (c *ProcessorConfig) validateDispatchMode() error {
 	switch c.DispatchMode {
 	case DispatchModeSync, DispatchMode(""):
-		return nil
+		c.DispatchMode = DispatchModeSync
+		return c.validateSyncConfig()
 	case DispatchModeAsync:
 		return c.AsyncConfig.validate()
 	default:
 		return fmt.Errorf("dispatch_mode must be %q or %q, got %q", DispatchModeSync, DispatchModeAsync, c.DispatchMode)
 	}
+}
+
+func (c *ProcessorConfig) validateSyncConfig() error {
+	if c.GlobalInferenceGateway == nil && len(c.ModelGateways) == 0 {
+		return fmt.Errorf("either global_inference_gateway or model_gateways must be configured")
+	}
+	if c.GlobalInferenceGateway != nil && len(c.ModelGateways) > 0 {
+		return fmt.Errorf("global_inference_gateway and model_gateways are mutually exclusive")
+	}
+
+	if c.GlobalInferenceGateway != nil {
+		if err := validateGatewayConfig("global_inference_gateway", *c.GlobalInferenceGateway); err != nil {
+			return err
+		}
+	}
+	for model, gw := range c.ModelGateways {
+		if err := validateGatewayConfig(fmt.Sprintf("model_gateways[%s]", model), gw); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (ac *AsyncDispatchConfig) validate() error {
--- a/internal/processor/config/config.go
+++ b/internal/processor/config/config.go
@@ -73,18 +73,21 @@ type AIMDConfig struct {
 	AdditiveIncrease int `yaml:"additive_increase"`
 }
 
-// DispatchMode constants control which inference dispatch backend is used.
+// DispatchMode selects the inference dispatch backend.
+type DispatchMode string
+
 const (
-	DispatchModeSync  = "sync"
-	DispatchModeAsync = "async"
+	DispatchModeSync  DispatchMode = "sync"
+	DispatchModeAsync DispatchMode = "async"
 )
 
 // AsyncDispatchConfig holds configuration for the llm-d-async dispatch backend.
 // Only used when DispatchMode == "async".
 type AsyncDispatchConfig struct {
 	// RedisURL is the full Redis connection URL (e.g. "redis://host:6379",
-	// "rediss://user:pass@host:6379" for TLS). TLS is encoded in the URL scheme.
-	RedisURL string `yaml:"redis_url"`
+	// "rediss://user:pass@host:6379" for TLS). Read from the mounted secret
+	// at runtime (SecretKeyRedisURL)
+	RedisURL string `yaml:"-"`
 
 	// RequestQueueName is the Redis sorted-set name for request submission.
 	// Must match the llm-d-async processor's --redis.ss.request-queue-name.
@@ -97,11 +100,6 @@ type AsyncDispatchConfig struct {
 	// ResultPollTimeout is the timeout per GetResult poll cycle.
 	// Controls how long each blocking poll waits before retrying.
 	ResultPollTimeout time.Duration `yaml:"result_poll_timeout"`
-
-	// PerRequestTimeout is the maximum time to wait for a single request's
-	// result before marking it as failed. Should be significantly larger than
-	// expected inference latency to avoid false positives.
-	PerRequestTimeout time.Duration `yaml:"per_request_timeout"`
 }
 
 type ProcessorConfig struct {
@@ -180,7 +178,7 @@ type ProcessorConfig struct {
 	// DispatchMode selects the inference dispatch backend.
 	// "sync" (default): direct HTTP via InferenceClient.
 	// "async": submit via llm-d-async producer, collect from result queue.
-	DispatchMode string `yaml:"dispatch_mode"`
+	DispatchMode DispatchMode `yaml:"dispatch_mode"`
 
 	// AsyncConfig holds llm-d-async dispatch settings. Only used when DispatchMode == "async".
 	AsyncConfig AsyncDispatchConfig `yaml:"async"`
@@ -318,7 +316,6 @@ func NewConfig() *ProcessorConfig {
 		DispatchMode: DispatchModeSync,
 		AsyncConfig: AsyncDispatchConfig{
 			ResultPollTimeout: 5 * time.Second,
-			PerRequestTimeout: 60 * time.Minute,
 		},
 	}
 }
@@ -396,7 +393,7 @@ func (c *ProcessorConfig) Validate() error {
 
 func (c *ProcessorConfig) validateDispatchMode() error {
 	switch c.DispatchMode {
-	case DispatchModeSync, "":
+	case DispatchModeSync, DispatchMode(""):
 		return nil
 	case DispatchModeAsync:
 		return c.AsyncConfig.validate()
@@ -406,9 +403,6 @@ func (c *ProcessorConfig) validateDispatchMode() error {
 }
 
 func (ac *AsyncDispatchConfig) validate() error {
-	if ac.RedisURL == "" {
-		return fmt.Errorf("async.redis_url must be set when dispatch_mode is %q", DispatchModeAsync)
-	}
 	if ac.RequestQueueName == "" {
 		return fmt.Errorf("async.request_queue_name must be set when dispatch_mode is %q", DispatchModeAsync)
 	}
@@ -417,13 +411,6 @@ func (ac *AsyncDispatchConfig) validate() error {
 	}
 	if ac.ResultPollTimeout <= 0 {
 		return fmt.Errorf("async.result_poll_timeout must be > 0")
-	}
-	if ac.PerRequestTimeout <= 0 {
-		return fmt.Errorf("async.per_request_timeout must be > 0")
-	}
-	if ac.PerRequestTimeout <= ac.ResultPollTimeout {
-		return fmt.Errorf("async.per_request_timeout (%v) must be > async.result_poll_timeout (%v)",
-			ac.PerRequestTimeout, ac.ResultPollTimeout)
 	}
 	return nil
 }
--- a/internal/processor/config/config.go
+++ b/internal/processor/config/config.go
@@ -73,6 +73,37 @@ type AIMDConfig struct {
 	AdditiveIncrease int `yaml:"additive_increase"`
 }
 
+// DispatchMode constants control which inference dispatch backend is used.
+const (
+	DispatchModeSync  = "sync"
+	DispatchModeAsync = "async"
+)
+
+// AsyncDispatchConfig holds configuration for the llm-d-async dispatch backend.
+// Only used when DispatchMode == "async".
+type AsyncDispatchConfig struct {
+	// RedisURL is the full Redis connection URL (e.g. "redis://host:6379",
+	// "rediss://user:pass@host:6379" for TLS). TLS is encoded in the URL scheme.
+	RedisURL string `yaml:"redis_url"`
+
+	// RequestQueueName is the Redis sorted-set name for request submission.
+	// Must match the llm-d-async processor's --redis.ss.request-queue-name.
+	RequestQueueName string `yaml:"request_queue_name"`
+
+	// ResultQueueName is the Redis list name for result collection.
+	// Must match the llm-d-async processor's --redis.ss.result-queue-name.
+	ResultQueueName string `yaml:"result_queue_name"`
+
+	// ResultPollTimeout is the timeout per GetResult poll cycle.
+	// Controls how long each blocking poll waits before retrying.
+	ResultPollTimeout time.Duration `yaml:"result_poll_timeout"`
+
+	// PerRequestTimeout is the maximum time to wait for a single request's
+	// result before marking it as failed. Should be significantly larger than
+	// expected inference latency to avoid false positives.
+	PerRequestTimeout time.Duration `yaml:"per_request_timeout"`
+}
+
 type ProcessorConfig struct {
 	// TaskWaitTime is the timeout parameter used when dequeueing from the priority queue
 	// This should be shorter than PollInterval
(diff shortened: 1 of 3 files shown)


internal/processor/config/config.go
```diff
@@ -73,6 +73,37 @@ type AIMDConfig struct {
 	AdditiveIncrease int `yaml:"additive_increase"`
 }
 
+// DispatchMode constants control which inference dispatch backend is used.
+const (
```
yizhaodev, line 79: define type please
`type DispatchMode string`

internal/processor/config/config.go
```diff
@@ -73,6 +73,37 @@ type AIMDConfig struct {
...
+	DispatchModeSync  = "sync"
+	DispatchModeAsync = "async"
+)
+
+// AsyncDispatchConfig holds configuration for the llm-d-async dispatch backend.
+// Only used when DispatchMode == "async".
+type AsyncDispatchConfig struct {
+	// RedisURL is the full Redis connection URL (e.g. "redis://host:6379",
+	// "rediss://user:pass@host:6379" for TLS). TLS is encoded in the URL scheme.
+	RedisURL string `yaml:"redis_url"`
```
yizhaodev: should be ``yaml:"-"``
redis_url shall be read from secret instead of config like others in batch gateway

internal/processor/config/config.go
```diff
@@ -145,6 +176,14 @@ type ProcessorConfig struct {
...
 	// FileClient holds configuration for the shared file storage client (fs or s3).
 	FileClientCfg sharedcfg.FileClientConfig `yaml:"file_client"`
+
+	// DispatchMode selects the inference dispatch backend.
+	// "sync" (default): direct HTTP via InferenceClient.
+	// "async": submit via llm-d-async producer, collect from result queue.
+	DispatchMode string `yaml:"dispatch_mode"`
+
+	// AsyncConfig holds llm-d-async dispatch settings. Only used when DispatchMode == "async".
+	AsyncConfig AsyncDispatchConfig `yaml:"async"`
```
yizhaodev: currently, processor support inference gateway for each model
```
model_gateways:
  "llama-3":
    url: "http://gateway-a:8000"
  "mistral":
    url: "http://gateway-b:8000"
```

But in llm-d-async, each llm-d-async instance is bounded to one inference gateway (`--redis.ss.igw-base-url`)

how do we want to handle inference gateway per model scenrio? 
 - option 1: only support 1 global inference gateway
 - option 2: deploy multiple llm-d-async for each model
 - option 3: update  llm-d-async, to support passing `igw-base-url` in request

@lioraron @vishbhat what you think?

internal/processor/config/config.go
```diff
@@ -337,6 +384,34 @@ func (c *ProcessorConfig) Validate() error {
...
+		return err
+	}
+
+	return nil
+}
+
+func (c *ProcessorConfig) validateDispatchMode() error {
+	switch c.DispatchMode {
+	case DispatchModeSync, DispatchMode(""):
+		return nil
```
yizhaodev: `ProcessorConfig.VAlidate()` check a lot of parameters
some of them should be moved there if they are required by Sync mode only

internal/processor/config/config.go
```diff
@@ -73,6 +73,35 @@ type AIMDConfig struct {
...
+	DispatchModeAsync DispatchMode = "async"
+)
+
+// AsyncDispatchConfig holds configuration for the llm-d-async dispatch backend.
+// Only used when DispatchMode == "async".
+type AsyncDispatchConfig struct {
+	// RedisURL is the full Redis connection URL (e.g. "redis://host:6379",
+	// "rediss://user:pass@host:6379" for TLS). Read from the mounted secret
+	// at runtime (SecretKeyRedisURL)
+	RedisURL string `yaml:"-"`
```
yizhaodev, line 90: how you plan to set `RedisSortedSetConfig.TenantID` later, harcode?

vishbhat: `TenantID` is available per job at runtime via `jobInfo.TenantID`, so we pass it through at the `SubmitRequest` call. Thats the plan for now.

yizhaodev: i think you call `NewRedisSortedSetProducer` and reuse the producer instance?
and `NewRedisSortedSetProducer` requires TenantID
https://github.com/llm-d-incubation/llm-d-async/blob/2b12d4c96eeb1d5209f2f7f979bc6101a539d053/producer/redis_sortedset_producer.go#L70


or you plan to have prodcuer for each tenant?

internal/processor/config/config.go
```diff
@@ -337,6 +384,34 @@ func (c *ProcessorConfig) Validate() error {
...
+	case DispatchModeSync, DispatchMode(""):
+		return nil
+	case DispatchModeAsync:
+		return c.AsyncConfig.validate()
+	default:
+		return fmt.Errorf("dispatch_mode must be %q or %q, got %q", DispatchModeSync, DispatchModeAsync, c.DispatchMode)
+	}
+}
+
+func (ac *AsyncDispatchConfig) validate() error {
```
j-mok-dev: maybe it's better to have a `RedisURL` check here, for example, `if ac.RedisURL == ""`. so we can make a missing secret mount fails fast at startup.

internal/processor/config/config.go
```diff
@@ -337,6 +384,34 @@ func (c *ProcessorConfig) Validate() error {
...
+	if err := c.validateDispatchMode(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (c *ProcessorConfig) validateDispatchMode() error {
+	switch c.DispatchMode {
+	case DispatchModeSync, DispatchMode(""):
```
j-mok-dev, line 386: nit: `validateDispatchMode` currently treats `""` as sync but doesn't normalize the value. `IsAsync()` works fine, but if other code checks `c.DispatchMode == DispatchModeSync`, `""` would return false. 

we can change this to 
```
case DispatchModeSync, DispatchMode(""):
    c.DispatchMode = DispatchModeSync
    return nil
```
to fix this issue.

internal/processor/config/config.go
j-mok-dev: Looking at the queue design in #432, each inference pool gets its own request/result queue pair (`request:{pool_name}` / `result:{pool_name}`). Since we already support per-model gateways in `model_gateways`, we could extend it with a `pool_name` field:
```yaml
model_gateways:
  "llama-3":
    url: "http://gateway-a:8000"    # used in sync mode
    pool_name: "pool-a"             # used in async mode to derive queue names
  "mistral":
    url: "http://gateway-b:8000"
    pool_name: "pool-b"
```
This way the existing multi-model structure works for both sync (uses url) and async (uses pool_name to enqueue into the right queue). The current AsyncDispatchConfig with a single global `request_queue_name`/`result_queue_name` wouldn't cover this scenario.

internal/processor/config/config.go
vishbhat: My initial drafted plan is to use a single shared producer instance with a fixed TenantID. 
Whats your suggestion here? To add a config field for `TenantID` ?

internal/processor/config/config.go
vishbhat: The `RedisURL` secret is read later in `clientset.New()`. So for the `RedisURL` check to work at validation time, the secret needs to be loaded into `AsyncConfig.RedisURL` before `Validate()`. I probably have to load the secret in `main.go` ?
@j-mok-dev Any examples of how its done for other secrets?

internal/processor/config/config.go
yizhaodev: what will the producer's TenantID be used for?

internal/processor/config/config.go
j-mok-dev: you're right. looking at the existing pattern in `clientset.go`, secrets (redis-url, postgresql-url, s3-secret-access-key) are all read at client initialization time via `ucom.ReadSecretFile()`, not at config validation. So the same approach would work here: read `SecretKeyRedisURL` when creating the async producer client, and fail there if it's empty. No need to special-case it in `Validate()`.

internal/processor/config/config.go
vishbhat: Afaik its used for prefixing Redis keys for result queue isolation.

yizhaodev: @evacchi 's design is to use "pool-name based" queue
if you put tenant id into queue name, that will conflit with his design
please discuss with him first?

vishbhat: @yizhaodev The plan for now is to hardcode the `TenantID` to something like `batch-gateway` in the code.

internal/processor/config/config.go
vishbhat: @yizhaodev @j-mok-dev Thats a good point. 
For this PR, I'd like to keep the config minimal. Need to analyze in detail to plan the config structure. Once the design solidifies, we can extend `ModelGatewayConfig` with a `pool_name` field (used in async mode to derive queue names) and adjust `AsyncDispatchConfig` accordingly in a followup PR.

yizhaodev: thanks
if this is a temp solution and will rewrite eventually. can we just keep it in development branch instead of merging into main?

vishbhat: Sure. This PR was only to enable config items to so that the operator changes are aware of incoming config.

internal/processor/config/config.go
```diff
@@ -189,6 +223,23 @@ type BucketConfig struct {
 	BucketCount  int     `yaml:"count"`
 }
 
+const asyncTenantID = "batch-gateway"
```
lioraron: Two misalignments with the companion design doc (PR #432) that need resolution before both PRs land:

1. **Queue prefix**: singular `request:` / `result:` here vs plural `requests:` / `results:` in #432's primary convention (the "Queue Naming Convention" table and all subsequent references).
2. **Tenant ID**: `"batch-gateway"` here vs `"$batch"` in #432 ("the batch-processor uses the reserved value `$batch`").

These produce different Redis keys, so the batch-processor and dispatcher would read/write different queues.

internal/processor/config/config.go
```diff
@@ -181,6 +210,11 @@ type ModelGatewayConfig struct {
 	TLSCACertFile         string `yaml:"tls_ca_cert_file,omitempty"`
 	TLSClientCertFile     string `yaml:"tls_client_cert_file,omitempty"`
 	TLSClientKeyFile      string `yaml:"tls_client_key_file,omitempty"`
+
+	// PoolName identifies the async dispatch pool for this model/gateway.
+	// Queue names are derived as "request:{PoolName}" / "result:{PoolName}".
```
lioraron: Comment says `"result:{PoolName}"` but `ResultQueueName()` actually produces `"result:{tenant}:{PoolName}"` — the tenant ID segment is missing.

```suggestion
	// Queue names are derived as "request:{PoolName}" / "result:{tenant}:{PoolName}".
```

evacchi: I think we will be using `results:...` see #432

evacchi (approved):
nothing to add to the existing comments, overall fine once the rest is addressed

internal/processor/config/config.go
vishbhat: Updated the config in this PR.

internal/processor/config/config.go
vishbhat: Thanks @lioraron. Addressed both of them.

internal/processor/config/config.go
```diff
@@ -145,6 +166,14 @@ type ProcessorConfig struct {
...
 	// FileClient holds configuration for the shared file storage client (fs or s3).
 	FileClientCfg sharedcfg.FileClientConfig `yaml:"file_client"`
+
+	// DispatchMode selects the inference dispatch backend.
+	// "sync" (default): direct HTTP via InferenceClient.
+	// "async": submit via llm-d-async producer, collect from result queue.
+	DispatchMode DispatchMode `yaml:"dispatch_mode"`
+
+	// AsyncConfig holds llm-d-async dispatch settings. Only used when DispatchMode == "async".
+	AsyncConfig AsyncDispatchConfig `yaml:"async"`
```
lioraron: Can we call this variable `AsyncDispatchConfig` and yaml tag `async_dispatch`?

internal/processor/config/config.go
```diff
@@ -181,6 +210,11 @@ type ModelGatewayConfig struct {
 	TLSCACertFile         string `yaml:"tls_ca_cert_file,omitempty"`
 	TLSClientCertFile     string `yaml:"tls_client_cert_file,omitempty"`
 	TLSClientKeyFile      string `yaml:"tls_client_key_file,omitempty"`
+
+	// PoolName identifies the async dispatch pool for this model/gateway.
+	// Queue names are derived as "requests:{PoolName}" / "results:{tenant}:{PoolName}".
+	// Required when dispatch_mode is "async". Ignored in sync mode.
+	PoolName string `yaml:"pool_name"`
```
lioraron: Can we call this variable `InferencePoolName` and yaml tag `inference_pool_name`?

internal/processor/config/config.go
```diff
@@ -181,6 +210,11 @@ type ModelGatewayConfig struct {
 	TLSCACertFile         string `yaml:"tls_ca_cert_file,omitempty"`
 	TLSClientCertFile     string `yaml:"tls_client_cert_file,omitempty"`
 	TLSClientKeyFile      string `yaml:"tls_client_key_file,omitempty"`
+
+	// PoolName identifies the async dispatch pool for this model/gateway.
+	// Queue names are derived as "requests:{PoolName}" / "results:{tenant}:{PoolName}".
```
lioraron: Suggest to drop the queue name format comment — the format is already expressed by `RequestQueueName()` / `ResultQueueName()` below, and duplicating it here risks stale documentation when the naming convention changes. Keep:
```
// InferencePoolName identifies the async dispatch pool for this model/gateway.
// Required when dispatch_mode is "async". Ignored in sync mode.
```

internal/processor/config/config.go
```diff
@@ -189,6 +223,23 @@ type BucketConfig struct {
 	BucketCount  int     `yaml:"count"`
 }
 
+const asyncTenantID = "$batch"
+
+// RequestQueueName returns the Redis sorted-set name for submitting async requests to the given pool.
+func RequestQueueName(poolName string) string {
+	return "requests:" + poolName
```
lioraron: Consider adding a namespace prefix for the async processor to avoid Redis key collisions. The Redis instance may likely be shared with other systems. Something like:
```
return "llm-d-async:requests:" + poolName
```
This applies to both `RequestQueueName` and `ResultQueueName`. The prefix has to be aligned with the config on the llm-d-async consumer side.

cc @evacchi

internal/processor/config/config.go
```diff
@@ -189,6 +223,23 @@ type BucketConfig struct {
...
+}
+
+// ResultQueueName returns the Redis list name for collecting async results from the given pool.
+func ResultQueueName(poolName string) string {
+	return "results:" + asyncTenantID + ":" + poolName
```
lioraron: Why is the tenant ID in the middle rather than at the end? The current format breaks the symmetry with the request queue. Suggest `llm-d-async:results:{pool}:{tenant}`.
Putting tenant last mirrors the request queue structure with an added suffix, and groups by pool first in Redis key scans — which matches how the dispatcher works (per-pool).

cc @evacchi

internal/processor/config/config.go
```diff
@@ -311,12 +367,47 @@ func (c *ProcessorConfig) Validate() error {
...
 	if c.GlobalInferenceGateway == nil && len(c.ModelGateways) == 0 {
 		return fmt.Errorf("either global_inference_gateway or model_gateways must be configured")
 	}
 	if c.GlobalInferenceGateway != nil && len(c.ModelGateways) > 0 {
 		return fmt.Errorf("global_inference_gateway and model_gateways are mutually exclusive")
 	}
+	return nil
+}
+
+func (c *ProcessorConfig) validateSyncConfig() error {
```
lioraron: validateSyncDispatchConfig?

internal/processor/config/config.go
```diff
@@ -328,15 +419,26 @@ func (c *ProcessorConfig) Validate() error {
 			return err
 		}
 	}
+	return nil
+}
 
-	if err := c.FileClientCfg.Retry.Validate(); err != nil {
-		return fmt.Errorf("file_client.retry: %w", err)
+func (c *ProcessorConfig) validateAsyncConfig() error {
```
lioraron: validateAsyncDispatchConfig?

internal/processor/config/config.go
vishbhat: @lioraron This is just to match llm-d async producers format.  If we need to change this, we need to change it in the async repo as well.

internal/processor/config/config.go
vishbhat: @lioraron The prefix does not exist in llm-d async, the names are already configurable via `inference_pool_name` so something like `llm-d-async:pool-a` could be used for the pool name. And hardcoding a prefix would break interop unless llm-d-async adopts the same convention.

lioraron: The concern isn't about `inference_pool_name` — it's about what `RequestQueueName()` / `ResultQueueName()` prepend to it. Right now these produce `requests:pool-a` which is generic enough to collide with other keys in a shared Redis instance. Since llm-d-async takes the full queue name as config (it's just a string), adding a prefix here doesn't break interop — it just means the dispatcher config uses the same prefixed name. The batch-gateway controls the convention; llm-d-async doesn't enforce one.

Suggest:
```go
func RequestQueueName(poolName string) string {
    return "llm-d-async:requests:" + poolName
}
```

internal/processor/config/config.go
lioraron: I checked the llm-d-async code — `queueConfig.ResultQueueName` is a free-form string, there's no enforced `{tenant}:{pool}` format. The ordering is ours to choose. Putting tenant last keeps the hierarchy consistent:
```
llm-d-async:requests:{pool}
llm-d-async:results:{pool}:{tenant}
```
This groups all keys for a pool together in Redis key scans (`llm-d-async:*:pool-a*`), which is useful for debugging and monitoring.

internal/processor/config/config.go
vishbhat: Understood. Thanks!

internal/processor/config/config.go
vishbhat: > there's no enforced {tenant}:{pool} format.

This is not true. 
Ref: https://github.com/llm-d-incubation/llm-d-async/blob/5b81d66567532e64d2ccb3017ab9115dc7ea8b55/producer/redis_sortedset_producer.go#L86
Tenant always comes before the queue name, its hardcoded.

lioraron: You're right. I opened [llm-d-async#200](https://github.com/llm-d-incubation/llm-d-async/issues/200) to make the producer accept a full result queue name - like the consumer already does. Could you follow up on this?


For this PR, can we update the `RequestQueueName/ResultQueueName` now to use the target convention (`llm-d-async:requests:{pool}`, `llm-d-async:results:{pool}:{tenant}`), anticipating the producer change? The helpers aren't wired to runtime code yet so there's no compatibility concern.

vishbhat: Sure. Will make the change and follow up on the the issue.

lioraron (approved):
Thanks!

PR #433, opened 2026-05-14
feat: handle x-llm-d-request-dropped-reason header from GIE

### LINKED ISSUE #311: [flow-control] Handle GIE flow-control queue eviction responses
## Summary

When GIE's flow control rejects or evicts a request from its queue, it returns specific HTTP error codes. The batch processor should recognize these as flow-control-specific conditions and handle them appropriately, rather than treating them as generic inference failures.

## Background

GIE's flow control can reject or evict requests in two scenarios:

1. **Queue at capacity** (`ErrQueueAtCapacity`): The flow control queue is full and cannot accept new requests. This is a transient condition.
2. **TTL expiry**: A request waited too long in the queue and was evicted. This may align with the batch job's SLO deadline expiry.

Currently, the batch processor would treat these as generic HTTP errors. With dedicated handling, the processor can make smarter decisions:

- **Queue at capacity**: Retry the request (the existing retry-on-429 logic may already handle this if GIE returns 429, but this needs verification).
- **TTL eviction**: Mark the request as `batch_expired` if the eviction aligns with the SLO deadline, avoiding unnecessary retries of a request that has already timed out.

## Proposed Changes

1. **Identify GIE's eviction response codes/headers** — determine the exact HTTP status codes and response body format GIE uses for queue capacity rejection and TTL eviction. This depends on GIE's implementation being finalized.

2. **Update error categorization** in `pkg/clients/http/http_client.go` to recognize flow-control-specific responses and map them to appropriate batch processor actions.

3. **Add tests** for the new error categorization paths.

## Dependencies

- GIE's eviction response codes need to be finalized and documented.

## Related

- Parent issue: #301
- GIE flow control: https://github.com/kubernetes-sigs/gateway-api-inference-extension/tree/main/pkg/epp/flowcontrol
- Batch Gateway flow control guide: `docs/guides/flow-control-setup.md`

lioraron:
## Summary

- When GIE returns 429 with `x-llm-d-request-dropped-reason: rejected-ttl-expired`, skip retry and write the request as `batch_expired` instead of surfacing an HTTP 429 error
- For all other dropped reasons (e.g. `rejected-saturated`), retry behavior is unchanged
- Uses context-value pattern (same as existing `NewCapacityRetryContext`) to propagate the header value out of `Post()` without changing its signature

## Changes

| File | Change |
|------|--------|
| `pkg/clients/http/http_client.go` | Header/reason constants, `NewDroppedReasonContext`, retry condition update, header capture in `Post()` |
| `pkg/clients/http/http_client_errors.go` | `DroppedReason` field on `ClientError` |
| `pkg/clients/inference/inference_client_impl.go` | Create dropped-reason context, populate `DroppedReason` on error |
| `internal/processor/worker/executor.go` | Map `rejected-ttl-expired` to `batch_expired` |
| `pkg/clients/http/http_client_test.go` | Retry-skip tests, context tracking tests |
| `pkg/clients/inference/inference_client_test.go` | `DroppedReason` propagation test |
| `internal/processor/worker/executor_test.go` | TTL-expired → `batch_expired` test, non-TTL 429 test |

## Backwards compatibility

All new code paths are no-ops when the header is absent — existing behavior is preserved for inference backends that don't send the header.

## Depends on

- [llm-d/llm-d-inference-scheduler#1027](https://github.com/llm-d/llm-d-inference-scheduler/pull/1027) (merged) — adds the `x-llm-d-request-dropped-reason` header to GIE responses

Closes #311

## Test plan

- [x] `TestDroppedReasonHandling` — 429 + `rejected-ttl-expired` is not retried (1 attempt); 429 + other reasons are retried; context captures header value
- [x] `TestInferenceClient/DroppedReason` — `Generate()` populates `ClientError.DroppedReason` from the response header
- [x] `TestExecuteOneRequest_DroppedReasonTTLExpired` — writes `batch_expired` error
- [x] `TestExecuteOneRequest_DroppedReasonOther` — non-TTL 429 produces normal HTTP error response
- [x] `make pre-commit` passes (formatting, vet, lint, unit tests, security scan)

### DIFF
--- a/pkg/clients/http/http_client.go
+++ b/pkg/clients/http/http_client.go
@@ -39,6 +39,9 @@ import (
 const (
 	HeaderNameReqID       string = "X-Request-ID"
 	HeaderNameContentType string = "Content-Type"
+	HeaderDroppedReason   string = "x-llm-d-request-dropped-reason"
+
+	DroppedReasonTTLExpired = "rejected-ttl-expired"
 
 	// rateLimitBackoffMultiplier scales the initial backoff for 429 responses.
 	// A 429 means the server explicitly asks the client to slow down, so we
@@ -57,6 +60,21 @@ var capacityRetryCtxKey capacityRetryKeyType
 func NewCapacityRetryContext(ctx context.Context) (context.Context, func() bool) {
 	var flag atomic.Bool
 	return context.WithValue(ctx, capacityRetryCtxKey, &flag), flag.Load
+}
+
+type droppedReasonKeyType struct{}
+
+var droppedReasonCtxKey droppedReasonKeyType
+
+// NewDroppedReasonContext returns a child context that captures the value of the
+// x-llm-d-request-dropped-reason response header from the final HTTP response.
+// Call the returned function after the request completes to read the result.
+func NewDroppedReasonContext(ctx context.Context) (context.Context, func() string) {
+	var reason atomic.Value
+	return context.WithValue(ctx, droppedReasonCtxKey, &reason), func() string {
+		v, _ := reason.Load().(string)
+		return v
+	}
 }
 
 // HTTPClient implements HTTP client with retry, TLS, and observability support
@@ -160,15 +178,19 @@ func NewHTTPClient(config Config, logger logr.Logger) (*HTTPClient, error) {
 			SetRetryMaxWaitTime(config.MaxBackoff).                                    // Max wait time between retries
 			SetRetryAfter(newRetryAfterFunc(config.InitialBackoff, config.MaxBackoff)) // Status-code-aware backoff
 
-		// Retry condition: retry on server errors, rate limits, and network errors
+		// Retry condition: retry on server errors, rate limits, and network errors.
+		// Exception: 429 with x-llm-d-request-dropped-reason: rejected-ttl-expired
+		// is not retried because the SLO already expired on the server side.
 		client.AddRetryCondition(func(r *resty.Response, err error) bool {
 			if err != nil {
 				return true // Retry on network errors
 			}
 
 			statusCode := r.StatusCode()
-			// Retry on 429 (rate limit) and 5xx (server errors)
-			return statusCode == http.StatusTooManyRequests || statusCode >= http.StatusInternalServerError
+			if statusCode == http.StatusTooManyRequests {
+				return r.Header().Get(HeaderDroppedReason) != DroppedReasonTTLExpired
+			}
+			return statusCode >= http.StatusInternalServerError
 		})
 
 		// Add retry hook for logging and capacity-retry tracking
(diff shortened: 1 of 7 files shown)


j-mok-dev:
Thanks for the header implementation! I have a few questions.
Checking the current GIE code, I couldn't find an implementation of the `x-llm-d-request-dropped-reason` header yet.

From the current code, the semantics seem to be:
- capacity rejected -> 429
- TTL eviction / request timed out in queue -> 503

So I'd like to confirm the intended contract once this header is added:

1. Will `rejected-ttl-expired` be returned only with 429, or also with 503?
2. Is the dropped-reason header intended to be more authoritative than the HTTP status code?
3. In batch-gateway, should `rejected-ttl-expired` always map to `batch_expired` regardless of status code, or only for a specific status/header combination?

lioraron:
Good questions! The header is already implemented in [llm-d-inference-scheduler](https://github.com/llm-d/llm-d-inference-scheduler) (not gateway-api-inference-extension — that may have been the source of confusion).

From the scheduler code (`pkg/common/error/error.go`), the contract is:

| Reason | Header value | HTTP status |
|--------|-------------|-------------|
| Capacity saturated | `rejected-saturated` | 429 (ResourceExhausted) |
| TTL expired | `rejected-ttl-expired` | 503 (ServiceUnavailable) |
| Client disconnected | `rejected-context-cancelled` | 503 (ServiceUnavailable) |

To answer your questions:

1. `rejected-ttl-expired` is returned with **503**, not 429.
2. Yes — the header is more authoritative. The status code tells you the request failed; the header tells you *why*. A 503 alone can't distinguish TTL expiration from a client disconnect.
3. `rejected-ttl-expired` maps to `batch_expired` regardless of status code. The semantic meaning is the same: the request's SLO deadline passed while queued, so retrying would be wasteful.

pkg/clients/http/http_client.go
```diff
@@ -160,15 +178,19 @@ func NewHTTPClient(config Config, logger logr.Logger) (*HTTPClient, error) {
...
 				return true // Retry on network errors
 			}
 
 			statusCode := r.StatusCode()
-			// Retry on 429 (rate limit) and 5xx (server errors)
-			return statusCode == http.StatusTooManyRequests || statusCode >= http.StatusInternalServerError
+			if statusCode == http.StatusTooManyRequests {
+				return r.Header().Get(HeaderDroppedReason) != DroppedReasonTTLExpired
+			}
+			return statusCode >= http.StatusInternalServerError
```
j-mok-dev, line 193: If `rejected-ttl-expired` comes back with 503, this still retries via the generic 5xx path. Should TTL-expired be treated as non-retryable regardless of status code if the header is authoritative?

pkg/clients/http/http_client_test.go
```diff
@@ -1326,6 +1326,128 @@ func TestCapacityRetryTracking(t *testing.T) {
 	})
 }
 
+func TestDroppedReasonHandling(t *testing.T) {
```
j-mok-dev, line 1329: Should this be `503` instead of `429`? In the router contract, `rejected-ttl-expired` is returned with `ServiceUnavailable`.

pkg/clients/http/http_client.go
lioraron: Good catch! Fixed in a187057: the retry condition now checks the header on both 429 and 503.
Scoped it to those two status codes rather than all 5xx, since other server errors (500, 502, 504) are genuine failures that should always be retried regardless of headers.

pkg/clients/http/http_client_test.go
lioraron: Right. Added a 503 test case and kept the 429 one, since the GIE eviction path uses 429. Fixed in a187057.

j-mok-dev (approved):
Thanks! LGTM.

Issue #434, opened 2026-05-14
[orphan-recovery] feat: orphan job detection and reconciliation for pod crash recovery


lioraron:
## Problem

When a batch-processor pod crashes (OOM kill, node eviction, pod deletion), jobs that were dequeued from the Redis priority queue can become **orphaned** — stuck in a non-terminal DB status with no processor working on them.

The existing startup recovery ([recovery.go](internal/processor/worker/recovery.go)) only handles **container restarts within the same pod** (it scans the local `emptyDir` workdir). If the pod is **replaced** (e.g., node eviction destroys `emptyDir`), those jobs are lost forever.

### Orphan scenarios

1. **Pod replacement**: Processor dequeues a job, then the pod is replaced. Workdir is gone; startup recovery on new pods can't find it.
2. **Silent abandonment**: A worker fails to re-enqueue or transition a job to a terminal state (e.g., DB unreachable during error handling), logs the error and moves on. The pod is healthy, so startup recovery won't run.

### The orphan window

After \`PQDequeue\` atomically removes a job from Redis, if the pod crashes before the worker creates a workdir and transitions the status:
- **Queue**: Job removed (won't be re-dequeued)
- **Workdir**: Doesn't exist yet (no startup recovery can find it)
- **DB**: Stuck with non-terminal status forever

### Prior work

- PR #359 — Architecture proposal document (4-layer reconciliation design)
- PR #135 — Proposed DB interface additions (PQDequeue/PQSignalDone/PQReEnqueue)

The key unresolved question from both PRs was **where the reconciler should run**. The [team discussion](https://github.com/llm-d-incubation/batch-gateway/pull/135#issuecomment-2738419693) favored **batch-gc** over adding leader election to processors, since batch-gc already runs as a single long-lived process.

---

## Design

Three mechanisms work together to detect orphans, recover them quickly, and guarantee data consistency:

1. **In-flight tracker with heartbeat** — processor records job ownership in Redis, refreshes every ~5m
2. **Orphan reconciler** — batch-gc periodically cross-references DB, queue, and in-flight hash to find and recover orphans
3. **Compare-and-swap (CAS) on DB status transitions** — prevents a reconciled job from being overwritten by a stale processor

### Data flow

```
Normal path:
  PQDequeue → InFlightSet(jobID, podID) → process (heartbeat every 5m) → CAS terminal write → InFlightDelete(jobID)

Crash recovery:
[...]

PR #437, opened 2026-05-15
feat: in-flight tracking, CAS on DBUpdate, processor instrumentation

lioraron:
## Summary

PR 1 of 2 for orphan job recovery ([#434](https://github.com/llm-d-incubation/batch-gateway/issues/434)).

Adds infrastructure so that crashed-pod orphan jobs can be detected and recovered by a reconciler (PR 2). This PR writes breadcrumbs but nothing reads them yet — zero behavioral risk.

### What's included

- **In-flight tracker**: New `InFlightClient` interface + Redis hash implementation (`llmd_batch:inflight`). Records `(jobID → processorID, timestamp)` on dequeue, removes on terminal transition.
- **Heartbeat**: 5-minute ticker in `runJob()` refreshes the in-flight entry so the reconciler can distinguish active jobs from orphans.
- **CAS on `DBUpdate`**: New `expectedStatus []byte` parameter across all implementations (Redis Lua script, PostgreSQL `WHERE status = $expected`). Returns `ErrConflict` sentinel on mismatch. Callers that don't need CAS pass `nil`.
- **`PQGetIDs`**: New method on `BatchPriorityQueueClient` — Lua script uses server-side `ZSCAN` + `cjson.decode` to return all queued job IDs as `map[string]bool`.
- **Processor instrumentation**: `InFlightSet` after dequeue, `InFlightDelete` on all terminal and re-enqueue paths, heartbeat goroutine in `runJob`.
- **Design doc**: `docs/design/orphan-recovery.md` covering all three mechanisms.

### New files
- `docs/design/orphan-recovery.md`
- `internal/database/redis/redis_inflight.go`
- `internal/database/redis/redis_cas_update.lua`
- `internal/database/redis/redis_pq_get_ids.lua`
- `internal/database/mock/mock_inflight_client.go`

### Key modified files
- `internal/database/api/database.go` — `InFlightClient`, `InFlightEntry`, `ErrConflict`, `PQGetIDs`, `DBUpdate` CAS param
- `internal/database/redis/redis_db.go` — CAS Lua script integration
- `internal/database/postgresql/db_core.go` — CAS via `WHERE status = $expected`
- `internal/processor/worker/worker.go` — in-flight registration, cleanup on all pre-launch paths
- `internal/processor/worker/job_runner.go` — heartbeat goroutine, `InFlightDelete` defer
- `internal/util/clientset/clientset.go` — `InFlight` field wiring

## Test plan

- [x] All existing unit tests pass (`make pre-commit` green)
- [x] Build succeeds (`go build ./...`)
- [x] Integration tests with Redis (`make test-integration`)
- [x] E2E deploy and verify processor starts cleanly with in-flight tracking enabled

Closes part of #434


lioraron:
## E2E Verification Results

Ran full E2E suite (`make dev-clean` → `make dev-deploy` → `make test-e2e`) on a clean Kind cluster with images built from this branch.

### Results

- **All 63 E2E tests pass** — no regressions from in-flight tracking, CAS, or processor instrumentation changes.
- **In-flight Redis hash is clean after all tests**: `HLEN llmd_batch:inflight` = 0. Every job that was processed had its `InFlightDelete` called successfully, confirming cleanup works on all terminal paths (normal completion, cancellation, expiration, graceful shutdown, rolling restart).
- **Processor starts cleanly** with the new `InFlight` client wired in — no crash loops, no errors on startup.

### What couldn't be verified from logs

The specific log messages (`"registered job in-flight"`, `"cleared in-flight entry"`) couldn't be verified because the pods that processed jobs were killed and replaced by the graceful shutdown E2E tests (`PodDeleteMidJob`, `RollingRestartReEnqueue`), and Kubernetes doesn't retain logs across pod replacements without a log aggregator. However, the empty Redis hash proves the full lifecycle (set → heartbeat → delete) works correctly end-to-end.

internal/processor/worker/job_runner.go
```diff
@@ -149,6 +158,11 @@ func (p *Processor) runJob(ctx context.Context, params *jobExecutionParams) {
 	params.eventWatcher = eventWatcher
 	go p.watchCancel(ctx, params)
 
+	// Start heartbeat: periodically refreshes the in-flight entry so the
+	// orphan reconciler knows this job is still being actively processed.
+	// Stops when ctx is cancelled (SIGTERM) or the job finishes.
+	go p.heartbeat(ctx, params.jobItem.ID)
```
j-mok-dev: I think the heartbeat should run on its own child context created inside `runJob()`, rather than reusing one of the existing contexts.

The existing contexts all have the wrong lifetime for this:
- `ctx` is too long-lived: it is only cancelled on SIGTERM, so heartbeat can outlive normal job completion.
- `sloCtx` / `requestAbortCtx` are too short-lived: they are for aborting execution, but `runJob()` still has terminal handling to do after they are cancelled.
- `userCancelCtx` has different semantics entirely.

The heartbeat really wants the lifetime of `runJob()` itself: start when `runJob()` starts, stop when `runJob()` returns for any reason. Something like:

```go
heartbeatCtx, heartbeatCancel := context.WithCancel(ctx)
defer heartbeatCancel()
go p.heartbeat(heartbeatCtx, params.jobItem.ID)
```

internal/database/redis/redis_db.go
```diff
@@ -150,11 +150,24 @@ func (c *DSClientRedis) dbUpdate(ctx context.Context,
...
 
+	key := getKeyForStore(indexes.ID, itemType)
 	cctx, ccancel := context.WithTimeout(ctx, c.timeout)
 	defer ccancel()
-	err = c.redisClient.HSet(cctx, getKeyForStore(indexes.ID, itemType), fields...).Err()
-	if err != nil {
-		return
+
+	if len(expectedStatus) > 0 {
+		result, casErr := redisScriptCASUpdate.Run(cctx, c.redisClient,
```
j-mok-dev: I think the Lua script is expecting a flat argv list here, but this call is passing `fields` as a single slice argument. the script wants:

`expectedStatus, "status", newStatus, ...`

but this call is sending:

`expectedStatus, []any{"status", newStatus, ...}`

So the CAS path does not seem to be invoking `redis_cas_update.lua` with the argv shape it expects. I think this needs to flatten the arguments before calling `Run`, for example, `args := append([]any{expectedStatus}, fields...)` and then `Run(..., args...)`.

internal/processor/worker/worker.go
```diff
@@ -397,6 +418,29 @@ func (p *Processor) releaseForNextPoll() {
 	p.release()
 }
 
+func (p *Processor) deleteInFlight(ctx context.Context, jobID string) {
+	if err := p.inflight.InFlightDelete(ctx, jobID); err != nil {
+		logr.FromContextOrDiscard(ctx).Error(err, "Failed to delete in-flight entry", "jobId", jobID)
+	}
+}
+
+func (p *Processor) heartbeat(ctx context.Context, jobID string) {
```
j-mok-dev, line 427: Question: is the DB status check / self-abort logic intentionally deferred to the reconciler follow-up PR? In issue #434, the PR1 plan for heartbeat seemed to include not just refreshing the in-flight timestamp, but also reading the current DB status and cancelling the job if a reconciler had already moved it. The implementation here only does the refresh, so I wanted to check whether that scope changed or if this part is still missing.

j-mok-dev:
I think it would be helpful to add a few focused tests around the new failure-sensitive paths. In particular, it would be great to cover:
- heartbeat stopping when `runJob()` returns, so the in-flight entry is not recreated after cleanup
- the Redis CAS path with a non-nil `expectedStatus`
- `PQGetIDs` behavior directly, since reconciler triage will depend on it

internal/processor/worker/job_runner.go
lioraron: Good catch. Fixed in fe6f491: heartbeat now runs on a dedicated `heartbeatCtx` derived from `ctx` with `defer heartbeatCancel()`, so it stops when `runJob()` returns. Added `TestHeartbeat_StopsOnContextCancel` to verify.

internal/database/redis/redis_db.go
lioraron: Good catch. Fixed in fe6f491: we now flatten with `args := append([]any{expectedStatus}, fields...)` and spread via `args...`. Added a CAS integration test with three cases (match, mismatch → ErrConflict, nil bypass) that exercises the Lua script end-to-end.

internal/processor/worker/worker.go
lioraron: This is deferred. The self-abort (reading DB status during heartbeat and cancelling the job if the reconciler already moved it) only becomes useful once the reconciler exists. Without a reconciler, no external actor can transition the job, so the check would be inactive code. It's tracked as part of the reconciler PR (Phase 2 in the design doc).

lioraron:
> I think it would be helpful to add a few focused tests around the new failure-sensitive paths. In particular, it would be great to cover:
> 
> * heartbeat stopping when `runJob()` returns, so the in-flight entry is not recreated after cleanup
> * the Redis CAS path with a non-nil `expectedStatus`
> * `PQGetIDs` behavior directly, since reconciler triage will depend on it

Agreed, all three covered in fe6f491:
- TestHeartbeat_StopsOnContextCancel: verifies heartbeat stops calling InFlightSet after context cancel
- CAS update (redis integration): table-driven with match/mismatch/nil-bypass cases, verifies DB state after each
- PQGetIDs (redis integration): empty queue, all enqueued returned, dequeued excluded

j-mok-dev (approved):
LGTM thanks!

PR #440, opened 2026-05-18
fix(test): avoid typed-nil failure in HTTP client integration test (#438)

### LINKED ISSUE #438: bug: TestHTTPClientIntegration concurrent subtest always fails (nil interface comparison)
## Bug

`TestHTTPClientIntegration/BasicInference/should_handle_concurrent_requests_correctly` fails deterministically on `main`. It is not flaky — it fails 5/5 runs.

**File**: `pkg/clients/inference/inference_client_integration_test.go:142-171`

## Root cause

Classic Go nil-interface bug. `Generate()` returns `(*ClientError)(nil)` on success, but the test sends the result into a `chan error`:

```go
results := make(chan error, numRequests)        // channel typed as `error` (interface)
// ...
_, inferr := client.Generate(context.Background(), req)
results <- inferr                               // (*ClientError)(nil) → non-nil error interface
```

When a concrete nil pointer (`*ClientError(nil)`) is assigned to an interface (`error`), Go wraps it in a non-nil interface value (type info present, value nil). The subsequent `inferr != nil` check is always `true`, so the test always reports failure — even though every request actually succeeded. The log output confirms this: `concurrent request X failed: <nil>`.

## Fix

Replace `chan error` with `chan *ClientError`, or convert explicitly:

```go
// Option A: typed channel
results := make(chan *ClientError, numRequests)

// Option B: explicit nil conversion
if inferr != nil {
    results <- inferr
} else {
    results <- nil
}
```

## Reproduction

```bash
go test -v -tags=integration ./pkg/clients/inference/ \
  -run 'TestHTTPClientIntegration/BasicInference/should_handle_concurrent_requests_correctly' \
  -count=5
```

Fails 5/5 on current `main` (commit $(git rev-parse --short HEAD)).

## Impact

`make test-integration` always fails. The test was likely broken since commit 8887e84 (chore: remove testify dependency, #210) which rewrote assertions from testify (which handles nil interfaces) to stdlib `!= nil` checks.

j-mok-dev:
## Why is this PR needed?

The concurrent `TestHTTPClientIntegration` subtest was always failing because it sent `*ClientError` values through a `chan error`. When `Generate()` succeeded, the typed-nil `*ClientError` was boxed into the `error` interface and no longer compared equal to `nil`, so the test reported a failure even though the request succeeded.

## What does this PR do?

- changes the concurrent integration test to use `chan *ClientError` instead of `chan error`
- keeps the test aligned with the `Generate()` signature and avoids the typed-nil interface trap

## How was this tested?

Manual / targeted verification:
- `go test -tags=integration ./pkg/clients/inference -run 'TestHTTPClientIntegration/BasicInference/should_handle_concurrent_requests_correctly'`
- `pre-commit run -a`

## Checklist

## Related Issues

Fixes #438

### DIFF
--- a/pkg/clients/inference/inference_client_integration_test.go
+++ b/pkg/clients/inference/inference_client_integration_test.go
@@ -142,7 +142,7 @@ func testHTTPClientBasicInference(t *testing.T) {
 	t.Run("should handle concurrent requests correctly", func(t *testing.T) {
 		// Verifies connection pooling and thread safety
 		const numRequests = 10
-		results := make(chan error, numRequests)
+		results := make(chan *ClientError, numRequests)
 
 		for i := 0; i < numRequests; i++ {
 			go func(id int) {


lioraron:
Thanks!
Closing as duplicate of #443 (includes the same fix plus a comment).

PR #450, opened 2026-05-22
fix(dev): use port-forward for processor observability in kind

j-mok-dev:
## Why is this PR needed?

The local dev deployment used a fixed `NodePort` for processor observability on port `30090`.
That made the deploy path vulnerable to timing-dependent port allocation conflicts, which is consistent with the flaky failure tracked in #449.

This change does not claim to prove the single root cause of #449 from CI logs alone.
Instead, it removes one fragile part of the dev-deploy and e2e path by eliminating the fixed processor `NodePort` while preserving the existing local `http://localhost:9090` contract used by tests.

## What does this PR do?

- removes the fixed processor `NodePort` from kind port mappings and local-access service creation
- starts a supervised `kubectl port-forward` for processor observability on `http://localhost:9090`
- retries the port-forward when the processor pod is restarted during restart-sensitive e2e flows
- waits for the forwarded processor endpoint to become ready before reporting deployment success
- cleans up the port-forward process in `dev-clean` and `dev-rm-cluster`
- keeps deletion of legacy `batch-gateway-processor-nodeport` services from older local environments
- adds optional debug logging for existing `NodePort` allocations
- updates the e2e README to document the new behavior

## How was this tested?

Verified locally:
- non-GIE e2e: 66 passed, 0 failed
- GIE e2e: 69 passed, 0 failed
- `HelmUpgrade` restart path passed with the supervised processor port-forward
- `ProcessorGracefulShutdown` restart paths passed with the supervised processor port-forward

## Checklist

## Related Issues

- Addresses #449


lioraron:
Thanks for investigating #449. A few concerns with this approach:

1. **The CI failure was only in the GIE variant's deploy step**, not across all variants. If fixed NodePorts were fragile, the apiserver (30080), observability (30081), Prometheus (30091), Jaeger (30086), and Grafana (30030) would all be equally affected — but they're left unchanged. This suggests the root cause is GIE-specific, not a NodePort conflict.

2. **The port-forward approach adds significant complexity** (background supervisor, PID file, stop-file signaling, retry loop, readiness polling, cleanup in two scripts) to replace a 15-line NodePort service that matches the pattern used by every other service. Also, a NodePort handles pod restarts automatically — the supervisor has to reimplement that with retry logic.

3. **The PR description acknowledges this is speculative**: "This change does not claim to prove the single root cause of #449 from CI logs alone."

I'd suggest a different approach:
- Keep the processor NodePort as-is (consistent with all other services)
- Investigate the actual error from CI run 26251379014's GIE deploy step — the GIE variant has additional install steps (CRDs, helm dependency build, InferencePool/InferenceObjective creation) that are more likely candidates
- Only change the port strategy if a port conflict is actually observed in the logs

j-mok-dev:
> Thanks for investigating #449. A few concerns with this approach:
> 
> 1. **The CI failure was only in the GIE variant's deploy step**, not across all variants. If fixed NodePorts were fragile, the apiserver (30080), observability (30081), Prometheus (30091), Jaeger (30086), and Grafana (30030) would all be equally affected — but they're left unchanged. This suggests the root cause is GIE-specific, not a NodePort conflict.
> 2. **The port-forward approach adds significant complexity** (background supervisor, PID file, stop-file signaling, retry loop, readiness polling, cleanup in two scripts) to replace a 15-line NodePort service that matches the pattern used by every other service. Also, a NodePort handles pod restarts automatically — the supervisor has to reimplement that with retry logic.
> 3. **The PR description acknowledges this is speculative**: "This change does not claim to prove the single root cause of [Integration tests failed on 2026-05-21 #449](https://github.com/llm-d-incubation/batch-gateway/issues/449) from CI logs alone."
> 
> I'd suggest a different approach:
> 
> * Keep the processor NodePort as-is (consistent with all other services)
> * Investigate the actual error from CI run 26251379014's GIE deploy step — the GIE variant has additional install steps (CRDs, helm dependency build, InferencePool/InferenceObjective creation) that are more likely candidates
> * Only change the port strategy if a port conflict is actually observed in the logs

Thanks for the review. you're right that port-forward adds unnecessary complexity.

I went back to the CI logs for [run 26251379014](https://github.com/llm-d-incubation/batch-gateway/actions/runs/26251379014) and found the exact failure:
```
The Service "batch-gateway-processor-nodeport" is invalid:
  spec.ports[0].nodePort: Invalid value: 30090: provided port is already allocated
```
So the root cause is confirmed: port 30090 collides with something the GIE EPP charts allocate. The other NodePorts (30080, 30081, etc.) happen to not overlap with anything in the GIE variant, which is why only the processor port fails.

Given your feedback, I think the simplest fix is:

- Drop the hardcoded nodePort: 30090 and let Kubernetes auto-assign it
- Read the allocated port dynamically in the script (e.g. `kubectl get svc ... -o jsonpath='{.spec.ports[0].nodePort}'`)

That keeps the NodePort pattern consistent with the other services while permanently eliminating the collision. I'll update the PR with this approach if you agree.

lioraron:
Sounds good — the auto-assign approach is the right fix. Drop the hardcoded `nodePort: 30090` and read the allocated port dynamically via `kubectl get svc`. That keeps things consistent with the existing NodePort pattern while permanently eliminating the collision.

j-mok-dev:
closing this in favor of the replacement PR.
This PR was heading in the right direction by removing the fixed processor observability port, but the host-reachability assumption did not hold reliably in local kind + Colima setups, especially across restarts and Helm upgrades.
The replacement PR supersedes this by resolving processor observability on demand in the e2e helpers.
Superseded by: #506

PR #452, opened 2026-05-25
feat: orphan reconciler for stuck batch jobs

### LINKED ISSUE #434: [orphan-recovery] feat: orphan job detection and reconciliation for pod crash recovery

lioraron:
## Summary

- Add an **orphan reconciler** that runs in `batch-gc` alongside the existing garbage collector. It periodically cross-references non-terminal jobs (DB), queued jobs (priority queue), and in-flight entries to detect orphaned batch jobs and transition them based on status and SLO expiry using CAS updates.
- Add **`NonTerminal` query** support to `BatchQuery` for both PostgreSQL (`status::jsonb->>'status' NOT IN (...)`) and Redis (new Lua script), enabling efficient server-side filtering.
- Wire the reconciler into **batch-gc** via `errgroup`, gated by `reconciler.enabled` config (defaults to `true`).
- Add **heartbeat logging** (start/refresh/stop) to the processor worker for operational visibility.

Closes #434

## Triage logic

| Current Status | SLO Expired? | Action |
|---|---|---|
| `cancelling` | — | CAS → `cancelled` |
| `validating` | yes | CAS → `expired` |
| `validating` | no | re-enqueue with original SLO |
| `in_progress` / `finalizing` | yes | CAS → `expired` |
| `in_progress` / `finalizing` | no | CAS → `failed` |

## New files
- `internal/gc/reconciler/reconciler.go` — reconciler implementation
- `internal/gc/reconciler/reconciler_test.go` — unit tests (triage, skip non-orphans, stale cleanup, CAS conflict, RunLoop)
- `internal/database/redis/redis_get_non_terminal.lua` — Redis Lua script for NonTerminal query

## Test plan
- [x] Triage: each status × SLO expired/not → correct transition
- [x] Skip: job in queue or freshly in-flight → not treated as orphan
- [x] CAS conflict: `ErrConflict` → counted as error, job not transitioned
- [x] Stale cleanup: in-flight entry for absent job → deleted; entry for non-terminal job → preserved
- [x] RunLoop: stops on context cancel
- [x] NonTerminal query: PostgreSQL mock verifies `NOT IN` clause
- [x] Config: defaults (enabled=true, 60m), custom interval, disabled skips validation, zero/negative interval → error
- [x] `make pre-commit` passes (fmt, vet, lint, tests, build, gosec)


internal/shared/openai/batch.go
```diff
@@ -58,10 +58,21 @@ func (s BatchStatus) String() string {
 	return string(s)
 }
 
-func (s BatchStatus) IsFinal() bool {
+func (s BatchStatus) IsTerminal() bool {
 	return s == BatchStatusCompleted || s == BatchStatusFailed || s == BatchStatusExpired || s == BatchStatusCancelled
```
yizhaodev: can you change to use returned statuses from `func TerminalStatuses`?

lioraron: Done — introduced a package-level `terminalStatuses` array (not slice, so it can't be appended to or resliced) as the single source of truth. `IsTerminal()` ranges over it directly (no allocation), and `TerminalStatuses()` returns a copy so callers can't mutate the canonical list.

internal/processor/worker/worker.go
```diff
@@ -424,23 +427,60 @@ func (p *Processor) deleteInFlight(ctx context.Context, jobID string) {
...
 			return
 		case <-ticker.C:
 			if err := p.inflight.InFlightSet(ctx, jobID, p.processorID); err != nil {
-				logger.Error(err, "Failed to refresh in-flight heartbeat", "jobId", jobID)
+				logger.Error(err, "Heartbeat: failed to refresh in-flight entry")
+			} else {
+				logger.V(logging.INFO).Info("Heartbeat: refreshed")
+			}
+
+			if p.checkReconcilerActed(ctx, jobID, logger) {
```
yizhaodev, line 449: should `heartbeat` call `checkReconcilerActed` before `p.inflight.InFlightSet`?

yizhaodev: If the reconciler just deleted the in-flight entry, the current order writes it back via `InFlightSet` before checking the terminal status.

lioraron: The current order (refresh first, then check) is intentional. Refreshing first signals liveness to the reconciler — it updates `LastSeen`, so if the reconciler scans during this window it sees a fresh heartbeat and skips the job. If we checked first and the reconciler had just started its cycle, we might abort prematurely before the reconciler commits its CAS write.

The reverse race (reconciler acts between our refresh and our check) is safe: the CAS write already committed the terminal status, so our check catches it on the same tick.

lioraron: Good observation — the recreated in-flight entry is harmless though. The reconciler's `cleanupStaleInflight` removes entries for jobs no longer in the non-terminal set, so it gets cleaned up on the next cycle automatically.

Keeping refresh-first is intentional: if the reconciler is mid-scan (hasn't committed its CAS write yet), the refreshed `LastSeen` tells it "this processor is still alive, skip this job." Checking first would risk aborting prematurely before the reconciler actually commits a terminal transition.

internal/gc/reconciler/reconciler.go
```diff
@@ -0,0 +1,410 @@
...
+	result := &Result{}
+
+	jobs, err := r.fetchNonTerminalJobs(ctx)
+	if err != nil {
+		logger.Error(err, "Reconciler: failed to fetch non-terminal jobs")
+		result.Errors++
+		r.notifyCycle(result)
+		return
+	}
+
```
yizhaodev, line 127: can you check `len(jobs)==0` here before next query

lioraron: Done — when `len(jobs) == 0`, the queue query (`PQGetIDs`) and triage loop are skipped. `InFlightGetAll` and `cleanupStaleInflight` still run so stale entries for already-terminal jobs are cleaned up.

cmd/batch-gc/main.go
```diff
@@ -66,19 +69,41 @@ func run() error {
...
 	if err != nil {
 		return fmt.Errorf("failed to create clients: %w", err)
 	}
 	defer func() { _ = clients.Close() }()
 
 	gc := collector.NewGarbageCollector(clients.BatchDB, clients.FileDB, clients.File, cfg.DryRun, cfg.Interval, cfg.MaxConcurrency, nil)
 
-	if err := gc.RunLoop(ctx); err != nil && ctx.Err() == nil {
-		return fmt.Errorf("garbage collector failed: %w", err)
+	if !cfg.Reconciler.Enabled {
```
yizhaodev: consider code refactoring

```
g, gCtx := errgroup.WithContext(ctx)
g.Go(func() error { return gc.RunLoop(gCtx) })

if cfg.Reconciler.Enabled {
    rec, err := reconciler.NewReconciler(...)
    if err != nil {
        return fmt.Errorf("failed to create reconciler: %w", err)
    }
    g.Go(func() error { return rec.RunLoop(gCtx) })
}

if err := g.Wait(); err != nil && ctx.Err() == nil {
    return fmt.Errorf("gc/reconciler failed: %w", err)
}
```

lioraron: Good call — adopted your suggestion. The errgroup now always runs the GC collector, and the reconciler is conditionally added when enabled. Single code path, no duplication.

yizhaodev (approved):

PR #458, opened 2026-05-27
feat: e2e dispatcher integration

### LINKED ISSUE #431: feat: Integrate llm-d-async as an alternative dispatch backend
### Problem Statement

The batch processor currently only supports synchronous HTTP dispatch for inference requests, directly coupling it to inference endpoints. This limits flexibility in queue management, retries, and integration with the broader llm-d ecosystem.

### Proposed Solution

Add support for [llm-d-async](https://github.com/llm-d-incubation/llm-d-async) as an alternative asynchronous dispatch backend. When enabled via `dispatch_mode: async`, the processor will:
- Submit requests to a Redis sorted-set via the `producer.Producer` interface
- Collect results from a Redis list, matching responses to in-flight requests
- Enforce per-request timeouts to mark individual requests as failed if no result arrives within a configurable duration

This decouples request submission from completion, enabling better queue management and retry handling at the messaging layer.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

evacchi:
## Why is this PR needed?

The batch-gateway processor currently dispatches inference requests via direct HTTP calls to model gateways.
To integrate with the llm-d-async dispatcher we need an async dispatch path that enqueues requests
to Redis sorted sets and reads results from Redis lists, using the llm-d-async wire format.

Test this with:

```
ENABLE_DISPACHER=true make dev-deploy 
make test-e2e 
```

As explained below the async interface calls from a more profound refactoring, but I suggest the team to take on this in a follow-up.

## What does this PR do?

Originally I used the `api` package directly, without realizing I should use the `producer` pkg (neither were published).

Now, if I understand correctly, the batch-gateway processor dispatches multiple requests per model concurrently. The `llm-d-async` dispatcher also processes requests concurrently (batchSize × budget). So results arrive in the result queue in arbitrary order, not in submission order. A naive SubmitRequest + GetResult per goroutine would return someone else's result.

The `AsyncInferenceClient` uses a background `resultDispatcher` goroutine to demux results from the shared result queue by request ID.

A single `resultDispatcher` per pool reads from the result queue (BRPOP) and routes each result to the correct caller via a `sync.Map` of `requestID -> chan`. Each `Generate()` call registers its channel before submitting, then selects on the channel or context cancellation. On context timeout, the waiter is unregistered to avoid leaks.

Note that because of the current assumptions in the code I had to add a separate
`p.processModelAsync(...)` alongside the existing `p.processModel(...)` in `executeJob(...)`. I refactored what I could in helper methods, but in general this should call for a more significant refactoring.

The `Async*` new set of interfaces is significantly different from the sync versions (they might be be aligned using channels).

### More details

**Async inference client interface** (`pkg/clients/inference/async_inference_client_interface.go`):
- New `AsyncInferenceClient` interface with `Submit(ctx, req)`, `GetResult(ctx)`, and `Close()`, symmetric with the sync `InferenceClient.Generate()` but non-blocking

**Async inference client implementation** (`pkg/clients/inference/async_inference_client_impl.go`):
- `asyncProducerClient` implements `AsyncInferenceClient`, backed by an `llm-d-async/producer.Producer`.
- A shared `resultDispatcher` goroutine per pool BRPOPs from the result queue and routes
  results to the correct per-job client via a `sync.Map` of `requestID -> channel`.
- Each job gets its own client (with its own internal results channel) from
  `AsyncGatewayResolver.ClientFor`, so concurrent batch jobs on the same model are isolated.
- `Close()` unregisters all pending waiters from the shared dispatcher.

**Async resolver** (`pkg/clients/inference/async_inference_client_resolver.go`):
- `AsyncGatewayResolver` is separate from `GatewayResolver` (sync). Stores shared `asyncPool`
  instances (producer + dispatcher per pool). `ClientFor` creates a fresh per-job client each call.
- Validates that no two models map to the same pool (single BRPOP reader per result queue).

**Async executor path** (`internal/processor/worker/executor.go`):
- `processModelAsync` — submit/collect pattern: sequential ZADD loop (fast), then
  `GetResult` loop collecting results as they arrive. No semaphores, no AIMD — the
  llm-d-async dispatcher controls inference concurrency via gates.
- Shared helpers extracted from `processModel`: `readRequestLine`, `buildOutputLine`,
  `writeResult`, `drainAndFinalize`, `newErrorOutputLine`, `fairnessID`.
- `executeJob` dispatches to `processModelAsync` when `p.asyncInference != nil`.

**Processor setup** (`internal/processor/worker/worker.go`):
- `initConcurrencyControls` extracted from `Run()`. In async mode, only the job-level
  worker semaphore is created — no global/endpoint semaphores or AIMD.

**Clientset** (`internal/util/clientset/clientset.go`):
- `AsyncInference *inference.AsyncGatewayResolver` field, separate from `Inference`.
- `WithAsyncInference(cfg)` sets `AsyncInference`; `Close()` closes whichever is set.

**Config & resolution** (`internal/processor/config/config.go`):
- `ResolveModelGateways` populates `ResolvedGateways.Async` with pool names from
  `model_gateways[*].inference_pool_name` when `dispatch_mode: async`.
- Global gateway rejected in async mode (per-model with pool names required).

**Helm chart** (`charts/batch-gateway/templates/processor-configmap.yaml`):
- `dispatch_mode`, `async_dispatch.result_poll_timeout`, `inference_pool_name` conditionally rendered.
- HTTP gateway fields rendered only when `url` is set, fixing a falsy-zero regression.

**E2E tests & dev tooling**:
- `test/e2e/dispatcher_test.go`: 4 tests — batch round-trip, multi-request batch,
  Redis dispatch gate, Prometheus-query dispatch gate.
- `scripts/dev-deploy-dispatcher.sh`: deploys three llm-d-async dispatcher instances
  (redis gate, endpoint-scrape gate, prometheus-query gate), configures Prometheus scrape
  targets, enables fake metrics on vllm-sim, reconfigures processor for async mode.
- `make dev-deploy-dispatcher` and `make test-e2e-dispatcher` targets.

## How was this tested?

    - `TestNewAsyncResolver`: per-model routing, labels, error on invalid URL
    - `TestAsyncInferenceClient_Generate`: enqueue/dequeue round-trip via producer, timeout, queue name derivation
    - `TestResolveModelGateways_Async`: async config resolution, sync mode unaffected
    - `TestDispatcher/BatchThroughDispatcher`: single request end-to-end via async dispatch
    - `TestDispatcher/MultiRequestBatch`: 3-request batch completion
    - `TestDispatcher/DispatchGate`: Redis gate blocks/unblocks dispatch
    - `TestDispatcher/PrometheusGate`: Prometheus-query gate blocks dispatch when vllm-sim
      reports saturation (fake metrics), unblocks when saturation clears
    - Deployed to Kind cluster via `make dev-deploy && make dev-deploy-dispatcher`
    - All 4 dispatcher e2e tests pass

## Checklist


internal/processor/config/config.go
```diff
@@ -222,18 +222,6 @@ type BucketConfig struct {
...
-// RequestQueueName returns the Redis sorted-set name for submitting async requests to the given pool.
-func RequestQueueName(poolName string) string {
-	return "llm-d-async:requests:" + poolName
-}
-
-// ResultQueueName returns the Redis list name for collecting async results from the given pool.
-func ResultQueueName(poolName string) string {
-	return "llm-d-async:results:" + poolName + ":" + asyncTenantID
-}
-
```
evacchi, line 244: aligned to https://github.com/llm-d-incubation/llm-d-async/pull/201

internal/processor/config/config.go
```diff
@@ -423,15 +411,13 @@ func (c *ProcessorConfig) validateSyncDispatchConfig() error {
 
 func (c *ProcessorConfig) validateAsyncDispatchConfig() error {
 	if c.AsyncDispatchConfig.ResultPollTimeout <= 0 {
-		return fmt.Errorf("async.result_poll_timeout must be > 0")
-	}
-	if err := c.validateGateways(); err != nil {
-		return err
+		return fmt.Errorf("async_dispatch.result_poll_timeout must be > 0")
```
evacchi, line 419: yaml key is actually `async_dispatch`

docker/Dockerfile.processor
```diff
@@ -6,6 +6,7 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+COPY .build/llm-d-async/ /llm-d-async/
```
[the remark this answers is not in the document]
evacchi: this will be removed before merging

go.mod
```diff
@@ -67,12 +72,17 @@ require (
...
+	go.yaml.in/yaml/v2 v2.4.3 // indirect
+	golang.org/x/net v0.53.0 // indirect
+	golang.org/x/sys v0.43.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 )
+
+replace (
+	github.com/llm-d-incubation/llm-d-async/api => ../llm-d-async/api
+	github.com/llm-d-incubation/llm-d-async/producer => ../llm-d-async/producer
+)
```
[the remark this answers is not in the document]
evacchi: this will be removed before merging

pkg/clients/inference/async_inference_client.go
```diff
@@ -0,0 +1,101 @@
...
+	c.logger.V(logging.TRACE).Info("Submitted async request", "requestID", req.RequestID)
+
+	result, err := c.producer.GetResult(ctx)
+	if err != nil {
+		return nil, &ClientError{
+			Category: httpclient.ErrCategoryServer,
+			Message:  fmt.Sprintf("poll async result: %v", err),
+			RawError: err,
+		}
+	}
```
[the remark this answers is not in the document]
evacchi: ok I was misunderstanding how this worked originally, this should be fixed now by using these "`resultDispatcher`s"

pkg/clients/inference/async_inference_client.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// Close stops the background result dispatcher goroutine.
+func (c *AsyncInferenceClient) Close() error {
+	c.dispatcher.stop()
+	return nil
+}
+
+func (c *AsyncInferenceClient) Generate(ctx context.Context, req *GenerateRequest) (*GenerateResponse, *ClientError) {
```
yizhaodev: It seems AsyncInferenceClient.Generate() will block until result received?
So the number of in-flight async requests is capped by the processor's worker concurrency limit, we won't really benfit from async mode?

evacchi: yes good point, but it looks to me like this is sort of the contract implied by the interface; maybe we should not use this interface here?

yizhaodev: what if batch processor simply dump all requests to llm-d-async without waiting result and use a another goroutine collect results

this way is to let batch disptacher to handle capacity / sorting etc

would that have better throughput?

go.mod
```diff
@@ -67,12 +72,17 @@ require (
...
-	go.yaml.in/yaml/v2 v2.4.2 // indirect
-	golang.org/x/net v0.52.0 // indirect
-	golang.org/x/sys v0.42.0 // indirect
+	go.yaml.in/yaml/v2 v2.4.3 // indirect
+	golang.org/x/net v0.53.0 // indirect
+	golang.org/x/sys v0.43.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 )
+
+replace (
```
yizhaodev: the replace is to get feature not in v0.7.0-rc.4?
should we import by git commit id rather than version to resolve it?

pkg/clients/inference/async_inference_client.go
evacchi: right, the big issue here is the signature for `Generate`:

      Generate(ctx context.Context, req *GenerateRequest) (*GenerateResponse, *ClientError) 

this pretty much implies a synchronous, blocking API. I'll try to sketch something with an async API

go.mod
evacchi: unfortunately that was not enough (explained in https://github.com/llm-d-incubation/llm-d-async/pull/214). Now that that PR is merged we are pointing to the correct version `v0.7.0-rc.6`

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,180 @@
...
+
+		if val, ok := d.waiters.LoadAndDelete(result.ID); ok {
+			ch := val.(chan<- *GenerateResponse)
+			ch <- &GenerateResponse{
+				RequestID: result.ID,
+				Response:  []byte(result.Payload),
+			}
+		} else {
+			d.logger.V(logging.TRACE).Info("Dropped result with no waiter", "resultID", result.ID)
+		}
```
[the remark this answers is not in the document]
evacchi: this is fine for now, let's do this in a follow-up

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,157 @@ dispatch:
...
-	undispatched := entries[dispatchedCount:]
+	// ── Phase 2: Collect ───────────────────────────────────────────────────
+	var modelErr error
+	collected := 0
+
+	for collected < len(pending) {
+		resp, err := asyncClient.GetResult(requestAbortCtx)
+		if err != nil {
+			break
+		}
```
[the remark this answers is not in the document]
evacchi: requestAbortCtx is derived from sloCtx (line 153). SLO expiry cancels sloCtx which cancels requestAbortCtx which makes GetResult return ctx.Err(). The comment is a false positive. No fix needed.

test/e2e/dispatcher_test.go
```diff
@@ -0,0 +1,320 @@
...
+	p := newDispatcherProducer(t, rdb, promPool)
+
+	// Saturate the sim — gate should close
+	// (query: 1 - clamp_max(vllm:num_requests_waiting / 5, 1) → 0 when waiting ≥ 5)
+	setSimWaitingRequests(t, 10)
+	defer setSimWaitingRequests(t, 0)
+	t.Log("Sim saturated (waiting-requests=10, gate should close)")
+
+	// Give Prometheus time to scrape the new metric value
+	time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
evacchi: this is acceptable for e2e tests where the scrape interval is known (5s), and the sleep (10s) gives 2x margin -- it's ok for now, we can address this in a follow-up

pkg/clients/inference/async_inference_client.go
evacchi: should be addressed now

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,157 @@ dispatch:
...
-	// requestAbortFn() was called by another goroutine before the sloCtx deadline propagated.
-	undispatched := entries[dispatchedCount:]
+	// ── Phase 2: Collect ───────────────────────────────────────────────────
+	var modelErr error
+	collected := 0
+
+	for collected < len(pending) {
+		resp, err := asyncClient.GetResult(requestAbortCtx)
+		if err != nil {
+			break
```
lioraron, line 665: When this `break` fires (SLO expiry, user cancel, pod shutdown), the remaining `len(pending) - collected` requests that were successfully submitted are silently lost — they don't appear in either the output file or the error file. `drainAndFinalize` below only drains `entries[submitCount:]` (entries never submitted), not the submitted-but-uncollected entries still in `pending`.

This violates the batch API invariant that `output_lines + error_lines == total_requests`. For example: 100 entries, 80 submitted, 50 collected before SLO expires → 50 written + 20 drained = 70 lines instead of 100.

After the collect loop breaks, the remaining `pending` entries should be drained as error lines (expired/cancelled) before calling `drainAndFinalize`.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,157 @@ dispatch:
...
+		if requestAbortCtx.Err() != nil {
+			break
+		}
+
+		req, batchReqID, parseErr, readErr := readRequestLine(inputFile, entry, logger)
+		if readErr != nil {
+			return readErr
+		}
+		if parseErr != nil {
+			lineBytes, _ := json.Marshal(parseErr)
```
lioraron: Both `json.Marshal` and `writers.write` errors are discarded here (and in the equivalent submit-error path at line 618). In the sync path (`processModel`), write failures are captured via `errOnce.Do(func() { modelErr = ... })` and treated as fatal. The async path silently swallows them, which can cause the error file to become incomplete with no indication of why.

Same applies to the submit-error path below:
```go
lineBytes, _ := json.Marshal(out)
_ = writers.write(lineBytes, true)
```

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,191 @@
...
+func (d *resultDispatcher) run(ctx context.Context) {
+	for {
+		pollCtx, pollCancel := context.WithTimeout(ctx, time.Second)
+		result, err := d.producer.GetResult(pollCtx)
+		pollCancel()
+		if err != nil {
+			if ctx.Err() != nil {
+				return
+			}
+			continue
```
lioraron, line 82: All `GetResult` errors — including Redis connection failures, auth errors, cluster failovers, and deserialization errors — are silently swallowed here. If Redis goes down, this spin-loops with 1-second timeouts generating thousands of errors/second with zero visibility. Operators will see requests timing out in `processModelAsync` with no root cause.

At minimum, non-timeout errors should be logged:
```go
if !errors.Is(err, context.DeadlineExceeded) {
    d.logger.Error(err, "Failed to read from result queue")
}
```

evacchi:
```
make dev-deploy
DISPATCHER_SOURCE=/Users/evacchi/Devel/github.com/llm-d-incubation/llm-d-async \
    make dev-deploy-dispatcher
make test-e2e-dispatcher
```

override using PR at https://github.com/llm-d-incubation/llm-d-async/pull/219

jaeger trace:

Screenshot 2026-06-03 at 16 59 47

evacchi:
squashed+rebased

internal/processor/config/config.go
```diff
@@ -423,15 +411,13 @@ func (c *ProcessorConfig) validateSyncDispatchConfig() error {
 
 func (c *ProcessorConfig) validateAsyncDispatchConfig() error {
 	if c.AsyncDispatchConfig.ResultPollTimeout <= 0 {
-		return fmt.Errorf("async.result_poll_timeout must be > 0")
-	}
-	if err := c.validateGateways(); err != nil {
```
acardace, line 440: nit: don't you think all of this is still gateways validation? Maybe all the logic should be in that single function `validateGateways` and maybe we could even drop the switch case.

internal/processor/config/config.go
```diff
@@ -423,15 +411,13 @@ func (c *ProcessorConfig) validateSyncDispatchConfig() error {
...
-		return err
+		return fmt.Errorf("async_dispatch.result_poll_timeout must be > 0")
 	}
 	if c.GlobalInferenceGateway != nil {
-		if c.GlobalInferenceGateway.InferencePoolName == "" {
-			return fmt.Errorf("global_inference_gateway.inference_pool_name must be set when dispatch_mode is %q", DispatchModeAsync)
-		}
+		return fmt.Errorf("global_inference_gateway is not supported with dispatch_mode %q; use model_gateways with inference_pool_name", DispatchModeAsync)
+	}
+	if len(c.ModelGateways) == 0 {
```
acardace, line 424: Wasn't introduced by this PR but I think having `InferencePoolName` as the only useful field for async buried in ModelGateways where all the other fields are effectively ignored is not the best API design, maybe we should have a separate config for Async?

evacchi:
rebased+pushed

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,195 @@
...
+	return &resultDispatcher{
+		producer: p,
+		logger:   logger,
+	}
+}
+
+func (d *resultDispatcher) ensureStarted() {
+	d.once.Do(func() {
+		ctx, cancel := context.WithCancel(context.Background())
+		d.cancel = cancel
```
lioraron, line 63: The `run` goroutine has two issues:

**1. No shutdown synchronization.** `Close()` calls `cancel()` but doesn't wait for the goroutine to exit. In-flight results between `GetResult` returning and `LoadAndDelete` can be lost. Add a `sync.WaitGroup` (or done channel) so `Close()` blocks until `run` finishes.

**2. No panic recovery.** If `run` panics (e.g. type assertion on line 75, nil result from producer), every `GetResult` caller sharing this pool hangs forever with no diagnostic output. Add a `defer recover()` that at minimum logs the panic.

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,195 @@
...
+				d.logger.Error(err, "Failed to read from result queue")
+			}
+			continue
+		}
+
+		if val, ok := d.waiters.LoadAndDelete(result.ID); ok {
+			ch := val.(chan<- *GenerateResponse)
+			resp := &GenerateResponse{
+				RequestID: result.ID,
+				Response:  []byte(result.Payload),
```
lioraron, line 93: Dropping a completed inference result (that cost GPU time) is data loss, but it's logged at TRACE level which is typically disabled in production. This should be at least Warning/Error level so operators notice the buffer is undersized.

Also — the buffer size of 100 (line 28) is a hardcoded magic number. Per project conventions, larger channel buffers need justification. Consider making it configurable or at least extracting it as a named constant with a comment explaining the sizing.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,182 @@ dispatch:
...
+		submitCount++
 	}
 
-	wg.Wait()
+	logger.V(logging.INFO).Info("Submit phase complete", "submitted", len(pending), "total", submitCount)
+
+	// ── Phase 2: Collect ───────────────────────────────────────────────────
+	var modelErr error
+
+	for len(pending) > 0 {
```
lioraron, line 658: When `GetResult` returns an error, the loop breaks silently — no logging and no assignment to `modelErr`. A Redis connection failure mid-collection would cause all remaining requests to drain as "result not collected before deadline" with no log explaining the actual cause.

Distinguish context cancellation (expected) from unexpected errors:

```go
if err != nil {
    if requestAbortCtx.Err() == nil {
        logger.Error(err, "Failed to collect async result", "pendingCount", len(pending))
        modelErr = fmt.Errorf("async result collection failed: %w", err)
    }
    break
}
```

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,182 @@ dispatch:
...
-	// Priority: SLO expiry > user cancel > system error > pod shutdown.
-	// Use sloCtx.Err() rather than requestAbortCtx.Err(): requestAbortCtx may report Canceled if
-	// requestAbortFn() was called by another goroutine before the sloCtx deadline propagated.
-	undispatched := entries[dispatchedCount:]
+		pr, ok := pending[resp.RequestID]
+		if !ok {
+			logger.V(logging.TRACE).Info("Ignoring result for unknown request", "requestID", resp.RequestID)
+			continue
+		}
+
```
lioraron, line 673: The collect loop continues after a `writeResult` failure, and each subsequent failure overwrites `modelErr` (losing the original error). The sync path (`processModel`) returns immediately on the first write error via `errOnce.Do`.

Break on the first write error to stay consistent with the sync path:

```go
if err := writeResult(...); err != nil {
    modelErr = err
    break
}
```

pkg/clients/inference/async_inference_client_resolver.go
```diff
@@ -0,0 +1,112 @@
...
+// Immutable after construction — safe for concurrent reads.
+type AsyncGatewayResolver struct {
+	pools   map[string]*asyncPool // model → pool
+	closers []io.Closer
+}
+
+// ClientFor creates a fresh per-job async client for the given model.
+// Returns nil if no matching pool exists.
+func (r *AsyncGatewayResolver) ClientFor(modelID string) AsyncInferenceClient {
+	pool, ok := r.pools[modelID]
```
lioraron, line 58: `closers` is built with `rdb` (the Redis client) first, then dispatchers and producers are appended. When `Close()` iterates this list, the Redis connection is torn down while dispatchers may still be polling it, causing spurious errors.

Reverse the order — close dispatchers first (to stop the polling goroutines), then producers, then the Redis client:

```go
closers = append(closers, d, p)
// ...
closers = append(closers, rdb) // add Redis client last
```

internal/util/clientset/clientset.go
```diff
@@ -41,14 +41,15 @@ import (
...
-	BatchDB   dbapi.BatchDBClient
-	FileDB    dbapi.FileDBClient
-	Queue     dbapi.BatchPriorityQueueClient
-	Event     dbapi.BatchEventChannelClient
-	Status    dbapi.BatchStatusClient
-	InFlight  dbapi.InFlightClient
-	Inference *inference.GatewayResolver
+	File           fsapi.BatchFilesClient
+	BatchDB        dbapi.BatchDBClient
+	FileDB         dbapi.FileDBClient
```
lioraron, line 47: Unlike `WithPerModelInference` (which defensively copies the map), this function stores `&cfg` without copying `cfg.Models`. Since `Models` is a map (reference type), mutations by the caller after this call would be visible inside the stored config.

Copy the map to match the established pattern:

```go
func WithAsyncInference(cfg inference.AsyncClientConfig) Option {
    copied := cfg
    copied.Models = maps.Clone(cfg.Models)
    return func(c *clientsetConfig) { c.asyncInference = &copied }
}
```

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,195 @@
...
+	asyncapi "github.com/llm-d-incubation/llm-d-async/api"
+	"github.com/llm-d-incubation/llm-d-async/producer"
+	"github.com/llm-d/llm-d-batch-gateway/internal/util/logging"
+	httpclient "github.com/llm-d/llm-d-batch-gateway/pkg/clients/http"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+)
+
+var _ AsyncInferenceClient = (*asyncProducerClient)(nil)
+
```
lioraron, line 36: The fallback deadline `now.Add(5 * time.Minute)` is hardcoded and undocumented. If inference takes longer than 5 minutes (large prompts, cold model loading), the dispatcher will consider the request expired. Make this configurable via `AsyncClientConfig` or at least extract it as a named constant.

evacchi:
addressed comments, rebased, tested e2e locally

internal/processor/worker/executor.go
```diff
@@ -279,18 +279,34 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
...
-				inputFile,
-				plansDir, safeModelID, modelID,
-				writers,
-				progress,
-				passThroughHeaders,
-				tenantID,
-			)
+			var err error
+			if p.asyncInference != nil {
+				err = p.processModelAsync(
```
acardace, line 284: Why not introducing an interface for this? The two functions share the exact same set of args, you can then pass the concrete impl to `executeJob`

internal/processor/worker/executor.go
```diff
@@ -835,9 +957,60 @@ func (p *Processor) executeOneRequest(
...
+) error {
+	if sloCtx.Err() == nil && userCancelCtx.Err() != nil {
+		out.Response = nil
+		out.Error = &outputError{
+			Code:    string(batch_types.ErrCodeBatchCancelled),
+			Message: "This request was cancelled while in progress.",
+		}
+		progress.record(progressCtx, false)
+	} else {
+		progress.record(progressCtx, out.isSuccess())
```
acardace: you can just drop the else branch and `progress.record(progressCtx, false)` in the if branch and just do `progress.record(progressCtx, out.isSuccess())` unconditionally.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,187 @@ dispatch:
...
+
+	asyncClient := p.asyncInference.ClientFor(modelID)
+	if asyncClient == nil {
+		logger.V(logging.INFO).Info("No async client for model, draining as model_not_found")
+		p.drainUnprocessedRequests(
+			requestAbortCtx, inputFile, entries, writers, progress,
+			inference.ErrCodeModelNotFound)
+		return nil
+	}
+	defer func() { _ = asyncClient.Close() }()
```
acardace: I'd log it anyway in case of failure.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,187 @@ dispatch:
...
+	type pendingRequest struct {
+		batchReqID string
+		customID   string
+	}
+
+	pending := make(map[string]*pendingRequest)
+	var submitCount int
+
+	for _, entry := range entries {
+		if requestAbortCtx.Err() != nil {
```
acardace, line 593: Shouldn't we at least log and possibly have an otel trace here? Otherwise this function goes completely silent if the it's immediately aborted.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,187 @@ dispatch:
...
-	// Drain undispatched entries to the error file based on the termination reason, and return the
-	// appropriate sentinel so executeJob can route without re-examining context state.
-	// Priority: SLO expiry > user cancel > system error > pod shutdown.
-	// Use sloCtx.Err() rather than requestAbortCtx.Err(): requestAbortCtx may report Canceled if
-	// requestAbortFn() was called by another goroutine before the sloCtx deadline propagated.
-	undispatched := entries[dispatchedCount:]
+	for len(pending) > 0 {
+		resp, err := asyncClient.GetResult(requestAbortCtx)
+		if err != nil {
+			if requestAbortCtx.Err() == nil {
```
acardace, line 661: This can be a single if statement.

internal/processor/worker/executor.go
```diff
@@ -506,63 +522,187 @@ dispatch:
...
+	var modelErr error
 
-	// Drain undispatched entries to the error file based on the termination reason, and return the
-	// appropriate sentinel so executeJob can route without re-examining context state.
-	// Priority: SLO expiry > user cancel > system error > pod shutdown.
-	// Use sloCtx.Err() rather than requestAbortCtx.Err(): requestAbortCtx may report Canceled if
-	// requestAbortFn() was called by another goroutine before the sloCtx deadline propagated.
-	undispatched := entries[dispatchedCount:]
+	for len(pending) > 0 {
+		resp, err := asyncClient.GetResult(requestAbortCtx)
```
acardace, line 659: Doesn't this make it synchronous? Shouldn't we have a different worker coming from a different pool waiting for the results and producing the output files?

My worry is that this basically is the same as the sync counterpart as it holds up the worker which keeps the semaphore token until this is done, effectively this counters the benefits of using async IIUC because `runJob` doesn't decrease the semaphore until `executeJob` is done.

acardace (commented):
Partial review as I think there's a fundamental flow with how async is integrated IIUC.

internal/processor/config/config.go
evacchi: agreed, but this should be addressed separately, let's create an issue

internal/processor/worker/executor.go
evacchi: I'll log the err returned by Close

internal/processor/worker/executor.go
evacchi: there's no other places here where otel in executor.go is being used, I'll add a log, but we might want to address otel separately in a systematic way.

internal/processor/config/config.go
evacchi: I folded the logic for sync/async+validateDispatchMode into validateGateways, I hope that's what you meant.

evacchi:
llm-d-incubation/llm-d-async v0.7.1 was tagged yesterday, maybe we can drop some/all of the source deps

evacchi:
squashed+rebased

internal/processor/config/config.go
evacchi: created https://github.com/llm-d/llm-d-batch-gateway/issues/526

Makefile
```diff
@@ -337,6 +337,10 @@ dev-deploy:
 dev-deploy-gie:
 	@ENABLE_GIE=true bash scripts/dev-deploy.sh
 
+## dev-deploy-dispatcher: Deploy the llm-d-async dispatcher into the dev Kind cluster (requires dev-deploy first)
+dev-deploy-dispatcher:
```
j-mok-dev: it currently breaks e2e tests because make test-e2e try to run dispatcher test when it's not deployed. 

`make test-e2e` runs `./...` which includes `TestDispatcher` and `TestDispatcherOTelTraces` and these fail in a sync-only deployment because there's no dispatcher.

Suggested fix:
1. Add a skip guard at the top of both test functions:
```go
if os.Getenv("TEST_DISPATCHER") == "" {
    t.Skip("skipping: requires dev-deploy-dispatcher (set TEST_DISPATCHER=1)")
}
```
test/e2e/dispatcher_test.go: TestDispatcher
test/e2e/dispatcher_otel_test.go: TestDispatcherOTelTraces

2. Set the env var in `make teset-e2e-dispatcher`
```
test-e2e-dispatcher:
	@echo "Running dispatcher E2E tests..."
	cd test/e2e && TEST_DISPATCHER=1 $(GO) test -v -count=1 -run TestDispatcher ./...
```

then make tese-e2e stays green witout the dispatcher, and `make test-e2e-dispatcher` still runs the full async path.

internal/processor/worker/executor.go
```diff
@@ -507,67 +523,196 @@ dispatch:
...
+		if err != nil && requestAbortCtx.Err() == nil {
+			logger.Error(err, "Failed to collect async result", "pendingCount", len(pending))
+			modelErr = fmt.Errorf("async result collection failed: %w", err)
+			break
+		}
+
+		pr, ok := pending[resp.RequestID]
```
lioraron, line 668: **Bug (release-blocking): Nil pointer dereference when context is cancelled during async collect**

When `GetResult` returns `(nil, ctx.Err())` because `requestAbortCtx` was cancelled (SLO expiry, user cancel, SIGTERM), the condition `err != nil && requestAbortCtx.Err() == nil` is **false**, so execution falls through to this line where `resp` is nil — causing a panic.

This will crash the processor on every non-happy-path termination during async collection.

**Fix:** break on any error, regardless of context state:
```go
if err != nil {
    if requestAbortCtx.Err() == nil {
        logger.Error(err, "Failed to collect async result", "pendingCount", len(pending))
        modelErr = fmt.Errorf("async result collection failed: %w", err)
    }
    break
}
```

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,220 @@
...
+		if val, ok := d.waiters.LoadAndDelete(result.ID); ok {
+			ch := val.(chan<- *GenerateResponse)
+			resp := &GenerateResponse{
+				RequestID: result.ID,
+				Response:  []byte(result.Payload),
+			}
+			select {
+			case ch <- resp:
+			default:
+				d.logger.Info("Result channel full, dropping result", "resultID", result.ID)
```
lioraron: **Silent result dropping when channel buffer is full**

The `defaultResultBufferSize` is 100. If a batch submits >100 requests for a single model and the async dispatcher returns results faster than the collect loop reads them, results are permanently dropped. The dropped result leaves a request stuck in `pending` until it's drained as an expiry error.

**User impact:** Successfully completed inference requests reported as expired errors — data loss with no recovery.

The log is at `Info` level, so this won't trigger alerts.

**Suggestions:**
1. Size the channel to match the number of submitted requests (known at submit time), or
2. At minimum, escalate to `Error` level and record a metric so operators are alerted.

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,220 @@
...
+		d.wg.Add(1)
+		go d.run(ctx)
+	})
+}
+
+func (d *resultDispatcher) run(ctx context.Context) {
+	defer d.wg.Done()
+	defer func() {
+		if r := recover(); r != nil {
+			d.logger.Error(fmt.Errorf("%v", r), "Panic in result dispatcher")
```
lioraron: **Panic recovery kills dispatcher goroutine permanently — no restart**

If the dispatcher panics (e.g., from the bare type assertion on line 91 or a library bug), `recover()` logs the panic but the goroutine exits. Because `ensureStarted()` uses `sync.Once`, it can never re-launch. All subsequent `GetResult` calls for this pool block until context timeout.

**Impact:** A single panic kills async result collection for all batches using that model/pool. Every batch times out and reports all requests as expired.

**Suggestions:**
1. Remove the recover and let the panic crash the process (standard Go approach — pod restart via K8s), or
2. Replace `sync.Once` with a mechanism that allows restart on failure.

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,220 @@
...
+	if d.cancel != nil {
+		d.cancel()
+		done := make(chan struct{})
+		go func() {
+			d.wg.Wait()
+			close(done)
+		}()
+		select {
+		case <-done:
+		case <-time.After(2 * time.Second):
```
lioraron, line 125: **Close() silently returns nil after 2s timeout, potentially leaking goroutine**

If the dispatcher is blocked in `producer.GetResult()` (BRPOP) and context cancellation doesn't interrupt the Redis call within 2 seconds, Close returns `nil` but the goroutine keeps running. It may then operate on a closed Redis connection (the resolver closes Redis after Close returns).

Returning `nil` prevents the caller from knowing resources are leaked.

**Fix:** Return an error when the timeout fires:
```go
case <-time.After(2 * time.Second):
    return fmt.Errorf("result dispatcher did not shut down within 2s")
```

pkg/clients/inference/async_inference_client_impl.go
```diff
@@ -0,0 +1,220 @@
...
+				return
+			}
+			if !errors.Is(err, context.DeadlineExceeded) {
+				d.logger.Error(err, "Failed to read from result queue")
+			}
+			continue
+		}
+
+		if val, ok := d.waiters.LoadAndDelete(result.ID); ok {
+			ch := val.(chan<- *GenerateResponse)
```
lioraron: **Bare type assertion violates project convention (CLAUDE.md: "Always use comma-ok form")**

If this assertion fails (e.g., due to a concurrent misuse or future code change), it panics inside the dispatcher goroutine. Combined with the `sync.Once`-guarded `recover()` above, a failed assertion here permanently kills the dispatcher with no restart.

**Fix:**
```go
ch, ok := val.(chan<- *GenerateResponse)
if !ok {
    d.logger.Error(fmt.Errorf("unexpected type %T in waiters map", val), "Type assertion failed")
    continue
}
```

internal/processor/worker/executor.go
```diff
@@ -507,67 +523,196 @@ dispatch:
...
-				kind := "output"
-				if isError {
-					kind = "error"
-				}
-				errOnce.Do(func() { modelErr = fmt.Errorf("write %s line at offset %d: %w", kind, entry.Offset, writeErr) })
+		if errors.Is(sloCtx.Err(), context.DeadlineExceeded) {
+			break
+		}
+
+		headers := maps.Clone(passThroughHeaders)
```
lioraron, line 620: **Missing unit tests for `processModelAsync`**

This function has no dedicated unit tests — it's only covered indirectly by e2e tests. The nil-pointer dereference bug (see other comment) would have been caught by a unit test exercising context cancellation during the collect phase.

Please add unit tests covering at least:
- Happy path (submit N, collect N)
- Context cancellation during collect
- Partial collection (some results arrive, then timeout)

scripts/dev-deploy-dispatcher.sh
```diff
@@ -0,0 +1,280 @@
+#!/bin/bash
```
lioraron, line 1: **Align dispatcher e2e framework with the existing GIE pattern**

The GIE integration follows a pattern that makes CI coverage straightforward:
1. `ENABLE_GIE=true` flag integrated into `dev-deploy.sh`
2. Tests run via the standard `make test-e2e` target
3. CI adds a matrix entry with `enable_gie: "true"`

The dispatcher currently uses separate `dev-deploy-dispatcher` / `test-e2e-dispatcher` targets, which means CI would need a separate job or workflow step to cover it. This also makes the dispatcher the only e2e variant that doesn't run in CI.

Please consider:
1. Integrate dispatcher deployment into `dev-deploy.sh` behind an `ENABLE_DISPATCHER=true` flag (like GIE)
2. Run dispatcher tests via the standard `make test-e2e` (they already live in `test/e2e/`)
3. This enables CI coverage by adding a matrix entry with `enable_dispatcher: "true"`

test/e2e/dispatcher/helm-values.yaml
```diff
@@ -0,0 +1,36 @@
...
+    samplerArg: "1.0"
+    redisTracing: false
+  metrics:
+    enabled: true
+    port: 9090
+    secure: false
+  redis:
+    enabled: true
+    url: "redis://redis-master.default.svc.cluster.local:6379"
+    resultQueueName: "llm-d-async:results:sim-pool"
```
yizhaodev: from code
```
		producer.RedisSortedSetConfig{
			RequestQueueName: asyncQueuePrefix + "requests:" + poolName,
			ResultQueueName:  asyncQueuePrefix + "results:" + poolName,
		},
```
it seem `sim-pool` and `sim-pool-gate` has its own result queue?

why this conf used shared result queue?

test/e2e/dispatcher/helm-values-prometheus.yaml
```diff
@@ -0,0 +1,32 @@
+ap:
+  imagePullPolicy: Never
+  messageQueueImpl: "redis-sortedset"
```
yizhaodev, line 3: question: do need to test gcp impl?

evacchi: no I don't think so, we don't have that sort of integration here

test/e2e/dispatcher/helm-values.yaml
evacchi: wow thanks for noticing!

Makefile
evacchi: should work now

pkg/clients/inference/async_inference_client_impl.go
evacchi: done

pkg/clients/inference/async_inference_client_impl.go
evacchi: done

internal/processor/worker/executor.go
evacchi: done

scripts/dev-deploy-dispatcher.sh
evacchi: Must have missed this one here. I'll make sure we use ENABLE_DISPATCHER and migrate/unify the logic for TEST_DISPATCHER=1 that was added upon @j-mok-dev's suggestion.

evacchi: @madhugoutham FYI ^

the workflow will become:

```
ENABLE_DISPATCHER=true make dev-deploy
ENABLE_DISPATCHER=true make test-e2e
```

test/e2e/dispatcher/helm-values.yaml
evacchi: fixed

scripts/dev-deploy-dispatcher.sh
```diff
@@ -0,0 +1,284 @@
...
+    set -euo pipefail
+    SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+    REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+    source "${SCRIPT_DIR}/dev-common.sh"
+fi
+
+# ── Configuration ────────────────────────────────────────────────────────────
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-batch-gateway-dev}"
+DISPATCHER_RELEASE="${DISPATCHER_RELEASE:-dispatcher}"
+DISPATCHER_VERSION="${DISPATCHER_VERSION:-v0.7.2}"
```
yizhaodev, line 14: @evacchi 
It looks like we need to pin to llm-d-async 0.7.2 for docker images

our midstream repo is: https://github.com/opendatahub-io/llm-d-async

I wonder how should we handle the upstream -> midstream sync?
- shall we sync only release 0.7.2 to midstream main and build image from midstream main?
- shall still sync all to midstream main, and create 0.7.2 branch in midstream and build from it?

how did you handle things like this in the past?
@lioraron @vishbhat

scripts/dev-deploy-dispatcher.sh
evacchi: also added to CI in 5b1f249

test/e2e/dispatcher/helm-values-prometheus.yaml
```diff
@@ -0,0 +1,32 @@
...
+    secure: false
+  redis:
+    enabled: true
+    url: "redis://redis-master.default.svc.cluster.local:6379"
+    pollIntervalMs: 500
+    batchSize: 10
+    queuesConfig:
+      - queue_name: "llm-d-async:requests:sim-pool-prom"
+        result_queue_name: "llm-d-async:results:sim-pool-prom"
+        request_path_url: "/v1/completions"
```
yizhaodev, line 25: question for `request_path_url: "/v1/completions"`
does this mean each queue will only support one endpoint?
do we have to define multiple queue for same pool to support various request path?

scripts/dev-deploy-dispatcher.sh
evacchi: we don't strictly _need_ to pin -- I just tested this morning against `main` by building the image locally:

```
DISPATCHER_SOURCE=/path/to/llm-d-async ENABLE_DISPATCHER=true make dev-deploy
```

and it works fine. 

The version is defaulting to 0.7.2 because it's a published image, so CI can just fetch it (otherwise we would need to clone and build from source each time).

If we want to build from source, we can do it with the flag above and ensure we clone the repo; we only need to make sure that upstream did not make a breaking change

test/e2e/dispatcher/helm-values-prometheus.yaml
evacchi: mm yes I think you are right. this is how the llm-d-async config was defined so far. we can probably submit changes there to change this behavior

evacchi: actually, no. 

the per-request endpoint (ir.PublicRequest.ReqEndpoint()) overrides the queue's RequestPathURL if set. So the queue config is just the default, each request can specify its own endpoint.

charts/batch-gateway/templates/processor-configmap.yaml
```diff
@@ -41,12 +41,23 @@ data:
...
+    {{- end }}
+
     {{- with .Values.processor.config.globalInferenceGateway }}
     global_inference_gateway:
       url: {{ .url | quote }}
       {{- if .inferenceObjective }}
       inference_objective: {{ .inferenceObjective | quote }}
       {{- end }}
+      {{- if .inferencePoolName }}
+      inference_pool_name: {{ .inferencePoolName | quote }}
```
yizhaodev, line 60: do you need inference_pool_name for global_inference_gateway?

because i saw this in below
```
return fmt.Errorf("global_inference_gateway is not supported with dispatch_mode %q; use model_gateways with inference_pool_name", DispatchModeAsync)
```

j-mok-dev (approved):

scripts/dev-deploy-dispatcher.sh
```diff
@@ -0,0 +1,284 @@
+#!/bin/bash
```
lioraron, line 1: `.dispatcher-port-forward.pid` and `.dispatcher-sim-port-forward.pid` are created by this script but not gitignored. They show up as untracked files after `ENABLE_DISPATCHER=true make dev-deploy`.

Add them to `.gitignore`.

test/e2e/dispatcher_test.go
```diff
@@ -0,0 +1,408 @@
...
+	if err != nil {
+		t.Fatalf("Failed to create producer for pool %s: %v", poolName, err)
+	}
+	t.Cleanup(func() { _ = p.Close() })
+	return p
+}
+
+func TestDispatcher(t *testing.T) {
+	if os.Getenv("ENABLE_DISPATCHER") == "" {
+		t.Skip("skipping: requires dispatcher (set ENABLE_DISPATCHER=true)")
```
lioraron: The GIE tests auto-detect the deployment via `detectGIEDeployed(t)` (checks for EPP deployments in the cluster) — no env var needed for `make test-e2e`. The dispatcher tests should follow the same pattern: detect whether dispatcher pods are running rather than requiring `ENABLE_DISPATCHER` to be passed explicitly.

This way `make test-e2e` works against whatever is deployed, consistent with how GIE tests behave.

test/e2e/dispatcher/helm-values-prometheus.yaml
```diff
@@ -0,0 +1,32 @@
...
+        igw_base_url: "http://vllm-sim.default.svc.cluster.local:8000"
+        gate_type: "prometheus-query"
```
yizhaodev, line 27: I am working on batch-gateway-operator to deploy llm-d-async
what gate_type and gate_params should be default values? or keep it as empty?

charts/batch-gateway/templates/processor-configmap.yaml
evacchi: no, global_inference_gateway doesn't need inference_pool_name. Async mode explicitly rejects global_inference_gateway, validation fails with that error message.

inference_pool_name is only needed on model_gateways entries, which should be where the template renders it. The error message is telling users to switch from global_inference_gateway to model_gateways when using async mode 

we can do amendments to the config format in another PR anyway

test/e2e/dispatcher/helm-values-prometheus.yaml
evacchi: I think prometheus-budget might be the safest, it should cascade automatically to different strategies in case the queries are missing data https://github.com/llm-d-incubation/llm-d-async#per-queue-dispatch-gates

yizhaodev (approved):

lioraron (approved):
Thanks @evacchi!

PR #464, opened 2026-06-02
test(e2e): restore RetryOn429 and use /admin/config for failure injection

### LINKED ISSUE #448: Stabilize RetryOn429 coverage
## Summary

This issue tracks the remaining follow-up for `RetryOn429` coverage after the flaky E2E path was removed.

The behaviour itself is still important:

- if the downstream returns a transient `429`
- the system should retry
- and the request should eventually succeed once the downstream recovers

The problem is that the previous E2E scenario could not exercise that path deterministically because it depended on probabilistic 429 injection from the simulator. As a result, the test was inherently flaky.

## Current status

- #445  captured the original flaky failure report for `TestE2E/FlowControl/HeaderAndRetry/RetryOn429`
- #447  disabled the flaky E2E scenario and kept the retry-specific behaviour covered in deterministic lower-level tests

At this point, coverage is not missing, but deterministic E2E coverage for this scenario is still unavailable.

## What changed in #447

`#447` made the following decision:

- skip the flaky `RetryOn429` E2E scenario for now
- keep retry-path coverage in deterministic lower-level tests instead
- add/keep coverage for both:
  - retry behaviour in the client layer
  - propagation of `HadCapacityRetry=true` in the worker layer

This keeps the important behaviour covered without depending on probability in CI.

## Why the old E2E was flaky

The old test effectively assumed that at least one request in the scenario would hit a transient `429` and then later succeed.

That assumption does not always hold when failure injection is probabilistic:
- some runs may see no `429` at all
- some runs may exercise a different retry pattern
- the test may pass or fail depending on chance rather than behaviour

Because of that, the retry-path assertion does not belong in E2E until the simulator can force a deterministic failure sequence.

## Desired end state

We should restore this as a real E2E scenario once deterministic failure injection exists upstream.

What we need from the simulator is support for a deterministic sequence such as:
- fail the first `N` attempts, then return `200`
- or return a fixed sequence like `429, 429, 200`

With that support, we can verify the full end-to-end contract without relying on chance.

## TODO
- [x] Keep the current `RetryOn429` E2E skipped until that support exists
- [x] [Open an upstream issue in `llm-d-inference-sim` for deterministic 429/failure-sequence injection](https://github.com/llm-d/llm-d-inference-sim/issues/506)
- [ ] Track the issue above
- [ ] Restore the `RetryOn429` E2E scenario once deterministic simulator behaviour is available
- [ ] Verify the restored test with repeated CI/local runs before re-enabling it permanently

## Exit criteria

This issue can be closed when all of the following are true:

- deterministic failure injection is available upstream
- the `RetryOn429` E2E test is restored
- the restored E2E passes reliably in repeated runs

## Related

- #445 original flaky test report
- #447 disabled the flaky E2E path and preserved retry coverage in deterministic lower-level tests
- Upstream issue: [llm-d-inference-sim#506](https://github.com/llm-d/llm-d-inference-sim/issues/506)

j-mok-dev:
## Why is this PR needed?

The AIMD and RetryOn429 E2E tests relied on `kubectl patch deployment` + rollout wait to change simulator failure rates, taking ~90s per toggle and introducing timing-based flakiness. The `RetryOn429` test was skipped entirely due to lack of deterministic failure injection. Additionally, metric assertions used absolute counter values, causing false failures on repeated runs against a long-lived processor.

## What does this PR do?

- **Use `/admin/config` for runtime failure injection** (requires `llm-d-inference-sim` >= v0.9.1): replaces `kubectl patch` + rollout with a direct HTTP POST that takes effect immediately (~1s, no pod restart).
- **Restore `RetryOn429` test**: deterministically sets 100% failure, waits for `model_inflight_requests > 0` to prove dispatch happened, then flips to 0% and asserts all requests complete with AIMD decrease signals.
- **Delta-based metric assertions**: all cumulative counters (`aimd_decreases_total`, `request_errors_by_model_total`) and the healthy endpoint's AIMD gauge are now validated as deltas from a pre-test snapshot, making tests safe to rerun on the same processor.
- **Consolidate curl pods**: merged `batch-gateway-epp-metrics-curl` and sim-admin curl pod into a single `batch-gateway-e2e-curl` with `sleep infinity`.
- **Propagate `TEST_SIM_SERVICE_*` env vars** in the Makefile so custom simulator service names from `dev-deploy` reach the test binary.

## How was this tested?

- [x] Unit tests pass (`make pre-commit`)
- [x] Full `make test-e2e` passes

## Checklist

## Related Issues

Closes #448

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -377,8 +377,6 @@ benchmark-local-teardown:
 test-e2e:
 	@echo "Running E2E tests..."
 	@OUT=$$(mktemp); \
-	export TEST_SIM_SERVICE_429=$${VLLM_SIM_429_NAME:-vllm-sim-429}; \
-	export TEST_SIM_SERVICE_AIMD=$${VLLM_SIM_AIMD_NAME:-vllm-sim-aimd}; \
 	cd test/e2e && $(GO) test -v -count=1 $(if $(TEST_RUN),-run $(TEST_RUN)) ./... 2>&1 | tee $$OUT; \
 	TEST_EXIT=$${PIPESTATUS[0]}; \
 	PASS_COUNT=$$(grep -- '--- PASS:' $$OUT 2>/dev/null | wc -l | tr -d ' '); \
--- a/Makefile
+++ b/Makefile
@@ -377,6 +377,8 @@ benchmark-local-teardown:
 test-e2e:
 	@echo "Running E2E tests..."
 	@OUT=$$(mktemp); \
+	export TEST_SIM_SERVICE_429=$${VLLM_SIM_429_NAME:-vllm-sim-429}; \
+	export TEST_SIM_SERVICE_AIMD=$${VLLM_SIM_AIMD_NAME:-vllm-sim-aimd}; \
 	cd test/e2e && $(GO) test -v -count=1 $(if $(TEST_RUN),-run $(TEST_RUN)) ./... 2>&1 | tee $$OUT; \
 	TEST_EXIT=$${PIPESTATUS[0]}; \
 	PASS_COUNT=$$(grep -- '--- PASS:' $$OUT 2>/dev/null | wc -l | tr -d ' '); \
--- a/test/e2e/flow_control_test.go
+++ b/test/e2e/flow_control_test.go
@@ -49,9 +49,7 @@ func testFlowControl(t *testing.T) {
 		}
 		t.Run("InferenceObjectiveHeader", doTestInferenceObjectiveHeader)
 		t.Run("SLOHeader", doTestSLOHeader)
-		t.Run("RetryOn429", func(t *testing.T) {
-			t.Skip("#445: disabled until llm-d-inference-sim supports deterministic 429 injection; retry coverage lives in lower-level deterministic tests")
-		})
+		t.Run("RetryOn429", doTestRetryOn429)
 		t.Run("RetryExhaustion", doTestRetryExhaustion)
 	})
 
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -345,6 +345,8 @@ dev-rm-cluster:
 test-e2e:
 	@echo "Running E2E tests..."
 	@OUT=$$(mktemp); \
+	export TEST_SIM_SERVICE_429=$${VLLM_SIM_429_NAME:-vllm-sim-429}; \
```
lioraron: These are the first env vars explicitly exported in the `test-e2e` target — all other `TEST_*` vars rely on `getEnvOrDefault` defaults in `e2e_test.go` and don't need Makefile propagation.

The defaults already match `dev-common.sh` (`vllm-sim-429`, `vllm-sim-aimd`), so these exports are only needed for the rare case where someone overrides `VLLM_SIM_429_NAME` during deploy and then runs tests separately. Consider dropping these two lines and relying on the defaults, consistent with how every other `TEST_*` var works.

j-mok-dev: Sorry for the delayed follow-up, and thanks for flagging this.

I removed those two Makefile exports and now rely on the existing defaults in `e2e_test.go`, consistent with the other `TEST_*` variables.

I’ll keep the deterministic control changes in this PR and handle #496 separately in a follow-up PR.

test/e2e/helpers_test.go
```diff
@@ -602,6 +604,135 @@ func validateBatchResults(t *testing.T, batch *openai.Batch, result batchResults
...
+// ── Simulator admin helpers ──────────────────────────────────────────────
+
+// setSimAdminConfig sends a POST to the simulator's /admin/config endpoint
+// via a curl pod running in the cluster. It is used to dynamically change
+// failure injection at runtime without restarting the simulator deployment.
+func setSimAdminConfig(t *testing.T, simService string, body string) {
+	t.Helper()
+
+	ensureE2ECurlPod(t)
+
```
lioraron, line 692: `setSimAdminConfig` uses `t.Fatalf` on failure, which is correct when called from test bodies but problematic in `t.Cleanup` functions — `t.Fatalf` in a cleanup halts that cleanup and prevents subsequent cleanup functions from running (LIFO chain is broken).

Both `doTestRetryOn429` (flow_control_test.go:149) and `doTestAIMDRecovery` (aimd_test.go:275) call `setSimAdminConfig` inside `t.Cleanup`. If the simulator's `/admin/config` endpoint is temporarily unreachable during cleanup, the curl pod deletion cleanup never runs, and the simulator is left in the wrong failure-rate state — poisoning subsequent test runs.

The old `doTestAIMDRecovery` code deliberately used `t.Errorf` in cleanup and printed `MANUAL RESTORE REQUIRED` for exactly this reason.

Consider either:
- Adding a `trySetSimAdminConfig` variant that returns an error (for use in cleanup), or
- Inlining the POST with `t.Errorf` in cleanup contexts

test/e2e/flow_control_test.go
```diff
@@ -583,22 +611,33 @@ func waitForRetryExhaustion(t *testing.T, batchID string, timeout time.Duration)
...
 	pattern := regexp.MustCompile(fmt.Sprintf(`request_errors_by_model_total\{model=%q\}\s+(\d+)`, model))
 	match := pattern.FindStringSubmatch(metrics)
 	if match == nil {
-		t.Logf("request_errors_by_model_total{model=%q} not found in metrics (expected: retries succeeded transparently)", model)
-		return
+		return 0
 	}
-	if match[1] != "0" {
-		t.Errorf("expected request_errors_by_model_total{model=%q} = 0 (retries should succeed), got %s", model, match[1])
+	val, _ := strconv.Atoi(match[1])
```
lioraron: Minor: `strconv.Atoi` error is discarded. The `\d+` regex makes this safe in practice, but the existing metric parsers in this file (`parseCounterByEndpoint`, `parseGaugeByEndpoint`) check parse errors. For consistency:

```go
val, err := strconv.Atoi(match[1])
if err != nil {
    t.Fatalf("failed to parse request_errors_by_model_total value %q: %v", match[1], err)
}
```

Same pattern appears in `waitForModelInflight` (helpers_test.go:709).

test/e2e/helpers_test.go
j-mok-dev: Thanks. I think the remaining cleanup callbacks still run in current Go, but I agree the cleanup path should be non-fatal here. I updated the restore path used from cleanup to report errors without `t.Fatalf`.

test/e2e/flow_control_test.go
j-mok-dev: thanks! Added explicit parse error handling in both places.

lioraron (approved):

PR #471, opened 2026-06-05
fix(ci): harden GitHub workflows for must-fix findings

j-mok-dev:
## Why is this PR needed?

The must-fix security report flagged multiple GitHub Actions workflow issues in this repository, including unpinned action references, unsafe action caching, a shellcheck finding, and secret-scanner false positives in PostgreSQL connection string examples.

Most of those findings can be fixed directly in-tree. Two remaining findings are `pull_request_target` warnings on workflows that need write access for fork PRs; those are intentionally kept and documented separately for risk acceptance because changing the trigger would break the current behaviour.

## What does this PR do?

- pins third-party and reusable GitHub Actions workflow references by commit SHA
- disables GitHub Actions caches where the report flagged cache-poisoning risk
- fixes the shellcheck `SC2064` finding in `scripts/dev-deploy.sh`
- replaces example PostgreSQL credentials in comments with placeholders to clear secret-scanner false positives
- adds inline security rationale to the two `pull_request_target` workflows that remain intentionally unchanged

## How was this tested?

- [x] Manual review of must-fix findings against the updated workflow diff

## Checklist

## Related Issues

- Related to RHOAIENG-63961

### DIFF
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -28,7 +28,7 @@ jobs:
         uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v4
+        uses: docker/setup-buildx-action@d7f5e7f509e45cec5c76c4d5afdd7de93d0b3df5 # v4
 
       - name: Log in to GitHub Container Registry
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4
--- a/.github/workflows/ci-release.yaml
+++ b/.github/workflows/ci-release.yaml
@@ -25,13 +25,13 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v4
+        uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
 
       - name: Log in to GitHub Container Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -67,7 +67,7 @@ jobs:
           echo "tag=${TAG}" >> $GITHUB_OUTPUT
 
       - name: Build and push Docker images
-        uses: docker/bake-action@v7
+        uses: docker/bake-action@6614cfa25eff9a0b2b2697efb0b6159e7680d584 # v7
         with:
           files: docker-bake.hcl
           targets: default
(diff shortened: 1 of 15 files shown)


lioraron:
Pinning the `llm-d/llm-d-infra` reusable workflows to a commit SHA (`@6c2d2e60...`) means every infra repo change requires updating the SHA in all downstream consumers. Since you control both repos, the security benefit over a versioned tag (e.g. `@v1`) is minimal — a compromised maintainer in your own org could retarget the SHA pin just as easily. Consider using a version tag on the infra repo instead to avoid the coordination overhead.

.github/workflows/ci-release.yaml
```diff
@@ -25,13 +25,13 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v4
+        uses: docker/setup-qemu-action@06116385d9baf250c9f4dcb4858b16962ea869c3 # v4
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
```
lioraron: The must-fix report flags `docker/setup-buildx-action@v4` as an unpinned action reference (finding #5). `setup-qemu-action`, `login-action`, and `bake-action` in this same file were all pinned — this one was missed.

j-mok-dev:
> Pinning the `llm-d/llm-d-infra` reusable workflows to a commit SHA (`@6c2d2e60...`) means every infra repo change requires updating the SHA in all downstream consumers. Since you control both repos, the security benefit over a versioned tag (e.g. `@v1`) is minimal — a compromised maintainer in your own org could retarget the SHA pin just as easily. Consider using a version tag on the infra repo instead to avoid the coordination overhead.

Good point. The infra repo doesn't have version tags yet. `@main` was the only ref available, so SHA-pinning was the only way to satisfy the scanner. Once the infra team introduces versioned tags, we'll switch to those.

.github/workflows/ci-release.yaml
j-mok-dev: Thanks. good catch. fixed in the latest push.

lioraron:
Suggestion: the CI hardening (SHA pinning, cache poisoning mitigations, pull_request_target documentation) is being discussed as an org-wide initiative rather than a per-repo fix — no other llm-d repo currently has this hardening, and applying it only here would make batch-gateway an outlier.

However, this PR also includes 3 non-workflow fixes that are legitimate and have no consistency concerns:

1. **False-positive Postgres secret in docs** — bracketed credentials in `charts/batch-gateway/values.yaml` (`postgresql://[user]:[pass]@[host]/[db]`)
2. **False-positive Postgres secret in docs** — same fix in `internal/database/postgresql/db_postgresql.go`
3. **Shellcheck SC2064** — single-quote the `trap` expression in `scripts/dev-deploy.sh`

Could we split these 3 fixes into a smaller PR that can be merged independently, while the CI hardening discussion is resolved at the org level?

j-mok-dev:
@lioraron sounds great. working on it.

j-mok-dev:
Closing this PR per review feedback.

I split the non-workflow fixes into smaller standalone PRs:
- values.yaml false-positive placeholder update
- db_postgresql.go false-positive placeholder update
- dev-deploy.sh shellcheck fix

The CI hardening changes will be handled separately once we align on the org-wide approach across llm-d repos.

PR #479, opened 2026-06-09
ci: enforce conventional PR titles and auto-label

vishbhat:
## Why is this PR needed?

Release notes are grouped by PR labels (see `.github/release.yml`), but the auto-label workflow currently only adds `ai-assisted`. Category labels such as `feature`, `bug`, and `documentation` must be applied manually, which is easy to forget and can place PRs in the wrong release-note section.

This PR automates applying a single category label when a PR is opened, using title conventions the project already follows (e.g. `feat:`, `fix:`, `docs:`, `ci:`).

## What does this PR do?

Extends the existing `auto-label-pr` workflow to detect and apply **one** release-category label alongside `ai-assisted` on PR open.
- Adds `.github/scripts/detect-pr-label.js` with detection logic:
  - **Title (primary):** conventional commit prefixes (`feat:` → `feature`, `fix:` → `bug`, etc.), breaking-change markers (`!:`), and keyword fallback for non-conventional titles
  - **Changed files (fallback):** only when the title has no match and **all** changed files fit one category (e.g. docs-only → `documentation`, workflow-only → `release-note-none`)
- Updates `.github/workflows/auto-label-pr.yml` to load the script, detect the label, and apply it via the GitHub API (with paginated file listing for the fallback)
- Adds `.github/scripts/detect-pr-label.test.js` — 62 unit tests runnable with Node's built-in test runner (`node --test detect-pr-label.test.js`, no npm dependencies)
- Adds `.github/scripts/README.md` documenting detection rules and how to run tests

If no category can be determined, only `ai-assisted` is applied.

## How was this tested?

## Checklist

### DIFF
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -1,0 +1,52 @@
+# PR Label Detection Scripts
+
+## Files
+
+- `detect-pr-label.js` - Main logic for detecting PR labels from titles
+- `detect-pr-label.test.js` - Node.js test suite
+
+## Running Tests
+
+Requires Node.js 18+ (built-in test runner, no dependencies):
+
+```bash
+node --test detect-pr-label.test.js
+```
+
+## Label Detection Rules
+
+The script detects labels with the following priority:
+
+1. **Breaking changes** (highest priority)
+   - `!: description`
+   - `type!: description`
+   - `type(scope)!: description`
+   - `breaking: description`
+
+2. **Conventional commit prefixes**
+   - `feat:` → `feature`
+   - `fix:` → `bug`
+   - `docs:` → `documentation`
+   - `deps:` → `dependencies`
+   - `enh:` / `improve:` → `enhancement`
+   - `test:` / `ci:` / `chore:` → `release-note-none`
+
+3. **Keyword fallback** (for non-conventional titles)
+   - Searches for keywords: `fix`, `bug`, `docs`, `feat`, etc.
+
+4. **File-path fallback** (when title has no match; all changed files must match)
+   - `docs/` or `*.md` → `documentation`
+   - `go.mod`, `go.sum`, `vendor/` → `dependencies`
+   - `.github/workflows/`, `test/`, `e2e/`, `*_test.go` → `release-note-none`
+
+## Examples
+
+```javascript
+detectFromTitle('feat: add feature')          // → 'feature'
+detectFromTitle('fix(api): resolve bug')      // → 'bug'
+detectFromTitle('feat!: breaking change')     // → 'breaking-change'
+detectFromTitle('Update documentation')       // → 'documentation'
+detectFromTitle('Random title')               // → null
+detectFromFiles(['docs/guide.md'])            // → 'documentation'
+detectFromFiles(['docs/a.md', 'internal/x.go']) // → null
+```
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -29,7 +29,7 @@ Breaking changes use `!` before `:`: `feat!: drop old API`, `fix(scope)!: breaki
 | `deps` | `dependencies` |
 | `ci`, `chore`, `refactor`, `test`, `style`, `perf`, `revert`, `build` | `release-note-none` |
 
-Labels align with categories in [`.github/release.yml`](../release.yml). Every PR also gets `ai-assisted`.
+Labels align with categories in [`.github/release.yml`](../release.yml).
 
 ## Files
 
--- a/.github/scripts/README.md
+++ b/.github/scripts/README.md
@@ -1,52 +1,49 @@
-# PR Label Detection Scripts
+# PR Title Validation and Label Mapping
+
+## Overview
+
+PR titles are validated and labeled by [`.github/workflows/auto-label-pr.yml`](../workflows/auto-label-pr.yml):
+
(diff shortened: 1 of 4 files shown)


.github/workflows/auto-label-pr.yml
```diff
@@ -11,13 +11,37 @@ jobs:
   add-label:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout
```
j-mok-dev: We need `pull_request_target` here so the workflow can still add labels on forked PRs. But once we introduce `actions/checkout`, the security rationale for keeping `pull_request_target` gets much weaker, since checkout is exactly what makes this trigger risky.

If there's a way to keep the label detection logic without using checkout (for example, by inlining it in `github-script`), I think that would be a safer direction.

j-mok-dev:
this workflow only runs on `pull_request_target` `opened`, but the detected label depends on mutable inputs (PR title and changed files). If the title changes later or additional commits change the file mix, the category label can become stale and release-note grouping will be wrong.

Would it make sense to also trigger on `edited`, `synchronize`, and `reopened`, and update/replace the existing category label instead of only adding it once?

.github/scripts/detect-pr-label.js
```diff
@@ -0,0 +1,74 @@
...
+  ['enhancement', /\b(enh(?:ancement)?|improve(?:ment)?)\b/],
+  ['release-note-none', /\b(tests?|refactor|ci|chore|style|perf|revert)\b/],
+];
+
+function detectFromTitle(title) {
+  const lower = title.toLowerCase();
+
+  // Priority 1: Breaking change indicators (! marker or breaking keyword in prefix)
+  // Handles: "!: description", "type!: description", "type(scope)!: description"
+  if (lower.includes('!:') || lower.match(/^(\w+)(?:\([^)]*\))?!:/)) {
```
lioraron: `lower.includes('!:')` matches `!:` **anywhere** in the title, not just at the conventional-commit prefix position. A title like `fix: handle error when status is !: done` would be mislabeled as `breaking-change` instead of `bug`.

The second part of the condition (the regex) is correctly anchored to the start. This first part should be too:

```suggestion
  if (lower.startsWith('!:') || lower.match(/^(\w+)(?:\([^)]*\))?!:/)) {
```

yizhaodev:
can we use labeler github action?
https://github.com/actions/labeler

.github/workflows/auto-label-pr.yml
```diff
@@ -1,23 +1,96 @@
...
+            chore
+            refactor
+            test
+            perf
+            style
+            revert
+            build
+
+      - name: Apply PR labels
         uses: actions/github-script@v9
```
lioraron: `actions/github-script@v9` is not SHA-pinned, but the new `amannn/action-semantic-pull-request` above is. Pin for consistency:

```suggestion
        uses: actions/github-script@373c709c69115d41ff229c7e5df9f8788daa9553 # v9
```

vishbhat: Addressed. Thanks!

j-mok-dev (approved):

.github/workflows/auto-label-pr.yml
```diff
@@ -1,23 +1,96 @@
...
+                await github.rest.issues.removeLabel({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: context.issue.number,
+                  name: label,
+                });
+              }
+            }
+
+            const labelsToAdd = ['ai-assisted'];
```
yizhaodev: we still auto adding `ai-assisted` ?

vishbhat: Thats the existing behaviour. There was no plan to remove that. 
@lioraron Please confirm.

lioraron: @vishbhat i think we can remove `ai-assisted` - can you remove the associated mechanism in this PR?

vishbhat: Removed. Thanks!

.github/scripts/README.md
```diff
@@ -0,0 +1,49 @@
+# PR Title Validation and Label Mapping
```
lioraron: The README title and content describe only the PR title validation/labeling mechanism, but the directory is `.github/scripts/` which may host other scripts in the future. Consider restructuring so the top-level heading is generic (e.g. `# CI Scripts`) and the current content becomes a section (e.g. `## PR Title Validation and Label Mapping`).

.github/workflows/auto-label-pr.yml
```diff
@@ -1,23 +1,90 @@
-name: Add default labels to PRs
+name: PR title check and labels
```
lioraron: The previous name ("Add default labels to PRs") better described the workflow's purpose — automatic label management. "PR title check and labels" leaks the implementation mechanism (title validation) into the name. Consider something like `Auto-label PRs` or `PR labels` — the title validation is an internal step, not the workflow's purpose from the caller's perspective.

lioraron (approved):

.github/scripts/README.md
vishbhat: Updated

.github/workflows/auto-label-pr.yml
vishbhat: Updated

Issue #293, opened 2026-04-01
[Bug]: Startup recovery can re-enqueue a job that is already in the queue, causing duplicate processing


j-mok-dev:
## Issue description

When a processor container shuts down gracefully (SIGTERM), in-flight jobs are re-enqueued to the Redis priority queue via `handleJobError` so another worker can pick them up. The local workdir is **not cleaned up** on this path because the job is expected to be retried.

If Kubernetes restarts a new container **in the same pod** (emptyDir survives), startup recovery scans the workdir, finds the stale job directory, and re-enqueues the job again — without checking whether the job is already in the queue.

This creates **two queue entries for the same job**. The problem is compounded by the fact that the two enqueue paths produce different JSON members:

- **Shutdown re-enqueue** (`handleJobError` → `enqueueOne`): uses the original dequeued `task`, which includes the `Data` field (e.g., `{"id":"...","slo":"...","data":"..."}`).
- **Recovery re-enqueue** (`buildRecoveryTask` → `enqueueOne`): constructs a new `BatchJobPriority` with only `ID` and `SLO` (e.g., `{"id":"...","slo":"..."}`).

Redis `ZAddNX` deduplicates by the full member byte string, so these two entries coexist as distinct members even though they represent the same job.

After dequeue, `IsJobRunnable()` accepts both `validating` and `in_progress`, so two workers can launch the same job concurrently — causing workdir/file contention and duplicate inference execution.

## Trigger conditions

All three must be true:

1. Processor receives SIGTERM during job execution (graceful shutdown).
2. Kubernetes restarts a new **container** in the same pod (not a pod replacement — emptyDir must survive).
3. The re-enqueued item has not yet been dequeued by another processor before startup recovery runs.

Note: Container crashes (OOM, panic) do **not** trigger this because the re-enqueue code in `handleJobError` does not execute. Pod-level restarts (eviction, node failure) destroy emptyDir, so startup recovery finds nothing.

## Root cause

1. **Startup recovery does not check whether the job is already in the queue** before calling `enqueueOne()`.
2. **Queue member format differs between enqueue paths**, so `ZAddNX` cannot deduplicate them.

## Suggested solution

- **Option A (member normalization)**: Make all enqueue paths produce identical member JSON for the same job (e.g., always omit `Data` from the member, or use only the job ID as the member and store payload separately). This lets `ZAddNX` block the duplicate.
[...]

PR #484, opened 2026-06-09
Fix PQDelete removing all members with the same SLO score

### LINKED ISSUE #294: [Bug]: PQDelete removes all queue entries with the same SLO score, not just the target batch
## Issue description

`PQDelete` uses `ZRemRangeByScore(score, score)` to remove a batch from the Redis priority queue during cancellation. The score is the SLO deadline as `UnixMicro()`. This deletes **all** queue members that share the same score, not just the target batch.

If two or more batches have the same SLO microsecond value — which can happen when batches are created close together with the same `completion_window`, or when the same SLO is reconstructed from a stored tag — cancelling one batch silently removes the other(s) from the queue.

The removed batches remain in `validating` status in the DB with no queue entry, so they are never picked up by a processor. Startup recovery cannot detect them either, because recovery only scans filesystem workdirs (which don't exist yet for queued jobs).

## How it happens

1. Two batches are created with the same `completion_window` at nearly the same time, producing the same `UnixMicro` SLO score.
2. Both are enqueued via `PQEnqueue` → `ZAddNX`. They coexist because their member bytes differ (different IDs).
3. User cancels batch A → `CancelBatch` calls `PQDelete` with batch A's SLO score.
4. `ZRemRangeByScore(score, score)` removes **every** member with that score — including batch B.
5. Batch B is now stuck: DB says `validating`, but no queue entry exists and no workdir exists for recovery to find.

## Root cause

`PQDelete` identifies queue entries by score alone (`ZRemRangeByScore`), not by member content. The `item.ID` parameter is used only for logging and validation, not for the actual Redis deletion.

## Existing test evidence

The Redis integration test (`redis_ds_client_test.go`) already enqueues 5 items with identical SLO values and successfully dequeues all 5, confirming that same-score members coexist in the sorted set. This same scenario is what makes `PQDelete` dangerous.

## Suggested solution

Switch from `ZRemRangeByScore` to `ZRem` with the exact member bytes. This requires the caller to reconstruct the identical member that was originally enqueued.

This is blocked on issue #293 (member normalization): currently the member is the full `json.Marshal(BatchJobPriority)` including the `Data` field, which the cancel path does not have. Once #293 normalizes the member to `{id, slo}` only, `PQDelete` can reconstruct the exact member and use `ZRem` for precise single-item deletion.

**Interim mitigation**: The collision probability is low (same microsecond SLO), but not zero. Document this as a known limitation until member normalization lands.

## Affected code

- `internal/database/redis/redis_queue.go` — `PQDelete` uses `ZRemRangeByScore`
- `internal/apiserver/batch/batch_handler.go` — `CancelBatch` calls `PQDelete` with reconstructed `{ID, SLO}`
- `internal/database/api/database.go` — `PQDelete` interface contract says "Specify the ID and SLO values for deleting" but the implementation ignores ID

apollofps:
## Why is this PR needed?

`PQDelete` uses `ZRemRangeByScore(score, score)` to remove a batch from the Redis priority queue. This deletes **all** members sharing that score, not just the target batch. When two batches have the same SLO microsecond value (same `completion_window` created close together), cancelling one silently removes the other — leaving it stuck in `validating` status with no queue entry and no recovery path.

### Note on #293 dependency

Issue #294 mentions being blocked on #293 (member normalization) because the caller can't reconstruct the exact member bytes for `ZRem`. This PR sidesteps that entirely by using an atomic Lua script that fetches members by score, decodes each to check the ID, and `ZREM`s only the match. No member reconstruction needed — the script works with the existing `BatchJobPriority` JSON format unchanged.

## What does this PR do?

- Adds `redis_pq_delete.lua`: an atomic Lua script that narrows by score via `ZRANGEBYSCORE`, then matches by job ID via `cjson.decode`, and removes exactly one member with `ZREM`. Same pattern as the existing `redis_pq_get_ids.lua`.
- Updates `PQDelete` in `redis_queue.go` to use the Lua script instead of `ZRemRangeByScore`.
- Adds a regression test: enqueue 3 items with identical SLO, delete the middle one, verify the other 2 survive via `PQGetIDs`.

## How was this tested?

All queue tests pass against real Redis (not just miniredis). Full project test suite passes with zero failures.

## Checklist

## Related Issues

Fixes #294

### DIFF
--- a/internal/database/redis/redis_ds_client.go
+++ b/internal/database/redis/redis_ds_client.go
@@ -96,6 +96,10 @@ var (
 	updateCASLua         string
 	redisScriptUpdateCAS = goredis.NewScript(updateCASLua)
 
+	//go:embed redis_pq_delete.lua
+	pqDeleteLua         string
+	redisScriptPQDelete = goredis.NewScript(pqDeleteLua)
+
 	//go:embed redis_pq_get_ids.lua
 	pqGetIDsLua         string
 	redisScriptPQGetIDs = goredis.NewScript(pqGetIDsLua)
--- a/internal/database/redis/redis_ds_client_test.go
+++ b/internal/database/redis/redis_ds_client_test.go
@@ -1451,6 +1451,57 @@ func TestRedisDSClient(t *testing.T) {
 		}
 	})
 
+	t.Run("Queue PQDelete removes only target item with same SLO", func(t *testing.T) {
+		if minirds != nil {
+			t.Skip("Miniredis model")
+		}
+		baseClient, _, _, exchClient := setupRedisDSClients(t, redisUrl, redisCaCert)
+		t.Cleanup(func() {
+			_ = baseClient.Close()
+		})
+
+		slo := time.Now().Add(time.Hour)
+		items := make([]*db_api.BatchJobPriority, 3)
+		for i := 0; i < 3; i++ {
+			items[i] = &db_api.BatchJobPriority{
+				ID:   uuid.New().String(),
+				SLO:  slo,
+				TTL:  1000,
+				Data: []byte(fmt.Sprintf("same-slo-%d", i)),
+			}
+			if err := exchClient.PQEnqueue(context.Background(), items[i]); err != nil {
+				t.Fatalf("Failed to enqueue item %d: %v", i, err)
+			}
+		}
+
+		// Delete only the middle item.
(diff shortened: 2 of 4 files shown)


lioraron:
Closing alongside #294 (won't fix). The SLO microsecond collision probability is near-zero, and member normalization from #293 would fix this as a side effect by enabling `ZRem` with exact member bytes instead of `ZRemRangeByScore`.

PR #485, opened 2026-06-10
fix: add tls MinVersion for security findings address

j-mok-dev:
## Why is this PR needed?

The security review flagged `internal/util/tls.GetTlsConfig()` for relying on the Go runtime default TLS minimum version. While current Go defaults already enforce TLS 1.2+, this PR makes that policy explicit in shared TLS config.

## What does this PR do?

- sets `MinVersion: tls.VersionTLS12` in `GetTlsConfig()`
- adds unit coverage to verify the TLS minimum version across client, server, and insecure paths

## How was this tested?

- [x] `go test ./internal/util/tls`
- [x] `make pre-commit` 
- [] Manual testing performed

## Checklist

- [X] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)
- [X] Code follows project [contributing guidelines](CONTRIBUTING.md)
- [X] CI checks pass (`make ci`)
- [X] E2E tests pass (`make test-e2e`)

## Related Issues

Related to RHOAIENG-63961

### DIFF
--- a/internal/util/tls/tls.go
+++ b/internal/util/tls/tls.go
@@ -47,6 +47,7 @@ const (
 
 func GetTlsConfig(loadType LoadType, insecure bool, certFile string, keyFile string, caCertFile string) (*tls.Config, error) {
 	var tlsConf tls.Config
+	tlsConf.MinVersion = tls.VersionTLS12
 	if certFile != "" {
 		certificate, err := tls.LoadX509KeyPair(certFile, keyFile)
 		if err != nil {
--- a/internal/util/tls/tls_test.go
+++ b/internal/util/tls/tls_test.go
@@ -1,0 +1,43 @@
+package tls
+
+import (
+	"crypto/tls"
+	"testing"
+)
+
+func TestGetTlsConfig(t *testing.T) {
+	tests := []struct {
+		name     string
+		loadType LoadType
+		insecure bool
+	}{
+		{
+			name:     "client defaults to TLS 1.2 minimum",
+			loadType: LOAD_TYPE_CLIENT,
+		},
+		{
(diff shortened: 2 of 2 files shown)


internal/util/tls/tls.go
```diff
@@ -47,6 +47,7 @@ const (
 
 func GetTlsConfig(loadType LoadType, insecure bool, certFile string, keyFile string, caCertFile string) (*tls.Config, error) {
 	var tlsConf tls.Config
+	tlsConf.MinVersion = tls.VersionTLS12
```
yizhaodev, line 50: can we make the TLS min version configurable, with a default of TLS 1.2?
Some customer environments may still run redis endpoints that only support TLS 1.1.

lioraron: TLS 1.1 has been deprecated since RFC 8996 (March 2021) and is considered insecure — Go itself logs deprecation warnings when TLS 1.0/1.1 connections are negotiated. Since this PR exists specifically to address a security finding, making the minimum version configurable to allow TLS 1.1 would undermine the fix?
If a customer environment has a Redis endpoint that only supports TLS 1.1, possibly a better path would be to upgrade that Redis instance rather than weakening the gateway's TLS floor?

lioraron (approved):

internal/util/tls/tls.go
j-mok-dev: I’d like to keep this PR scoped to the security finding and leave the minimum at TLS 1.2.

Allowing TLS 1.1 through configuration would weaken the remediation, and TLS 1.1 is already deprecated. If we have a concrete customer compatibility need, I think that should be discussed separately as a follow-up.

PR #492, opened 2026-06-17
docs: rewrite deploy-demo README to match deploy-k8s.sh

yizhaodev:
Rename `README.md` to `deploy-k8s.md` and rewrite to match current script behavior.


examples/deploy-demo/deploy-k8s.md
```diff
@@ -0,0 +1,148 @@
+# Demo Scripts
```
lioraron, line 1: Renaming `README.md` to `deploy-k8s.md` means GitHub will no longer auto-render documentation when browsing `examples/deploy-demo/`. Visitors see a raw file listing with no context.

Since there is only one script in this directory, keeping the file as `README.md` would preserve discoverability. Alternatively, a minimal `README.md` with a link to `deploy-k8s.md` would also work.

yizhaodev: midstream will add `deploy-rhoai.sh` with its own `deploy-rhoai.md`. 

Keeping script-specific doc names instead of `README.md` avoids conflicts during upstream/midstram sync

lioraron (approved):

PR #497, opened 2026-06-17
test: add regression test suite for backward compatibility validation

### LINKED ISSUE #468: Add dedicated regression test suite for backward compatibility validation
## Summary

There is no dedicated regression test suite or CI gate. A few unit/e2e tests have inline "regression guard" comments, but there is no systematic approach to ensuring backward compatibility of the OpenAI-compatible API or preventing performance regressions.

## Motivation

- The project commits to no breaking API changes after GA
- Performance regressions can silently degrade batch processing throughput
- Explicit regression tests make the compatibility contract enforceable in CI

## Scope

Candidate areas for regression tests:
- **API compatibility:** Validate request/response schemas against OpenAI Batch API spec snapshots
- **Performance baselines:** Benchmark-based gates for critical paths (batch creation, file parsing, worker throughput)
- **Behavioral regressions:** Known-bug reproduction tests tagged as regression (prevent re-introduction)

## Acceptance Criteria

- [ ] Convention established (e.g., `//go:build regression` tag or `test/regression/` directory)
- [ ] API schema compatibility tests against a pinned OpenAI spec snapshot
- [ ] Benchmark regression detection integrated into CI (e.g., `benchstat` comparison)
- [ ] Makefile target `make test-regression` added
- [ ] At least critical-path regressions from past bugs documented as test cases

vishbhat:
## Why is this PR needed?

This PR addresses [#468](https://github.com/llm-d/llm-d-batch-gateway/issues/468) by establishing a dedicated regression test suite for backward compatibility validation of the OpenAI-compatible API surface.

Currently, the project lacks systematic coverage for:

- **API Schema Integrity**: Changes to struct field names or JSON serialization tags could pass undetected
- **Wire Format Stability**: Modifications to field serialization behavior (e.g., `omitempty` tags) could inadvertently alter API responses
- **Type Compatibility**: Additions or removals of fields in core response types risk breaking OpenAI API compatibility

While existing unit and integration tests validate specific business logic, they do not enforce the structural invariants of the complete API response schema.

## What does this PR do?

This PR introduces a focused regression test suite comprising:

### Schema Compatibility Tests (7 test cases)
- Validates JSON key structure preservation across marshal/unmarshal cycles for `Batch`, `ListBatchResponse`, `CreateBatchRequest`, `FileObject`, `ListFilesResponse`, `FileDeleteResponse`, and `ErrorResponse`
- Detects unintended schema changes such as field renames, field removal, or missing JSON serialization directives
- Implemented in `test/regression/schema_batch_test.go`, `test/regression/schema_file_test.go`, `test/regression/schema_error_test.go` with supporting golden-file snapshots

### Serialization Behavior Guards (3 test cases)
- Verifies that optional fields are omitted from JSON output when zero-valued (e.g., `Batch.model`, `FileObject.status_details`, `CreateBatchRequest.metadata`)
- Prevents accidental modification of struct serialization tags
- Implemented in `test/regression/bug_regressions_test.go`

### Build and Documentation
- Isolated via `//go:build regression` tag; excluded from `make test`, executed via `make test-regression`
- Integrated into `make test-all` test pipeline
- Documented in CLAUDE.md, CONTRIBUTING.md, and new test/regression/README.md

## How was this tested?

## Checklist

## Related Issues

Fixes #468

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,7 +313,7 @@ deps-verify:
 ## test-regression: Run regression tests (API schema compat, past-bug guards)
 test-regression:
 	@echo "Running regression tests..."
-	@$(GO) test $(TEST_FLAGS) -v -tags=regression -count=1 ./test/regression/... || \
+	@$(GO) test $(TEST_FLAGS) -v -count=1 ./test/regression/... || \
 		(echo "\n❌ Regression tests failed" && exit 1)
 	@echo "\n✅ Regression tests passed!"
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help build build-apiserver build-processor build-gc run-apiserver run-processor run-gc run-apiserver-dev run-processor-dev run-gc-dev build-release package-release publish-helm-chart generate-release test test-coverage test-coverage-func clean lint fmt vet tidy install-tools deps-get deps-verify bench check check-container-tool ci image-build image-build-apiserver image-build-processor image-build-gc test-integration test-all test-e2e test-helm dev-deploy dev-clean dev-rm-cluster pre-commit
+.PHONY: help build build-apiserver build-processor build-gc run-apiserver run-processor run-gc run-apiserver-dev run-processor-dev run-gc-dev build-release package-release publish-helm-chart generate-release test test-coverage test-coverage-func clean lint fmt vet tidy install-tools deps-get deps-verify bench check check-container-tool ci image-build image-build-apiserver image-build-processor image-build-gc test-regression test-integration test-all test-e2e test-helm dev-deploy dev-clean dev-rm-cluster pre-commit
 
 SHELL := /usr/bin/env bash
 
@@ -310,6 +310,13 @@ deps-verify:
 	@echo "Verifying dependencies..."
 	$(GO) mod verify
 
+## test-regression: Run regression tests (API schema compat, past-bug guards)
+test-regression:
+	@echo "Running regression tests..."
+	@$(GO) test $(TEST_FLAGS) -v -tags=regression -count=1 ./test/regression/... || \
+		(echo "\n❌ Regression tests failed" && exit 1)
+	@echo "\n✅ Regression tests passed!"
+
 ## test-integration: Run integration tests (in-process server with mock backends and external service integration, no cluster needed)
 test-integration:
 	@echo "Running integration tests..."
@@ -317,8 +324,8 @@ test-integration:
 		(echo "\n❌ Integration tests failed" && exit 1)
 	@echo "\n✅ Integration tests passed!"
 
-## test-all: Run all tests (unit + integration)
-test-all: test test-integration
+## test-all: Run all tests (unit + regression + integration)
+test-all: test test-regression test-integration
 
 KIND_CLUSTER_NAME ?= batch-gateway-dev
 
--- a/test/regression/regression_test.go
+++ b/test/regression/regression_test.go
@@ -22,7 +22,7 @@ import (
 	"encoding/json"
 	"os"
 	"path/filepath"
-	"reflect"
+	"slices"
 	"sort"
 	"testing"
 )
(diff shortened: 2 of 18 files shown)


test/regression/schema_file_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+package regression
+
+import (
+	"testing"
+
+	"github.com/llm-d/llm-d-batch-gateway/internal/shared/openai"
+)
+
+func TestSchemaCompat_FileObject(t *testing.T) {
```
j-mok-dev: I think these functions should be grouped. example: `schema_batch_test.go`

test/regression/regression_test.go
```diff
@@ -0,0 +1,156 @@
...
+	return keys
+}
+
+func assertJSONKeysMatch(t *testing.T, label string, golden, actual []byte) {
+	t.Helper()
+
+	goldenKeys := jsonKeys(t, label+" (golden)", golden)
+	actualKeys := jsonKeys(t, label+" (actual)", actual)
+
+	if !reflect.DeepEqual(goldenKeys, actualKeys) {
```
j-mok-dev: instead of importing reflect here, we can use `slices.Equal` and drop the reflect dependency.

test/regression/regression_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	keys := make([]string, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func assertJSONKeysMatch(t *testing.T, label string, golden, actual []byte) {
```
j-mok-dev, line 51: the field type change cannot be checked in this test. for example, if *int64 becomes int64, then null becomes 0 which is a breaking change for api client. maybe we can add a guard to check if a field value in golden is null then it is still null after a roundtrip?

test/regression/schema_file_test.go
vishbhat: Thanks. Ill move them.

test/regression/regression_test.go
vishbhat: Thats a good suggestion. Thank you!

test/regression/regression_test.go
vishbhat: Thats a good point. Thanks!

j-mok-dev (approved):
LGTM!

Makefile
```diff
@@ -310,15 +310,22 @@ deps-verify:
 	@echo "Verifying dependencies..."
```
lioraron, line 310: The regression tests are not run in any CI pipeline. `pre-commit.yml` runs `go-unit-tests` (which excludes the `regression` build tag) and `make test-integration` — neither picks these up. No other workflow runs `make test-regression` or `make test-all`.

Without CI integration, these tests can only catch regressions if someone remembers to run them locally. Consider either adding `make test-regression` as a step in `pre-commit.yml` (they should be fast — just JSON marshal/unmarshal) or dropping the build tag and letting them run with regular unit tests.

vishbhat: Thanks @lioraron , thats a good point. I added regression tests to pre-commit.

test/regression/regression_test.go
```diff
@@ -0,0 +1,197 @@
+//go:build regression
```
lioraron: Consider dropping the `//go:build regression` tag. These tests are sub-second JSON round-trips — the same speed as unit tests. With the build tag, they are excluded from `go test ./...` and from the local `go-unit-tests` pre-commit hook, so a developer who changes a struct tag won't catch the break locally.

Without the tag, they would run as part of `go test ./...` both locally and in CI, with zero extra configuration. The dedicated CI step added in the latest commit would also become unnecessary.

You can still satisfy the "repo has regression tests" requirement through the directory structure (`test/regression/`) and the `make test-regression` target — just change the target to `go test ./test/regression/...` (no `-tags` flag needed). The tests are already clearly labeled by directory, README, and function name prefix (`TestRegression_`, `TestSchemaCompat_`).

vishbhat: Done

lioraron (approved):

PR #505, opened 2026-06-23
feat: benchmark infrastructure for batch-gateway effectiveness

kahilam:
## Summary
- Port benchmark scripts from evacchi/batch-gateway#1 into benchmarks/ directory
- Add generate_prompts.py with Faker-based prompt generation and system-prompt diversity (3-5 distinct prompts for prefix-cache evaluation)
- Add benchmark.py orchestrator supporting 6 scenarios (0-5)
- Add setup.sh / teardown.sh for per-scenario K8s deployment
- Add README.md with methodology, usage, and interpretation guide

## Test plan
- [x] `generate_prompts.py` produces valid OpenAI-format JSONL
- [x] `benchmark.py --help` shows all flags
- [x] Scripts are syntactically valid (`bash -n setup.sh`, `bash -n teardown.sh`, `python3 -m py_compile benchmark.py`)
- [x] Bug fixes applied (indentation in generate_prompts.py, truncated variable in setup.sh, truncated function in teardown.sh)

Cluster-level validation (CoreWeave deployment, Kind + inference-sim e2e) deferred to PR 2 / issue #491.


benchmarks/generate_prompts.py
```diff
@@ -0,0 +1,220 @@
...
+                    "model": model,
+                    "messages": [
+                        {"role": "system", "content": system_prompt},
+                        {"role": "user", "content": user_prompt},
+                    ],
+                },
+            }
+            f.write(json.dumps(line, separators=(",", ":")) + "\n")
+
+        if (i + 1) % 500 == 0:
```
lioraron: Bug: this `if` block is outside the `for` loop — it's at the `for` indentation level (8 spaces), not inside it (12 spaces). It only runs once after the loop finishes with the final value of `i`, so:
- For `num_requests < 500`: never prints (correct behavior is the same, so harmless)
- For `num_requests >= 500`: prints once after the loop with the last `i` value, instead of every 500 iterations during generation

Should be indented one more level to be inside the `for` loop:
```python
            f.write(json.dumps(line, separators=(",", ":")) + "\n")

            if (i + 1) % 500 == 0:
                print(f"  Generated {i + 1}/{num_requests}", file=sys.stderr)
```

benchmarks/setup.sh
```diff
@@ -0,0 +1,203 @@
...
+#   GUIDE_NAME         — inference pool name (default: optimized-baseline)
+#   BG_IMAGE_REPO      — batch-gateway image repo override
+#   BG_IMAGE_TAG       — batch-gateway image tag override
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+
+# Defaults
+MODEL="${MODEL:-Qwen/Qwen3-8B}"
+GUIDE_NAME="${GUIDEaseline}"
```
lioraron: Bug: this variable assignment is corrupted — `${GUIDEaseline}` is not a valid variable reference. Should be:
```bash
GUIDE_NAME="${GUIDE_NAME:-optimized-baseline}"
```
(matching `teardown.sh` line 25 and the documented default in the header comment above)

benchmarks/teardown.sh
```diff
@@ -0,0 +1,90 @@
...
+
+if [ -z "${KUBE_CONTEXT:-}" ]; then
+    echo "ERROR: KUBE_CONTEXT is not set" >&2
+    exit 1
+fi
+
+K="kubectl --context=${KUBE_CONTEXT}"
+H="helm --kube-context=${KUBE_CONTEXT}"
+
+lo "[$(date +%H:%M:%S)] $*"; }
```
lioraron: Bug: `log()` function definition is truncated — `lo "..."` instead of `log() { echo "..."`. Compare with `setup.sh` line 48 which has the correct form:
```bash
log() { echo "[$(date +%H:%M:%S)] $*"; }
```
This will cause a syntax error when running `teardown.sh` — bash will try to execute `lo` as a command and the unmatched `}` will fail.

lioraron:
## Reproducibility: local repo checkouts vs published artifacts

`setup.sh` requires `LLM_D_REPO` and `ROUTER_REPO` pointing to local git checkouts of `llm-d` and `llm-d-router`. This means benchmark results depend on whatever commit each operator happens to have checked out — two people running the "same" benchmark can get different results for reasons unrelated to batch-gateway.

The default should be published OCI artifacts with pinned versions:
```bash
# Instead of:
helm install "${GUIDE_NAME}" "${ROUTER_REPO}/config/charts/llm-d-router-gateway/" \
    -f "${LLM_D_REPO}/guides/recipes/router/base.values.yaml" ...

# Default to:
helm install "${GUIDE_NAME}" oci://ghcr.io/llm-d/llm-d-router-gateway --version 0.3.0
```

This applies to all non-infrastructure components:
- **Router chart** — pin to a released OCI version (currently installed from local `ROUTER_REPO`)
- **Router values / guides** — pin to a tagged version from `llm-d` (currently installed from local `LLM_D_REPO`)
- **batch-gateway chart** — pin to a released version or explicit `--set-string image.tag=...` (currently installed from local `REPO_ROOT`)
- **vLLM image tag** — pin in the Kustomize overlay (currently unpinned)

Redis and PostgreSQL don't need pinning — they're infrastructure dependencies where version drift won't affect benchmark results.

Benefits of pinning as default:
- **Reproducible** — pinned versions, anyone gets the same stack
- **Simpler prerequisites** — no need to clone 2 extra repos
- **Explicit compatibility** — benchmark documents which component versions were tested

Local repo overrides via `LLM_D_REPO`/`ROUTER_REPO` env vars should still be supported as an opt-in for development — but the default path should work with just the batch-gateway repo.

See the [artifact sourcing section](https://github.com/llm-d/llm-d-batch-gateway/issues/491#artifact-sourcing) added to issue #491.

lioraron:
## Benchmark reproducibility: parameter versioning and run metadata

Two related gaps:

### 1. No canonical parameter profile

All benchmark parameters are CLI args with hardcoded defaults scattered across `benchmark.py` and `generate_prompts.py`. If anyone changes a default or passes different flags, results become incomparable to previous runs with no way to detect the drift.

Suggest a checked-in config file (e.g., `benchmarks/profiles/default.yaml`) that both scripts read by default:
```yaml
# benchmarks/profiles/default.yaml
prompt:
  num_requests: 1000
  num_system_prompts: 5
  prompt_tokens: 256
  seed: 42
benchmark:
  burst_rate: 15
  idle_rate: 1
  burst_seconds: 90
  idle_seconds: 90
  cycles: 3
  batch_size: 1000
  num_jobs: 3
  model: Qwen/Qwen3-8B
```
CLI flags would override for development, but the default path produces comparable runs.

### 2. Incomplete parameter recording in results

The HTML report's Configuration table includes model, burst/idle rates, durations, cycles, and job count — but is missing `prompt_tokens`, `num_system_prompts`, `seed`, and stack versions (router, vLLM, batch-gateway).

More importantly, there's no machine-readable output. A `run-metadata.json` written alongside each run's results would enable:
- Programmatic comparison between runs
- Automated drift detection ("these two runs used different parameters")
- CI integration that validates parameter consistency

```json
{
  "profile": "default",
  "parameters": { "...all params..." },
  "versions": {
    "batch_gateway": "v0.2.0",
    "router": "v0.3.0",
    "vllm": "0.8.5"
  },
  "timestamp": "2026-06-23T14:30:00Z"
}
```

kahilam:
@lioraron , added benchmarks/profiles/default.yaml as the canonical parameter profile (both scripts read it by default, CLI flags still override). Also added run-metadata.json output with all parameters, stack versions, and timestamp. HTML report now includes prompt config too.

lioraron:
@kahilam — heads up: I updated [this earlier comment](https://github.com/llm-d/llm-d-batch-gateway/pull/505#issuecomment-4781555366) to be more specific about what should be version-pinned. In addition to the router chart, the list now includes:

- **batch-gateway chart** — pin to a released version or explicit image tag
- **vLLM image tag** — pin in the Kustomize overlay
- **Router values / guides** — pin to a tagged version from `llm-d`

Redis and PostgreSQL don't need pinning — they're infrastructure dependencies that won't affect benchmark results.

benchmarks/teardown.sh
```diff
@@ -0,0 +1,90 @@
...
+
+    log "  ${NS}: resources deleted"
+
+    # Delete namespace
+    ${K} delete ns "${NS}" --ignore-not-found 2>/dev/null || true
+    log "  ${NS}: namespace deleted"
+}
+
+# Handle --all flag
+if [ "${1:-}" = "--all" ]; then
```
lioraron: The `--all` CLI flag breaks the env-var convention used everywhere else. Consider using `SCENARIO=all` instead for consistency:
```bash
# Instead of:
KUBE_CONTEXT=my-ctx ./benchmarks/teardown.sh --all

# Use:
KUBE_CONTEXT=my-ctx SCENARIO=all ./benchmarks/teardown.sh
```
Then this block becomes:
```bash
if [ "${SCENARIO}" = "all" ]; then
    ...
fi
```

benchmarks/README.md
```diff
@@ -0,0 +1,228 @@
...
+├── setup.sh                  # Deploy full stack per scenario
+├── teardown.sh               # Cleanup
+├── README.md                 # This file
+├── helm-values/
+│   ├── scenario-0-interactive-only.yaml
+│   ├── scenario-1-no-batch-gateway.yaml
+│   ├── scenario-2-ungated.yaml
+│   ├── scenario-3-aimd.yaml
+│   ├── scenario-4-aimd-flow-control.yaml
+│   └── scenario-5-async.yaml
```
lioraron: For sub-dirs - maybe no need to specify the files, and instead only add a comment describing the dir?
There is also a new dir `profiles`.

benchmarks/README.md
```diff
@@ -0,0 +1,228 @@
...
+└── results/                  # gitignored — local run output
+    └── .gitkeep
+```
+
+## Related
+
+- [Issue #491](https://github.com/llm-d/llm-d-batch-gateway/issues/491) — Full benchmark spec
+- [Flow Control Setup Guide](../docs/guides/flow-control-setup.md) — AIMD + Router configuration
+- [Batch Processor Architecture](../docs/design/batch_processor_architecture.md) — Processor internals
+- [@evacchi's prior work](https://github.com/evacchi/batch-gateway/pull/1) — Original benchmark infrastructure
```
lioraron: This is a temp PR and can be removed from here.
This is already mentioned in the issue 491.

kahilam:
> @kahilam — heads up: I updated [this earlier comment](https://github.com/llm-d/llm-d-batch-gateway/pull/505#issuecomment-4781555366) to be more specific about what should be version-pinned. In addition to the router chart, the list now includes:
> 
>     * **batch-gateway chart** — pin to a released version or explicit image tag
> 
>     * **vLLM image tag** — pin in the Kustomize overlay
> 
>     * **Router values / guides** — pin to a tagged version from `llm-d`
> 
> 
> Redis and PostgreSQL don't need pinning — they're infrastructure dependencies that won't affect benchmark results.

Makes sense. For the pinned versions, should I use:

batch-gateway: the latest released chart/image tag (which version?)
vLLM: vllm/vllm-openai:0.8.5?
Router values: is there a tagged release of the llm-d guides I can reference in OCI mode?

lioraron:
@kahilam to answer your questions:

**batch-gateway chart** — use the local chart from `${REPO_ROOT}/charts/batch-gateway/`. The benchmark is testing batch-gateway itself, so pinning to a released version would defeat the purpose. Everything *else* should be pinned so that the only variable is the batch-gateway code under test.

**vLLM** — already pinned in your Kustomize overlay to `vllm/vllm-openai:v0.23.0`, so that's covered.

**Router chart** — latest release is `v0.9.2`. Use the published OCI artifact with `--version 0.9.2`.

**Router values / guides** — these aren't published as standalone OCI artifacts, they're files in the `llm-d` repo. Recommend downloading from a pinned git tag at setup time, e.g.:
```bash
LLM_D_TAG="${LLM_D_TAG:-v0.7.0}"
curl -sL "https://raw.githubusercontent.com/llm-d/llm-d/${LLM_D_TAG}/guides/recipes/router/base.values.yaml" -o /tmp/base.values.yaml
```
This eliminates the `LLM_D_REPO` prerequisite entirely — the benchmark works with just the batch-gateway repo cloned. The tag is pinned and deterministic, and `LLM_D_REPO` can still override for development (if set, use local files instead of downloading).

lioraron:
## Test plan scope

The two unchecked items in the test plan — CoreWeave deployment and Kind + inference-sim e2e — go beyond the scope of PR 1 (port infrastructure). This PR ports scripts and templates; it doesn't need to prove they run end-to-end on a cluster to merge.

Suggest the following test plan for this PR:
- [x] `generate_prompts.py` produces valid OpenAI-format JSONL
- [x] `benchmark.py --help` shows all flags
- [ ] Scripts are syntactically valid (`bash -n setup.sh`, `bash -n teardown.sh`, `python3 -m py_compile benchmark.py`)
- [ ] Bug fixes applied (indentation in `generate_prompts.py`, truncated variable in `setup.sh`, truncated function in `teardown.sh`)

The cluster-level validation should move to later PRs / follow-up tasks:
- **CoreWeave deployment** → PR 2 (baseline scenarios) test plan, where it becomes meaningful
- **Kind + inference-sim e2e** → new task in issue #491, since `setup.sh` doesn't currently support swapping inference-sim for vLLM

lioraron:
@kahilam FYI — I've added a new **PR 1.5: Kind + inference-sim e2e path** task to [issue #491](https://github.com/llm-d/llm-d-batch-gateway/issues/491), between PR 1 and PR 2. This would add a `MODE=sim` flag to `setup.sh` that substitutes inference-sim for vLLM on a local Kind cluster, enabling full-pipeline validation without CoreWeave access.

This means the CoreWeave deployment and Kind e2e items in this PR's test plan can be deferred — they're out of scope for the infrastructure port. See [this comment](https://github.com/llm-d/llm-d-batch-gateway/pull/505#issuecomment-4782237784) for the suggested test plan for this PR.

lioraron:
@kahilam When all the comments are addressed and the PR is ready for final review, please inform. Thanks!

kahilam:
@lioraron Updated the PR description — test plan now scoped to local validation only. Cluster-level checks deferred to PR 2 / #491.

lioraron:
Thanks @kahilam!
Is the PR ready for final review, or anything else remaining for this PR?

kahilam:
> Thanks @kahilam! Is the PR ready for final review, or anything else remaining for this PR?

I believe it's ready for final review.

lioraron (approved):
Thanks @kahilam, great work!

j-mok-dev:
Questions:
The README says scenario 1 sends batch prompts as regular requests via guidellm, but `benchmark.py` seems to only start extra batch load for `scenario >= 2`, so I may be missing where that scenario 1 traffic is launched.

Also, Kubernetes documents a 1 MiB limit for ConfigMap data, so I wonder whether the default generated JSONL still fits with the current ConfigMap-based transport.

kahilam:
> Questions: The README says scenario 1 sends batch prompts as regular requests via guidellm, but `benchmark.py` seems to only start extra batch load for `scenario >= 2`, so I may be missing where that scenario 1 traffic is launched.
> 
> Also, Kubernetes documents a 1 MiB limit for ConfigMap data, so I wonder whether the default generated JSONL still fits with the current ConfigMap-based transport.

Good catches:

Scenario 1 traffic: Correct — the batch-equivalent guidellm load for scenario 1 isn't wired up yet. I believe It's scoped for PR 2 (baseline scenarios) where we'll add a second guidellm job that sends batch prompts directly to the inference endpoint without batch-gateway.

ConfigMap 1 MiB limit: Valid concern. With 1000 requests x ~256 tokens, the JSONL can exceed 1 MiB. I'll either reduce the default to stay within limits or switch to PVC-based upload for the data payload. Can address in a follow-up.

PR #506, opened 2026-06-23
test: re-resolve processor observability for e2e runs

j-mok-dev:
## Why is this PR needed?

The previous attempt in #450 removed the fixed processor observability port, but the replacement still assumed that the processor endpoint would be directly reachable from the host. That does not hold reliably in local kind + Colima setups, where auto-detected NodePort addresses may not be reachable and long-lived port-forwards can break across processor restarts.

This PR supersedes #450 by moving processor observability resolution closer to the actual e2e checks instead of trying to maintain one stable host endpoint for the whole test run.

## What does this PR do?

- updates local dev deployment so processor observability is exposed through a service intended for local port-forward access rather than a fixed host-reachable port
- changes e2e processor observability checks to resolve access on demand in the Go test helpers, so metrics, readiness, and pprof checks still work after processor restarts and Helm upgrades
- aligns the local dev and e2e documentation with the new access model

## How was this tested?

## Checklist

## Related Issues

- Supersedes #450

### DIFF
--- a/test/e2e/helpers_test.go
+++ b/test/e2e/helpers_test.go
@@ -15,6 +15,7 @@
 package e2e_test
 
 import (
+	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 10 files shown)


test/e2e/helpers_test.go
```diff
@@ -663,3 +664,139 @@ func fetchOutputFile(t *testing.T, batch *openai.Batch) string {
...
+	})
+
+	return status, body, err
+}
+
+func waitForProcessorReady(t *testing.T, timeout time.Duration) {
+	t.Helper()
+
+	deadline := time.Now().Add(timeout)
+	var lastErr error
```
lioraron, line 981: `waitForProcessorReady` spawns a new `kubectl port-forward` for every poll iteration (via `readProcessorObsEndpoint` → `withProcessorObsURL`). Two concerns:

1. **Fatal vs retryable port-forward failures.** `withProcessorObsURL` calls `t.Fatalf` if kubectl port-forward fails to start (e.g., no ready pods matching the service selector). In `doTestPodDeleteMidJob` (line 85 of `processor_graceful_shutdown_test.go`), `waitForProcessorReady` is called right after deleting the processor pod — there is no `waitForRollout` before it. If the replacement pod is not yet scheduled when the first iteration runs, `kubectl port-forward svc/...` will fail with "unable to find matching pod" and `withProcessorObsURL` will fatal instead of letting the loop retry. The old `waitForReady` handled this gracefully — `http.Get` would return a connection error, the loop would sleep and retry.

2. **Per-iteration kubectl overhead.** Each port-forward setup takes 1-3 seconds (process spawn + port negotiation), so the effective poll interval is 2-4 seconds rather than 1 second. Over a 2-3 minute timeout this is fine functionally, but it's worth being aware of.

Suggestion for (1): have `withProcessorObsURL` return an error for port-forward startup failures instead of calling `t.Fatalf`, so callers like `waitForProcessorReady` can treat them as retryable. Callers that need fatal-on-failure can wrap it themselves.

lioraron:
The PR motivation is that NodePort-based processor observability is not reliably reachable in Kind + Colima setups. Could you share evidence that the new port-forward approach was validated in those environments — e.g., `make test-e2e` output from a Kind + Colima setup showing the processor observability tests passing?

j-mok-dev:
> The PR motivation is that NodePort-based processor observability is not reliably reachable in Kind + Colima setups. Could you share evidence that the new port-forward approach was validated in those environments — e.g., `make test-e2e` output from a Kind + Colima setup showing the processor observability tests passing?

Yes. I re-ran this locally on macOS with Colima + Kind and captured the output.

Validation I ran:
- `colima status`
- `make dev-rm-cluster`
- `make dev-deploy-gie`
- `make test-e2e`

From that local run:
- `TestE2E/Observability/Processor/health` passed
- `TestE2E/Observability/Processor/ready` passed
- `TestE2E/Observability/Processor/metrics` passed
- `TestE2E/Observability/Pprof/Processor` passed
- `TestE2E/FlowControl/GIE/*` passed

Final result:
- `Passed: 69 | Failed: 0 | Skipped: 1`

deployment
e2e_result

test/e2e/helpers_test.go
j-mok-dev: Thanks. I updated the processor observability helper so port-forward startup failures are now returned as retryable errors instead of failing the test immediately.
`waitForProcessorReady()` now retries transient startup failures during pod replacement, which preserves the old readiness polling behavior for cases where the replacement processor pod is not ready yet.
I also added cleanup for failed port-forward attempts so retries do not leave orphaned `kubectl port-forward` processes behind.
I reran the e2e suite after the change, and it all passed.

test/e2e/observability_test.go
```diff
@@ -120,6 +119,33 @@ func testPprof(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("Processor", func(t *testing.T) {
+		pf, err := startProcessorObsPortForward(t)
+		if err != nil {
```
lioraron, line 125: `t.Skipf` on port-forward failure silently hides a broken processor observability endpoint — the test suite reports green when it should report red.

Compare with `doTestProcessorObservabilityEndpoints` (line 218 in this file), which correctly uses `t.Fatalf` for port-forward failure. The inconsistency means health/ready/metrics failures are caught, but pprof failures are silently swallowed.

`t.Skipf` is appropriate for the `StatusNotFound` case (pprof intentionally disabled), but port-forward establishment failure is an infrastructure problem that should fail the test.

```go
pf, err := startProcessorObsPortForward(t)
if err != nil {
    t.Fatalf("processor port-forward failed: %v", err) // not Skipf
}
defer pf.Close()

status, _, err := pf.Read("/debug/pprof/")
if err != nil {
    t.Fatalf("GET /debug/pprof/ failed: %v", err) // not Skipf
}
if status == http.StatusNotFound {
    t.Skip("pprof not enabled on processor observability endpoint")
}
```

test/e2e/helpers_test.go
```diff
@@ -663,3 +664,198 @@ func fetchOutputFile(t *testing.T, batch *openai.Batch) string {
...
+			_ = cmd.Process.Kill()
+		}
+		if waitDone != nil {
+			<-waitDone
+		}
+		_ = reader.Close()
+	}()
+
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("start processor port-forward: %w", err)
```
lioraron, line 870: The scanner goroutine (lines 719-726) is never stopped on the success path — it keeps reading from `reader` and sending to `lineCh`. Once kubectl emits 16+ lines of output after the "Forwarding from" line (connection logs, warnings, etc.), the goroutine deadlocks on `lineCh <- scanner.Text()` because nobody reads from `lineCh` after the function returns.

This violates the project convention: _"Goroutines: Never fire-and-forget. Always provide a shutdown signal mechanism."_

Also, the buffer size of 16 needs justification per CLAUDE.md (_"Prefer small channel buffers (0 or 1). Larger buffers need justification"_) — it masks the leak by delaying the deadlock.

Suggestion: store a `context.CancelFunc` on the struct and check it in the goroutine's send, or restructure so the goroutine terminates once the forwarding line is found. Note that `Close()` killing the process + closing `reader` will unblock `scanner.Scan()`, but only if the goroutine isn't stuck on the channel send.

j-mok-dev: Thanks. I updated `startProcessorObsPortForward()` so the scanner goroutine only participates in startup detection until the initial `Forwarding from ...` line is observed.

After startup, it stops sending to the startup channel and only continues draining output until shutdown, and both `Close()` and startup-failure cleanup now wait for the scanner to exit.

e2e test all passed.

test/e2e/observability_test.go
j-mok-dev: Thanks. I updated the processor pprof test so port-forward establishment failures and `/debug/pprof/` read failures now fail the test instead of being skipped.

We now only skip the processor pprof check when the endpoint returns `404`, which preserves the existing "pprof disabled" behavior without masking a broken observability path.

lioraron (approved):

PR #507, opened 2026-06-24
feat: Kind + inference-sim e2e path and workload shape alignment

kahilam:
## Summary

- Add `MODE=sim` to `setup.sh` — deploys inference-sim instead of real vLLM (no GPU needed, skips router/Istio)
- Replace fixed `--prompt-tokens` with configurable ISL/OSL distributions (lognormal, normal, uniform, fixed)
- Scale system prompt config to 32 prompts x 2048 tokens (matching llm-d ecosystem range)
- Add TPOT metric (p50/p95/p99) to PhaseMetrics and HTML report
- Add `make benchmark-local` / `make benchmark-local-teardown` Makefile targets

## Part of #491 (PR 1.5 of 4)

Enables full-pipeline validation on a local Kind cluster without CoreWeave access.

## Test plan

- [x] `bash -n setup.sh` — syntax valid
- [x] `python3 -m py_compile benchmark.py` — compiles
- [x] `python3 benchmarks/generate_prompts.py --help` — shows ISL/OSL flags
- [x] `make benchmark-local` — full e2e on Kind: setup, generate, benchmark, report
- [x] `run-metadata.json` written with correcrs and metrics list

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -353,6 +353,10 @@ BENCHMARK_RESULTS_DIR ?= benchmarks/results/local-run
 
 ## benchmark-local: Run benchmark e2e on local Kind cluster with inference-sim (no GPU required)
 benchmark-local:
+	@kind get clusters 2>/dev/null | grep -q $(KIND_CLUSTER_NAME) || \
+		{ echo "ERROR: Kind cluster '$(KIND_CLUSTER_NAME)' not found. Run 'make dev-deploy' first."; exit 1; }
+	@docker exec $(KIND_CLUSTER_NAME)-control-plane crictl images 2>/dev/null | grep -q batch-gateway-apiserver || \
+		{ echo "ERROR: batch-gateway images not loaded in Kind. Run 'make dev-deploy' first."; exit 1; }
 	@echo "=== Benchmark local e2e (MODE=sim, scenario $(BENCHMARK_SCENARIO)) ==="
 	@echo "Step 1/4: Setting up infrastructure..."
 	@MODE=sim KUBE_CONTEXT=$(BENCHMARK_CONTEXT) SCENARIO=$(BENCHMARK_SCENARIO) BG_IMAGE_TAG=0.0.1 BG_PULL_POLICY=IfNotPresent bash benchmarks/setup.sh
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help build build-apiserver build-processor build-gc run-apiserver run-processor run-gc run-apiserver-dev run-processor-dev run-gc-dev build-release package-release publish-helm-chart generate-release test test-coverage test-coverage-func clean lint fmt vet tidy install-tools deps-get deps-verify bench check check-container-tool ci image-build image-build-apiserver image-build-processor image-build-gc test-regression test-integration test-all test-e2e test-helm dev-deploy dev-clean dev-rm-cluster pre-commit
+.PHONY: help build build-apiserver build-processor build-gc run-apiserver run-processor run-gc run-apiserver-dev run-processor-dev run-gc-dev build-release package-release publish-helm-chart generate-release test test-coverage test-coverage-func clean lint fmt vet tidy install-tools deps-get deps-verify bench check check-container-tool ci image-build image-build-apiserver image-build-processor image-build-gc test-regression test-integration test-all test-e2e test-helm dev-deploy dev-clean dev-rm-cluster pre-commit benchmark-local benchmark-local-teardown
 
 SHELL := /usr/bin/env bash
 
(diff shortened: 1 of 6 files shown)


lioraron (commented):
Review of the KIND + inference-sim e2e path and workload shape alignment changes.

Overall the approach is sound — the sim/gpu mode split is clean, the ISL/OSL distribution support is well-structured, and the legacy compat shim for `--prompt-tokens` is a nice touch. A few actionable findings below.

benchmarks/generate_prompts.py
```diff
@@ -88,29 +86,56 @@
...
+    elif distribution == "normal":
+        val = rng.gauss(mean, stdev)
+    elif distribution == "uniform":
+        low = max(min_val, mean - stdev)
+        high = mean + stdev
+        val = rng.uniform(low, high)
+    else:
+        val = mean
+
+    return max(min_val, int(round(val)))
```
lioraron: **Bug: only 5 distinct prefix-cache groups despite `num_system_prompts=32` default.**

The template cycling (`i % len(SYSTEM_PROMPT_TEMPLATES)`) means prompts 0, 5, 10, 15, 20, 25, 30 all start with identical ~200-char "senior software engineer" base text. The faker padding added afterward does differ (faker state advances), but for prefix-cache evaluation — the stated purpose — a KV cache compares token-by-token from the left. Prompts sharing a ~50-token prefix will all cache-hit against each other.

With `system_prompt_tokens=2048`, only the first ~10% of each prompt is template-derived; the rest is padding. So you effectively have 5 prefix-cache groups, not 32.

Options:
1. Add more template variants (at least 32)
2. Place padding *before* the template text so the prefix is always unique
3. Inject a unique identifier early in the prompt (e.g. `f"[Persona {i}] " + template`) to break prefix sharing

Makefile
```diff
@@ -347,6 +347,27 @@ dev-rm-cluster:
...
+BENCHMARK_CONTEXT ?= kind-$(KIND_CLUSTER_NAME)
+BENCHMARK_SCENARIO ?= 3
+BENCHMARK_RESULTS_DIR ?= benchmarks/results/local-run
+
+## benchmark-local: Run benchmark e2e on local Kind cluster with inference-sim (no GPU required)
+benchmark-local:
+	@echo "=== Benchmark local e2e (MODE=sim, scenario $(BENCHMARK_SCENARIO)) ==="
+	@echo "Step 1/4: Setting up infrastructure..."
+	@MODE=sim KUBE_CONTEXT=$(BENCHMARK_CONTEXT) SCENARIO=$(BENCHMARK_SCENARIO) BG_IMAGE_TAG=0.0.1 BG_PULL_POLICY=IfNotPresent bash benchmarks/setup.sh
+	@echo "Step 2/4: Generating prompts..."
```
lioraron, line 363: **Bug: `benchmark-local` passes `BG_IMAGE_TAG=0.0.1 BG_PULL_POLICY=IfNotPresent` but never builds or loads the images into Kind.**

`dev-deploy.sh` handles this (runs `make image-build`, then `kind load docker-image`), but `benchmark-local` skips straight to `setup.sh`. The Helm install will fail with `ImagePullBackOff` because the `0.0.1` tag doesn't exist in any registry and wasn't loaded into Kind.

Either:
1. Add image build + kind-load steps before calling `setup.sh`, or
2. Add a prerequisite comment/check that `make dev-deploy` (or at least `make image-build` + kind load) was run first, or
3. Make `benchmark-local` depend on a `benchmark-local-images` target that handles build + load

benchmarks/setup.sh
```diff
@@ -191,6 +278,23 @@ if [ -n "${VALUES_FILE}" ]; then
...
+            --set "processor.config.modelGateways.${MODEL}.url=http://inference-sim.${NAMESPACE}.svc.cluster.local:8000"
+            --set "processor.config.modelGateways.${MODEL}.requestTimeout=5m"
+            --set "processor.config.modelGateways.${MODEL}.maxRetries=3"
+            --set "processor.config.modelGateways.${MODEL}.initialBackoff=1s"
+            --set "processor.config.modelGateways.${MODEL}.maxBackoff=60s"
+        )
+    fi
 
     ${H} install batch-gateway \
         "${REPO_ROOT}/charts/batch-gateway/" \
```
lioraron, line 304: **Correctness concern: sim mode merges `sim-model` gateway entry into existing `modelGateways` from the scenario values file.**

The scenario-3 values file already defines `modelGateways` for `Qwen/Qwen3-0.6B`, `Qwen/Qwen3-8B`, and `Qwen/Qwen3-32B-AWQ`, all pointing to `http://llm-d-inference-gateway-istio:80` — which doesn't exist in sim mode (no Istio, no router).

Helm `--set` will *merge* the new `sim-model` entry alongside those existing Qwen entries. The processor will start with gateway configs pointing to unreachable URLs. This probably doesn't cause a startup failure (the processor only dials on-demand), but it's unnecessary config pollution and could be confusing when debugging.

Consider using `--set processor.config.modelGateways=null` before adding the sim-model entry, or provide a dedicated sim-mode values file that only defines `sim-model`.

Makefile
```diff
@@ -347,6 +347,27 @@ dev-rm-cluster:
 	@kind delete cluster --name batch-gateway-dev || echo "Cluster not found or already deleted"
 	@echo "✅ Kind cluster deleted"
 
+BENCHMARK_CONTEXT ?= kind-$(KIND_CLUSTER_NAME)
+BENCHMARK_SCENARIO ?= 3
+BENCHMARK_RESULTS_DIR ?= benchmarks/results/local-run
+
+## benchmark-local: Run benchmark e2e on local Kind cluster with inference-sim (no GPU required)
+benchmark-local:
```
lioraron, line 355: **Nit: `benchmark-local` doesn't verify a Kind cluster exists before running.**

`setup.sh` will fail at the first `kubectl` command if `kind-batch-gateway-dev` doesn't exist. A guard like `kind get clusters | grep -q batch-gateway-dev || { echo "Run 'make dev-deploy' first"; exit 1; }` at the start would give a clear error message instead of a cryptic kubectl failure.

benchmarks/generate_prompts.py
```diff
@@ -120,31 +145,44 @@ def generate_user_prompt(fake: Faker, target_chars: int) -> str:
...
-    target_chars = prompt_tokens * 4
+    system_prompts = generate_system_prompts(num_system_prompts, seed, system_prompt_tokens)
 
     with open(output_path, "w") as f:
         for i in range(num_requests):
             system_prompt = system_prompts[i % num_system_prompts]
+
+            # Sample ISL for this request (user prompt portion)
+            isl_tokens = sample_from_distribution(rng, isl_distribution, isl_mean, isl_stdev, min_val=16)
+            target_chars = isl_tokens * 4
```
lioraron, line 180: **Missing max_val clamping — generated ISL can exceed `max-model-len`.**

`sample_from_distribution` has `min_val` but no `max_val`. With the default lognormal(1500, 1200), the long right tail regularly produces ISL > 4096 (the default `max-model-len`). The issue spec ([#491](https://github.com/llm-d/llm-d-batch-gateway/issues/491)) explicitly defines ISL max=4096 and OSL max=2048.

Without clamping, generated prompts can exceed the model context length and cause inference failures (or silent truncation).

Suggested fix: add a `max_val` parameter to `sample_from_distribution` and clamp the output:
```python
def sample_from_distribution(rng, distribution, mean, stdev, min_val=1, max_val=None):
    ...
    val = max(min_val, int(round(val)))
    if max_val is not None:
        val = min(val, max_val)
    return val
```
Then pass `max_val=4096` for ISL and `max_val=2048` for OSL. Also add `isl.max` / `osl.max` fields to `profiles/default.yaml` and corresponding CLI args.

benchmarks/setup.sh
```diff
@@ -12,25 +12,37 @@ set -euo pipefail
...
 REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
 
 # Defaults
-MODEL="${MODEL:-Qwen/Qwen3-8B}"
+MODE="${MODE:-gpu}"
+if [ "${MODE}" = "sim" ]; then
+    MODEL="${MODEL:-sim-model}"
+else
+    MODEL="${MODEL:-Qwen/Qwen3-8B}"
+fi
```
lioraron, line 39: **Model revision should be pinned for reproducibility.**

HuggingFace models are git repos — `Qwen/Qwen3-8B` without a revision always pulls the latest `main` snapshot. Model authors can update weights, tokenizer configs, or architecture at any time, making results non-comparable across runs.

vLLM supports `--revision <commit-sha>` to pin to a specific model version. This should be:
1. Configurable via env var (e.g., `MODEL_REVISION`) and passed to the vLLM Kustomize overlay / args
2. Recorded in `run-metadata.json` alongside the other stack versions

Updated in [#491](https://github.com/llm-d/llm-d-batch-gateway/issues/491) artifact sourcing and run metadata sections.

lioraron:
**Documentation: update `benchmarks/README.md` for local Kind + sim usage.**

The README's "Local Testing (Kind + inference-sim)" section currently just says `make dev-deploy` with no actual instructions for running the benchmark pipeline locally. This PR adds `make benchmark-local` / `make benchmark-local-teardown` and `MODE=sim` in `setup.sh`, but doesn't update the README to document them.

Please add to that section:
- `make benchmark-local` usage (what it does, what it deploys, expected output)
- `make benchmark-local-teardown` for cleanup
- `MODE=sim` env var and the `SIM_IMAGE` / `SIM_TTFT` / `SIM_ITL` knobs
- Prerequisites (Kind cluster must exist — `make dev-deploy` or manual `kind create cluster`)
- Note that batch-gateway images must be built and loaded into Kind before running

kahilam:
Hi @lioraron  All review comments addressed.

lioraron (approved):
Looks good, thanks @kahilam!

PR #508, opened 2026-06-24
fix: support Python 3.8 in generate_prompts.py

lioraron:
## Summary

- Add `from __future__ import annotations` to `generate_prompts.py` to fix `TypeError: 'type' object is not subscriptable` on Python 3.8 (caused by `list[str]` type hint)
- Add `.venv/`, `__pycache__/`, and `*.pyc` to `.gitignore`

## Test plan

- [x] `python3 benchmarks/generate_prompts.py --help` works on Python 3.8
- [x] `make benchmark-local` completes successfully

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,8 @@ bin/
 .claude/
 .cursor/
 .codegraph/
+.venv/
+__pycache__/
+*.pyc
 benchmarks/results/*
 !benchmarks/results/.gitkeep
--- a/benchmarks/generate_prompts.py
+++ b/benchmarks/generate_prompts.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+from __future__ import annotations
 """
 Generate JSONL batch input files with configurable ISL/OSL distributions
 and system-prompt diversity.


yizhaodev:
look good to me
but I think python 3.8 is already end-of-life?

myabe we should create benchmarks/pyproject.toml to enforce 3.9 or above
for example:
```
  [project]
  requires-python = ">=3.9"
```

yizhaodev (approved):

lioraron:
> look good to me but I think python 3.8 is already end-of-life?
> 
> myabe we should create benchmarks/pyproject.toml to enforce 3.9 or above for example:
> 
> ```
>   [project]
>   requires-python = ">=3.9"
> ```

Thanks.
The `from __future__` import annotations is a one-line fix that keeps the script working e.g. on stock Ubuntu 20.04 (ships Python 3.8.10). The benchmark scripts only use basic stdlib + faker/pyyaml — nothing that needs 3.9+. Adding a version gate would create friction for contributors on older distros with no functional benefit?

PR #509, opened 2026-06-24
fix: prevent temp file collision in concurrent fs storage uploads

### LINKED ISSUE #504: bug: Cancel/InProgress test flake — output/error file mismatch after in-progress cancel
## Summary

`TestE2E/Batches/Cancel/InProgress` fails due to a data consistency issue in the cancel flow. The output and error files contain duplicate/misplaced entries that don't match the batch's completed/failed counts.

## Failing run

https://github.com/llm-d/llm-d-batch-gateway/actions/runs/27975894349

## Observed behavior

Batch has 25 requests: 5 "fast" + 20 "slow". After cancel:

- Batch status reports `completed=5, failed=20` (correct total = 25)
- Output file has **20 lines** (expected 5 — only the completed ones)
- Error file has **20 lines**
- Combined = 40, but total requests = 25
- The 5 "fast" custom IDs (`fast-1` through `fast-5`) are **missing** from both files
- The output file contains cancelled requests (with `"response": null, "error": {"code": "batch_cancelled"}`) that belong in the error file

```
output lines (20) + error lines (20) = 40, but total requests = 25
output lines (20) != completed count (5)
input custom_id "fast-1" not found in output or error file
input custom_id "fast-2" not found in output or error file
input custom_id "fast-3" not found in output or error file
input custom_id "fast-4" not found in output or error file
input custom_id "fast-5" not found in output or error file
```

## Expected behavior

- Output file should contain exactly the 5 completed requests (`fast-1` through `fast-5`)
- Error file should contain exactly the 20 cancelled requests (`slow-1` through `slow-20`)
- Output lines + error lines = total requests (25)

## Root cause hypothesis

Race condition in the cancel path: when a batch is cancelled while in progress, the file writer appears to write cancelled requests to the output file instead of (or in addition to) the error file, while the successfully completed "fast" requests are lost.

## Related issues

- Not #449 (NodePort collision / cluster setup)
- Not #496 (AIMD/Recovery flake)

vishbhat:
## Why is this PR needed?

`TestE2E/Batches/Cancel/InProgress` fails intermittently due to a data corruption bug in the filesystem storage client. When a batch is cancelled while in-progress, `uploadPartialResults` (and `finalizeJob`) upload the output and error files **concurrently** — two goroutines call `fs.Client.Store()` targeting the same tenant directory. `Store()` generates temp file names using `time.Now().UnixNano()`, and when both goroutines execute in the same nanosecond, they produce identical temp names. The second `Create` silently truncates the first goroutine's data, causing one file's content to overwrite the other.

This explains all symptoms from #504:
- Output file contains 20 cancelled entries instead of 5 completed ones
- Error file also has 20 entries — same data appears in both files (40 lines for 25 requests)
- `fast-1` through `fast-5` results are missing (destroyed by the truncation)

## What does this PR do?

Replaces the single `root.Create(tmpName)` call with `root.OpenFile()` using `O_CREATE|O_EXCL` flags inside a retry loop. `O_EXCL` is a kernel-level guarantee: if the temp file already exists, the call fails atomically with `os.ErrExist` instead of silently truncating. The retry generates a fresh nanosecond timestamp, which is guaranteed to differ after the failed syscall overhead.

Adds `TestStore_ConcurrentSameDir` — a regression test that launches 10 goroutines storing to the same directory concurrently and verifies each file has distinct, correct content.

## How was this tested?

## Checklist

## Related Issues

Fixes #504

### DIFF
--- a/internal/files_store/fs/client.go
+++ b/internal/files_store/fs/client.go
@@ -113,10 +113,20 @@ func (c *Client) Store(ctx context.Context, fileName, folderName string, fileSiz
 	}
 
 	// Create temp file name within the target directory.
-	tmpName := filepath.Join(dir, fmt.Sprintf(".tmp-%d", time.Now().UnixNano()))
-	tmpFile, err := c.root.Create(tmpName)
-	if err != nil {
-		return nil, err
+	var tmpFile *os.File
+	var tmpName string
+	for range 3 {
+		tmpName = filepath.Join(dir, fmt.Sprintf(".tmp-%d", time.Now().UnixNano()))
+		tmpFile, err = c.root.OpenFile(tmpName, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o600)
+		if err == nil {
+			break
+		}
+		if !errors.Is(err, os.ErrExist) {
+			return nil, err
+		}
+	}
+	if tmpFile == nil {
+		return nil, fmt.Errorf("failed to create unique temp file in %s after retries", dir)
 	}
 	defer func() {
 		// Clean up on error.
--- a/internal/files_store/fs/client.go
+++ b/internal/files_store/fs/client.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -59,6 +60,8 @@ type Client struct {
 
 // Compile-time check that Client implements api.BatchFilesClient.
 var _ api.BatchFilesClient = (*Client)(nil)
+
+var tmpCounter atomic.Int64
 
 // New creates a new filesystem-based BatchFilesClient.
 func New(basePath string) (*Client, error) {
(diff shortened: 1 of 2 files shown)


lioraron (approved):

internal/files_store/fs/client.go
```diff
@@ -113,10 +113,20 @@ func (c *Client) Store(ctx context.Context, fileName, folderName string, fileSiz
 	}
 
 	// Create temp file name within the target directory.
-	tmpName := filepath.Join(dir, fmt.Sprintf(".tmp-%d", time.Now().UnixNano()))
-	tmpFile, err := c.root.Create(tmpName)
-	if err != nil {
-		return nil, err
+	var tmpFile *os.File
+	var tmpName string
+	for range 3 {
```
j-mok-dev, line 121: Nice fix, thanks! One minor robustness concern: temp file uniqueness still depends on the timestamp, so adding clock-independent entropy (UUID/counter) would make this path more robust. Not blocking.

j-mok-dev (approved):

internal/files_store/fs/client.go
vishbhat: @j-mok-dev Please check https://github.com/llm-d/llm-d-batch-gateway/pull/509/changes/db6ad1849360c97be5a8420ed426e00b88fbf6c8 , is this okay?

j-mok-dev: yes. Thanks!!

PR #512, opened 2026-06-25
test(e2e): assert request-level completion in graceful shutdown tests

### LINKED ISSUE #510: bug: ProcessorGracefulShutdown test passes with failed requests
## Summary

`TestE2E/ProcessorGracefulShutdown` can pass even when the restarted processor loses work.

Both `PodDeleteMidJob` and `RollingRestartReEnqueue` currently accept a batch that reaches `status=completed`, but they do not assert that all requests were eventually completed successfully. In local e2e runs, the tests passed with non-zero failed counts, which means the current assertions allow false positives.

## Current behavior

The tests currently wait for either `completed` or `failed`, then only assert that the final batch status is `completed`.

As a result, runs like the following still pass:

- `PodDeleteMidJob`: `completed=4, failed=6, total=10`
- `RollingRestartReEnqueue`: `completed=8, failed=2, total=10`

This means the test is not yet proving the intended graceful shutdown guarantee.

## Why this matters

The purpose of these tests is to validate the processor's SIGTERM -> re-enqueue path.

If a restarted processor drops or permanently fails some in-flight requests, the batch may still end in `completed` while silently violating the expected recovery semantics. That makes the current test weaker than it looks and can mask real regressions.

## Expected behavior

If graceful shutdown and re-enqueue are working as intended, the test should verify a stronger invariant, such as:

- all requests are eventually accounted for, and
- no requests are unexpectedly lost, and
- the final success/failure distribution matches the intended shutdown semantics

Depending on the intended contract, that may mean asserting one of:

- `failed == 0`
- `completed == total`
- output/error artifacts contain the expected number of lines and statuses

## Proposed follow-up

Strengthen `TestE2E/ProcessorGracefulShutdown` so it validates request-level outcomes, not just final batch status.

Possible approaches:

1. Assert `completed == total` and `failed == 0` if graceful shutdown is expected to preserve all work.
2. If some failures are acceptable, document that contract explicitly and validate the exact expected result distribution.
3. Validate output/error files so the test checks artifact integrity in addition to batch status.

## Reproduction

Observed during local e2e runs:

- `TestE2E/ProcessorGracefulShutdown/PodDeleteMidJob` passed with `completed=4, failed=6`
- `TestE2E/ProcessorGracefulShutdown/RollingRestartReEnqueue` passed with `completed=8, failed=2`

## Related code

- `test/e2e/processor_graceful_shutdown_test.go`

madhugoutham:
## Why is this PR needed?

`TestE2E/ProcessorGracefulShutdown` (both `PodDeleteMidJob` and `RollingRestartReEnqueue`) only asserts that the batch reaches `completed` status. It does not check whether all requests actually succeeded. A batch can finish as `completed` with `failed=6, completed=4` and the test still passes, hiding silent request loss during the re-enqueue recovery path.

## What does this PR do?

Adds `RequestCounts.Completed == Total` assertions to both graceful shutdown tests, matching the pattern already used in `aimd_test.go` and `flow_control_test.go`.

## How was this tested?

`go vet` and `go build` clean. Assertion pattern matches existing tests (`aimd_test.go:98`, `flow_control_test.go:94`).

 ## Checklist

## Related Issues

 Fixes #510

### DIFF
--- a/test/e2e/processor_graceful_shutdown_test.go
+++ b/test/e2e/processor_graceful_shutdown_test.go
@@ -100,6 +100,12 @@ func doTestPodDeleteMidJob(t *testing.T) {
 	if finalBatch.Status != openai.BatchStatusCompleted {
 		t.Errorf("expected batch to complete after pod delete, got %s", finalBatch.Status)
 	}
+	if finalBatch.RequestCounts.Completed != finalBatch.RequestCounts.Total {
+		t.Errorf("expected all %d requests to complete after graceful shutdown, got completed=%d failed=%d",
+			finalBatch.RequestCounts.Total,
+			finalBatch.RequestCounts.Completed,
+			finalBatch.RequestCounts.Failed)
+	}
 }
 
 // doTestRollingRestartReEnqueue submits a batch with the same slow-request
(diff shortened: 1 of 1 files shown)


j-mok-dev (approved):
lgtm!

lioraron (approved):

PR #513, opened 2026-06-25
feat: baseline scenarios 0-2 with metrics, rate sweep, and multi-trial support

kahilam:
## Summary

* Add scenario 1 orchestrator support — launches `guidellm-batch` as a second guidellm instance sending batch-equivalent prompts as regular requests (demonstrates latency degradation without batch-gateway)
* Implement full metrics collection — parse guidellm CSVs for TTFT/ITL/TPOT at p50/p95/p99, add Prometheus query functions for GPU utilization and running/waiting requests
* Generate side-by-side HTML report — latency comparison table, TTFT/TPOT bar charts per scenario, batch timeline chart
* Emit structured `results.json` per run — inference-perf compatible schema with per-phase latency, throughput, error rates, and batch timelines
* Add pluggable data generation — `--data-source` flag supporting `faker` (default), `synthetic` (inference-perf style deterministic), and `sharegpt` (HuggingFace real conversations)
* Add rate sweep mode — `--rate-sweep` runs each scenario across a range of request rates with auto-calibration to discover saturating rate
* Add multi-trial support — `--trials N` runs each scenario N times and reports mean with 95% confidence intervals
* Add `--gpu-count`, `--max-model-len`, `--warmup` CLI flags for full configurability

## Part of #491 (PR 2 of 4)

Baseline scenarios and metrics infrastructure for comparing dispatch strategies.

## Test plan

* `python3 -c "import ast; ast.parse(open('benchmarks/benchmark.py').read())"` — compiles
* `python3 -c "import ast; ast.parse(open('benchmarks/generate_prompts.py').read())"` — compiles
* `make benchmark-local` — full e2e on Kind: setup, generate, benchmark, report, results.json
* New `results.json` emitted alongside report and metadata
* Warmup cycle labeling visible in output (`3 cycles (1 warmup)`)

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -74,6 +74,9 @@ class BenchmarkConfig:
     num_system_prompts: int
     results_dir: Path
     target: str
+    gpu_count: int = 1
+    max_model_len: int = 4096
+    warmup_cycles: int = 1
 
 
 @dataclass
@@ -86,8 +89,12 @@ class PhaseMetrics:
     itl_p50: float = 0.0
     itl_p95: float = 0.0
     itl_p99: float = 0.0
+    tpot_p50: float = 0.0
+    tpot_p95: float = 0.0
+    tpot_p99: float = 0.0
     req_latency_p50: float = 0.0
     req_latency_p95: float = 0.0
+    req_latency_p99: float = 0.0
     ok_rps: float = 0.0
     err_rps: float = 0.0
     error_rate: float = 0.0
@@ -350,22 +357,28 @@ def submit_batches(cfg, namespace):
 
 
 def start_interactive_traffic(cfg, namespace):
-    """Start guidellm burst/idle cycles."""
-    log(f"  Starting interactive traffic: {cfg.cycles} cycles, "
+    """Start guidellm burst/idle cycles with warmup support."""
+    log(f"  Starting interactive traffic: {cfg.cycles} cycles "
+        f"({cfg.warmup_cycles} warmup), "
         f"burst@{cfg.burst_rate}/s for {cfg.burst_seconds}s, "
         f"idle@{cfg.idle_rate}/s for {cfg.idle_seconds}s")
 
     cycle_lines = []
     for c in range(1, cfg.cycles + 1):
+        suffix = "warmup" if c <= cfg.warmup_cycles else ""
+        idle_output = f"idle-{c}-warmup.csv" if suffix else f"idle-{c}.csv"
+        burst_output = f"burst-{c}-warmup.csv" if suffix else f"burst-{c}.csv"
+        label = f" [WARMUP]" if suffix else ""
+
         cycle_lines.extend([
-            f'echo "=== Phase {c}: IDLE ({cfg.idle_rate} req/s, {cfg.idle_seconds}s) ==="',
+            f'echo "=== Phase {c}: IDLE ({cfg.idle_rate} req/s, {cfg.idle_seconds}s){label} ==="',
             f'guidellm benchmark run --target "$T" $COMMON '
             f'--profile constant --rate {cfg.idle_rate} --max-seconds {cfg.idle_seconds} '
-            f'--output-dir /results --outputs "idle-{c}.csv"',
-            f'echo "=== Phase {c}: BURST ({cfg.burst_rate} req/s, {cfg.burst_seconds}s) ==="',
+            f'--output-dir /results --outputs "{idle_output}"',
+            f'echo "=== Phase {c}: BURST ({cfg.burst_rate} req/s, {cfg.burst_seconds}s){label} ==="',
             f'guidellm benchmark run --target "$T" $COMMON '
             f'--profile constant --rate {cfg.burst_rate} --max-seconds {cfg.burst_seconds} '
-            f'--output-dir /results --outputs "burst-{c}.csv"',
+            f'--output-dir /results --outputs "{burst_output}"',
         ])
 
     script_lines = [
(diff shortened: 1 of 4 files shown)


benchmarks/benchmark.py
```diff
@@ -563,25 +641,217 @@ def collect_results(cfg, scenario, namespace):
...
+            return [0.0] * len(ps)
+        return [float(np.percentile(data, p)) for p in ps]
+
+    ttft_ps = percentiles(ttfts, [50, 95, 99])
+    itl_ps = percentiles(itls, [50, 95, 99])
+    tpot_ps = percentiles(tpots, [50, 95, 99])
+    lat_ps = percentiles(latencies, [50, 95, 99])
+
+    total = completed + errors
+    duration = max(latencies) - min(latencies) if len(latencies) > 1 else 1.0
```
lioraron: `duration = max(latencies) - min(latencies)` computes the range of per-request `total_latency` values, not wall-clock elapsed time. If all requests take similar time, this approaches 0 and inflates `ok_rps`/`err_rps` to infinity.

`total_latency` is how long each individual request took — the difference between the slowest and fastest request is not a meaningful denominator for throughput. You need wall-clock elapsed time (e.g., from timestamps in the CSV, or `sum(latencies)` as a rough proxy if requests are sequential).

benchmarks/benchmark.py
```diff
@@ -739,9 +1075,283 @@ def run_scenario(cfg, scenario):
...
+
+def auto_calibrate_rate(cfg, namespace):
+    """Discover the saturating rate with a short preliminary sweep.
+
+    Runs short bursts at increasing rates until error rate exceeds 5%
+    or TTFT p99 exceeds 3x the baseline (lowest rate).
+    """
+    log("  Auto-calibrating saturating rate...")
+    calibration_rates = [1, 5, 10, 15, 20, 25, 30, 40, 50]
+    baseline_ttft = None
```
lioraron, line 1273: `baseline_ttft` is assigned `None` here but never updated inside the loop. The TTFT-based saturation detection described in the docstring ("TTFT p99 exceeds 3x the baseline") is unimplemented — the loop never parses probe results, so the only working exit condition is pod failure status, which is a very crude signal. The function will almost always fall through to using the max probed rate (50 req/s).

Either implement the TTFT-based detection or remove the dead `baseline_ttft` variable and update the docstring to match what the function actually does.

benchmarks/benchmark.py
```diff
@@ -722,10 +1058,10 @@ def run_scenario(cfg, scenario):
     if scenario >= 2:
         timeline = monitor_scenario(cfg, scenario, namespace)
     else:
-        # Scenarios 0-1: just wait for guidellm to finish
+        # Scenarios 0-1: wait for guidellm job(s) to finish
         timeline = []
         total_wait = cfg.cycles * (cfg.burst_seconds + cfg.idle_seconds) + 120
-        log(f"  Waiting up to {total_wait}s for interactive traffic to complete...")
+        log(f"  Waiting up to {total_wait}s for traffic to complete...")
```
lioraron, line 1114: For scenario 1, two jobs run concurrently: `guidellm-interactive` and `guidellm-batch` (started by `start_batch_as_interactive_traffic`). This wait loop only checks `guidellm-interactive` — if the interactive job finishes first, `collect_results` runs while `guidellm-batch` is still writing to the PVC, and `batch-traffic.csv` may be missing or incomplete.

Add a wait for `guidellm-batch` completion when `scenario == 1`:
```python
if scenario == 1:
    kubectl(["wait", "--for=condition=complete", "job/guidellm-batch",
             f"--timeout={total_wait}s"], cfg.context, namespace, check=False)
```

benchmarks/generate_prompts.py
```diff
@@ -148,6 +169,78 @@ def generate_user_prompt(fake: Faker, target_chars: int) -> str:
...
+    """
+    try:
+        from datasets import load_dataset
+    except ImportError:
+        print("ERROR: 'datasets' package required for sharegpt source. "
+              "Install with: pip install datasets", file=sys.stderr)
+        sys.exit(1)
+
+    print("  Loading ShareGPT dataset from HuggingFace...", file=sys.stderr)
+    ds = load_dataset("anon8231489123/ShareGPT_Vicuna_unfiltered",
```
lioraron, line 203: This `load_dataset` call will fail at runtime. The `anon8231489123/ShareGPT_Vicuna_unfiltered` dataset stores its data as raw JSON files that don't follow HuggingFace's naming conventions for automatic split detection — there is no `"train"` split. See [this discussion](https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered/discussions/28).

Fix: specify `data_files` explicitly:
```python
ds = load_dataset("anon8231489123/ShareGPT_Vicuna_unfiltered",
                  data_files="ShareGPT_V3_unfiltered_cleaned_split.json",
                  split="train")
```
Or use the `Aeala/ShareGPT_Vicuna_unfiltered` mirror which has proper split configuration.

benchmarks/benchmark.py
```diff
@@ -739,9 +1075,283 @@ def run_scenario(cfg, scenario):
...
+                phase_groups[key] = []
+            phase_groups[key].append(phase)
+
+    aggregated_phases = []
+    for (phase_name, cycle), phases in sorted(phase_groups.items()):
+        n = len(phases)
+        if n == 0:
+            continue
+
+        def mean_and_ci(values):
```
lioraron: `mean_and_ci` is defined but never called — all means are computed inline with `sum(...) / n` below. `tpot_p99_vals` (line 1139) is also computed but never used. The CI metadata attached at the end (lines 1172-1182) recomputes variance from scratch instead of using this helper.

Either use `mean_and_ci` for all metric fields (and store CI for TPOT too, not just TTFT), or remove the dead code.

lioraron:
`make benchmark-local` should tear down the benchmark namespace by default — for a sanity test, fire-and-forget is the expected flow. Users who need to inspect the cluster can opt out.

Suggestion: add a `BENCHMARK_KEEP_CLUSTER` flag that skips teardown when set:

```makefile
benchmark-local:
	# ... existing steps ...
	@echo "Step 4/4: Done!"
	@echo "Report: $(BENCHMARK_RESULTS_DIR)/report.html"
	@if [ -z "$(BENCHMARK_KEEP_CLUSTER)" ]; then \
		echo "Tearing down benchmark namespace..."; \
		KUBE_CONTEXT=$(BENCHMARK_CONTEXT) SCENARIO=$(BENCHMARK_SCENARIO) bash benchmarks/teardown.sh; \
	else \
		echo "Keeping cluster (BENCHMARK_KEEP_CLUSTER set). Teardown with: make benchmark-local-teardown"; \
	fi
```

Usage: `make benchmark-local` tears down automatically, `make benchmark-local BENCHMARK_KEEP_CLUSTER=1` preserves the namespace for inspection.

benchmarks/generate_prompts.py
```diff
@@ -164,22 +257,37 @@ def generate_jsonl(
...
-            system_prompt = system_prompts[i % num_system_prompts]
-
-            # Sample ISL for this request (user prompt portion)
-            isl_tokens = sample_from_distribution(rng, isl_distribution, isl_mean, isl_stdev, min_val=16, max_val=isl_max)
-            target_chars = isl_tokens * 4
-            user_prompt = generate_user_prompt(fake, target_chars)
+            if data_source == "sharegpt":
+                system_prompt, user_prompt = sharegpt_data[i]
+            elif data_source == "synthetic":
+                system_prompt = system_prompts[i % num_system_prompts]
```
madhugoutham: nit: should we guard against num_system_prompts == 0 here? This would raise ZeroDivisionError.

kahilam: It is a simple fix — add a guard before the modulo.It would look like:

```
elif data_source == "synthetic":
    system_prompt = system_prompts[i % num_system_prompts] if num_system_prompts 
```

kahilam:
Hi @lioraron I addressed all you comments.

lioraron (approved):
Thanks @kahilam!

PR #523, opened 2026-06-29
fix: normalize PQ member to prevent duplicate re-enqueue

### LINKED ISSUE #293: [Bug]: Startup recovery can re-enqueue a job that is already in the queue, causing duplicate processing

### LINKED ISSUE #294: [Bug]: PQDelete removes all queue entries with the same SLO score, not just the target batch

madhugoutham:
## Why is this PR needed?

`PQEnqueue` marshals the full `BatchJobPriority` struct as the Redis sorted set member. Shutdown re-enqueue preserves the `Data` field from the original task, but recovery (`buildRecoveryTask`) and GC reconciler construct tasks without it. Redis `ZAddNX` deduplicates by exact byte string, so the two JSON representations are treated as distinct members. After a pod restart, both the shutdown and recovery paths enqueue the same job, two workers pick it up, and the job is processed twice.

Separately, `PQDelete` uses `ZRemRangeByScore(score, score)` which removes all members sharing that SLO microsecond value, not just the target job.

## What does this PR do?

- Normalize the sorted set member to `{id, slo}` only (new private `pqMember` struct), with SLO truncated to UTC microseconds so all 7 enqueue paths produce identical bytes for the same job.
- Switch `PQDelete` from `ZRemRangeByScore` to `ZRem` with exact normalized member bytes, targeting only the intended job.
- Move queue wait metric from `Data.CreatedAt` (queue member) to `jobInfo.BatchJob.CreatedAt` (database), since Data is no longer stored in the member.
- Remove `GetJobPriorityDataFromQueueItem` (no longer called).

Note: during a rolling upgrade, old-format members (with embedded Data) can persist up to 24h (CompletionWindow). Cancel of these entries falls to the slow path (mark as cancelling, wait for processor). Same risk level accepted for #294.

## How was this tested?

**Integration (real Redis):** 4 new tests dedup with/without Data, cancel-flow precision alignment (nanosecond vs microsecond SLO), same-SLO delete isolation (#294 regression), backward compat with old-format members.

**E2E (Kind):** Submitted batch, killed processor mid-job (SIGTERM), verified batch completed 5/5, zero duplicate queue entries, zero error logs, queue wait metric recorded from DB.

## Checklist

## Related Issues

Fixes #293, fixes #294


lioraron (approved):
Thanks!

Issue #528, opened 2026-06-30
enh: introduce a `JobRunner`


evacchi:
### Problem Statement

`executeJob` and `processModel` are methods on `Processor`, but all their meaningful state is per-job (progress tracking, output writers, input file, plan directory, cancel functions)

This coupling makes it hard to add per-job state without either growing the parameter lists further or stuffing more fields into the unstructured params bag.

### Proposed Solution

Introduce a `JobRunner` struct that owns the per-job state currently spread across `jobExecutionParams` and function parameters. Move `runJob, executeJob, processModel`, and the error/cancel handlers from `Processor` methods to `JobRunner` methods. Fold `jobExecutionParams` fields directly into `JobRunner`.

Shared `Processor` resources (semaphores, inference resolver, config) remain accessible via a `jr.processor` back-reference.

This is a structural refactor with no behavioral change.

### Willingness to Contribute

Yes, I can submit a PR

lioraron:
@evacchi My understanding is that this issue is superseded by https://github.com/llm-d/llm-d-batch-gateway/pull/580 and #581  that is now merged.
So closing this issue.
If the understanding is wrong, pls reopen.

PR #530, opened 2026-06-30
refactor: simplified `executionProgress` tracker

### LINKED ISSUE #527: enh: refactor `executionProgress`
### Problem Statement

Current progress tracking is tangled in the execution code — executionProgress lives in executor.go, uses atomics and CAS-based throttling, and is called directly from multiple goroutines. This makes it hard to reason about, test in isolation, or change independently from the dispatch logic.

### Proposed Solution

We introduce a centralized, asynchronous ProgressTracker that runs as a single goroutine consuming `ProgressEvent`s from a channel. Callers send events instead of calling atomic methods directly, so:

- No shared mutable state — counters are plain int64 fields, no atomics or CAS needed.
- Throttling is ticker-based — pushes to Redis every progressUpdateInterval when dirty, replacing the CAS race on lastUpdate. Semantically equivalent, simpler to follow.
- Independently testable — the tracker can be exercised with a mock StatusUpdater and a channel, without standing up the full execution pipeline.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

evacchi:
## Why is this PR needed?

Current progress tracking is tangled in the execution code and it uses different locking primitives. This won't scale if we eventually decouple request dispatching from response handling (needed for the async dispatcher, beneficial for synchronous calls too)

## What does this PR do?

The `executionProgress` methods are unchanged on the surface, but instead a channel is internally used for synchronization. This will scale better in the future if we centralize progress handling:

e.g. we could introduce some sort of `globalExecutionProgress` that receives all progress events and dispatches them to each per-job `executionProgress` tracker. This is however left for a follow-up.

## How was this tested?

All tests still pass.

## Checklist

## Related Issues

Closes #527, related #528

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -174,11 +174,20 @@ func (ep *executionProgress) run(ctx context.Context) {
 	defer ticker.Stop()
 	dirty := false
 
+	// pushCtx returns ctx if it is still active, otherwise context.Background()
+	// so that the final push after cancellation still reaches the status store.
+	pushCtx := func() context.Context {
+		if ctx.Err() != nil {
+			return context.Background()
+		}
+		return ctx
+	}
+
 	for {
 		select {
 		case ev, ok := <-ep.ch:
 			if !ok {
-				ep.push(ctx)
+				ep.push(pushCtx())
 				return
 			}
 			if ev.success {
@@ -190,27 +199,8 @@ func (ep *executionProgress) run(ctx context.Context) {
 
 		case <-ticker.C:
 			if dirty {
-				ep.push(ctx)
+				ep.push(pushCtx())
 				dirty = false
-			}
-
-		case <-ctx.Done():
-			for {
-				select {
-				case ev, ok := <-ep.ch:
-					if !ok {
-						ep.push(context.Background())
-						return
-					}
-					if ev.success {
-						ep.completed++
-					} else {
-						ep.failed++
-					}
-				default:
-					ep.push(context.Background())
-					return
-				}
 			}
 		}
 	}
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -195,6 +195,10 @@ func (ep *executionProgress) run(ctx context.Context) {
 			} else {
 				ep.failed++
 			}
+			if !dirty {
+				ep.push(pushCtx())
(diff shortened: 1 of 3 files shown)


internal/processor/worker/executor.go
```diff
@@ -96,68 +96,126 @@ func (o *outputLine) isSuccess() bool {
...
+// start launches the tracker goroutine.
+func (ep *executionProgress) start(ctx context.Context) {
+	go func() {
+		ep.run(ctx)
+		close(ep.done)
+	}()
+}
+
+// record sends a progress event to the tracker goroutine.
+func (ep *executionProgress) record(_ context.Context, success bool) {
```
j-mok-dev, line 152: This is now a blocking send into a bounded channel. Because `run()` also does `ep.push()` synchronously on the same goroutine, a slow `StatusSet` can stop `ep.ch` draining and make `record()` block once the buffer fills. That backpressures request completion/drain, whereas the previous atomic-counter version kept counting progress independently of Redis latency.

internal/processor/worker/executor.go
```diff
@@ -96,68 +96,126 @@ func (o *outputLine) isSuccess() bool {
...
+	// pushCtx returns ctx if it is still active, otherwise context.Background()
+	// so that the final push after cancellation still reaches the status store.
+	pushCtx := func() context.Context {
+		if ctx.Err() != nil {
+			return context.Background()
+		}
+		return ctx
+	}
+
+	for {
```
j-mok-dev, line 186: This changes the throttle semantics slightly: the first progress update is now delayed until the next tick, rather than being able to push immediately once outside the throttle window. That can add up to `progressUpdateInterval` of extra delay.

internal/processor/worker/executor.go
evacchi: well it is, but the channel limit is 1024, which should be pretty generous. StatusSet should be pretty fast (backed by Redis). We can however make the draining loop slightly tighter by special-casing the first record() call. will do.

internal/processor/worker/executor.go
evacchi: yes, as explained above we can add a special case for first progress update

evacchi:
upon further inspection, I decided this is not that significant an improvement over the current state, so I think we can close

PR #534, opened 2026-06-30
fix(otel): inject trace_id/span_id into logr context

### LINKED ISSUE #488: [Observability] Correlate logs with traces: inject trace_id/span_id into the logr logger
## Summary

The processor's tracing is detailed (cross-Redis context propagation, detached spans for shutdown re-enqueue, `otelhttp` outbound injection), and all logging is structured `logr`. But the two signals are not linked: **logs carry no `trace_id`/`span_id`**, so when debugging a failed batch you cannot pivot from a log line to its Jaeger trace, nor from a trace back to the surrounding logs.

A `grep` over `internal/`, `pkg/`, and `cmd/` finds zero `trace_id`/`span_id`/`traceID` references in non-test logging code, despite 124 `logr` call sites that pull the logger from `context`.

This aligns with llm-d-router [#1630 — "Logs are not correlated with traces: trace_id/span_id never injected into the logger"](https://github.com/llm-d/llm-d-router/issues/1630). The goal is to bring batch-gateway and router to observability parity.

## Why it matters

- **Trace → logs**: From a slow/errored span in Jaeger, an operator has the `trace_id` but no way to find the corresponding application logs.
- **Logs → trace**: From an error log line during incident triage, there is no `trace_id` to jump into the trace and see the full cross-component (apiserver → processor → inference) timeline.
- The plumbing already exists — spans are created through a single helper and loggers already flow through `context` — so this is a small, well-localized change with high debuggability payoff.

## Proposed change

When a span is started, enrich the logger stored in the returned `context` with the active span's IDs via `WithValues`, mirroring the router approach.

- `internal/util/otel/otel.go` — in `StartSpan` (and therefore `DetachedContext`, which calls it), after the span is created, derive `trace_id`/`span_id` from `span.SpanContext()` and attach them to the context logger:

```go
func StartSpan(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
    ctx, span := otel.Tracer(defaultServiceName).Start(ctx, name, opts...)
    if sc := span.SpanContext(); sc.IsValid() {
        log := logr.FromContextOrDiscard(ctx).WithValues(
            "trace_id", sc.TraceID().String(),
            "span_id", sc.SpanID().String(),
        )
        ctx = logr.NewContext(ctx, log)
    }
    return ctx, span
}
```

Because downstream code already retrieves the logger via `logr.FromContextOrDiscard(ctx)`, every log line emitted within a span's scope automatically gains the fields — no call-site changes required.

## Notes / scope

- When tracing is disabled (no `OTEL_EXPORTER_OTLP_ENDPOINT`), spans are non-recording and `SpanContext()` is invalid, so the `IsValid()` guard makes this a no-op — no spurious empty fields in logs.
- Field naming should match router (`trace_id` / `span_id`) so dashboards and log queries are portable across both projects.

## Acceptance criteria

- [ ] Log lines emitted within a span carry `trace_id` and `span_id` fields.
- [ ] No fields are added when tracing is disabled.
- [ ] Field names align with llm-d-router.
- [ ] A short note in `docs/guides/metrics.md` (or a tracing guide) documents the log↔trace correlation fields.

madhugoutham:
## Why is this PR needed?

Logs and traces are disconnected. The batch-gateway creates 19 OTel spans and writes 98 logr call sites, but no log line carries a trace_id or span_id. During incidents, operators cannot pivot from a log line to its Jaeger trace or from a trace back to the surrounding logs.

## What does this PR do?

Enriches the logr logger with trace_id and span_id inside StartSpan(). Because all downstream code retrieves the logger via logr.FromContextOrDiscard(ctx), every log line emitted within a span automatically gains trace correlation fields. No call-site changes needed.

When tracing is disabled (no OTEL_EXPORTER_OTLP_ENDPOINT), SpanContext is invalid and the IsValid() guard skips enrichment. Logs stay clean.

DetachedContext also benefits because it calls StartSpan internally.

To prevent duplicate trace_id/span_id keys on nested spans (Zap appends without dedup), a base logger is captured before the first enrichment. Subsequent StartSpan calls enrich from that base rather than the accumulated context logger

## How was this tested?

## Checklist

## Related Issues

Closes #488
Aligns with llm-d-router#1630


lioraron (approved):

internal/util/otel/otel.go
```diff
@@ -49,9 +49,31 @@ const (
...
+// Nested StartSpan calls enrich from this base rather than from the
+// accumulated context logger, preventing duplicate trace_id/span_id fields.
+type baseLoggerKey struct{}
+
 // StartSpan creates a new span using the batch-gateway tracer.
+// When the span carries a valid trace context, the logger in the returned
+// context is enriched with trace_id and span_id so that all downstream
+// log lines emitted via logr.FromContextOrDiscard(ctx) are automatically
+// correlated with the active trace.
 func StartSpan(ctx context.Context, name string, opts ...trace.SpanStartOption) (context.Context, trace.Span) {
```
yizhaodev, line 62: should we also provide a  `EndSpan` helper that restores the base logger?
and replace span.End() with EndSpan?
```
  func EndSpan(ctx context.Context, span trace.Span) context.Context {
      span.End()
      // restore logger
      return ctx
  }
```

madhugoutham: No call site logs after ending a span, so there's nothing for the helper to restore.

PR #525, opened 2026-06-30
fix(benchmarks): GPU cluster compatibility for end-to-end runs

kahilam:
## Summary

- Fix silent ConfigMap creation failure in `benchmark.py`: `kubectl create configmap` does not support the `-l` flag — split into create + label commands
- Disable `networkPolicy` and `PodDisruptionBudget` for Redis/PostgreSQL Helm installs for RBAC compatibility on shared CKS clusters
- Update Router local-repo mode (`ROUTER_REPO`) to use new `router.*` chart values format (llm-d-router v0.9.0+), removing `LLM_D_REPO` dependency and forcing fresh `helm dependency build`

### Prerequisites

- Access to a CKS cluster with GPU nodes (e.g., `coreweave-waldorf` context)
- A namespace with RBAC permissions for ServiceAccounts, NetworkPolicies, PodDisruptionBudgets
- `kubectl`, `helm`, `gh` CLI tools installed
- GitHub PAT with `read:packages` scope for GHCR access
- vLLM already deployed and serving a model (e.g., Qwen3-8B on 2×A100)

### Steps

**1. Clone the Router repo locally (required until OCI chart v0.9.2 is published):**

```bash
git clone https://github.com/llm-d/llm-d-router.git /tmp/llm-d-router
```

**2. Deploy scenario 3 infrastructure (Redis, PostgreSQL, Router, Batch Gateway):**

```bash
cd llm-d-batch-gateway

BENCHMARK_SCENARIO=3 \
MODE=gpu \
NAMESPACE=kmokhtar-dev \
KUBE_CONTEXT=coreweave-waldorf \
ROUTER_REPO=/tmp/llm-d-router \
GHCR_USER=<github-username> \
GHCR_TOKEN=<github-pat> \
bash benchmarks/setup.sh
```

**3. Set up Prometheus port-forward (for metrics collection):**

```bash
kubectl --context coreweave-waldorf -n llm-d-monitoring \
  port-forward svc/llmd-kube-prometheus-stack-prometheus 9090:9090
```

**4. Create a ServiceMonitor for the batch-gateway processor (if not already present):**

```bash
kubectl --context coreweave-waldorf -n kmokhtar-dev apply -f - <<'EOF2'
apiVersion: v1
kind: Service
metadata:
  name: batch-gateway-processor-metrics
  labels:
    app: batch-gateway-processor
spec:
  selector:
    app.kubernetes.io/component: processor
    app.kubernetes.io/instance: batch-gateway
  ports:
    - name: metrics
[...]

PR #535, opened 2026-06-30
feat(benchmarks): GPU benchmark automation (make benchmark-gpu)

kahilam:
## Summary

Reduce the GPU cluster benchmark procedure to a single `make` target that runs all scenarios end-to-end (setup → benchmark → teardown per scenario).

- **ServiceMonitor automation**: `setup.sh` (GPU mode, scenarios ≥ 3) creates the processor metrics Service and ServiceMonitor automatically. Discovery label parameterized via `PROMETHEUS_RELEASE` env var (default: `llmd-kube-prometheus-stack`)
- **Port-forward automation**: `benchmark.py` manages a background `kubectl port-forward` to Prometheus internally. Accepts `--prometheus-namespace` and `--prometheus-service` flags. Falls back to `PROMETHEUS_URL` env var for clusters with direct Prometheus access
- **`make benchmark-gpu` target**: Wraps `setup.sh` + `benchmark.py` + `teardown.sh` for GPU cluster runs across all scenarios (0-4), analogous to `make benchmark-local` for Kind
- **EPP image tag parameterization**: Already landed in main via PR #525 (`ROUTER_EPP_TAG`)

## Usage

```bash
# Run all scenarios (0-4) on a GPU cluster
BENCHMARK_GPU_CONTEXT=coreweave-waldorf \
BENCHMARK_GPU_NAMESPACE=kmokhtar-dev \
GHCR_USER=kahilam GHCR_TOKEN="$GH_TOKEN" \
ROUTER_REPO=/tmp/llm-d-router \
make benchmark-gpu

# Run specific scenarios
BENCHMARK_GPU_SCENARIOS="3 4" \
BENCHMARK_GPU_CONTEXT=coreweave-waldorf \
GHCR_USER=kahilam GHCR_TOKEN="$GH_TOKEN" \
make benchmark-gpu
```

## Test plan

- [ ] `make benchmark-gpu` runs scenarios 0-4 sequentially on CKS GPU cluster
- [ ] ServiceMonitor created automatically (no manual `kubectl apply` needed)
- [ ] Prometheus port-forward started/stopped automatically by benchmark.py
- [ ] AIMD charts populated in report for scenarios 3-4
- [ ] `PROMETHEUS_URL` env var still works as fallback (backward compatible)
- [ ] CI passes (`make pre-commit`)

Closes the "PR 5: GPU benchmark automation" item in #491.


kahilam:
Hi @lioraron Ran all 5 scenarios end-to-end on CoreWeave GPU cluster (Qwen3-8B, 1x GPU) with make benchmark-gpu BENCHMARK_GPU_SCENARIOS="0 1 2 3 4" BENCHMARK_GPU_BATCH_SIZE=100. All scenarios passed, HTML reports generated with latency charts, batch timelines, AIMD dynamics, and flow control saturation.

With 100 requests/job and 15 req/s interactive burst, the GPU handles the load comfortably — AIMD stays flat at the configured limit (no backoff triggered). To show AIMD sawtooth dynamics, we'd need higher load (500+ batch requests or 30+ req/s burst).

lioraron:
The current approach calls `benchmark.py --scenarios $s` separately per scenario from the Makefile loop. This means each invocation only sees one scenario's results, so the cross-scenario comparison report (summary table with all scenarios side-by-side, batch timeline overlay, latency bar charts) is never generated — each scenario gets its own isolated `report.html`.

`benchmark.py` already has `generate_html_report(cfg, results)` that produces the cross-scenario report when it receives multiple `ScenarioResult` objects — but that only works when all scenarios run within a single `benchmark.py` invocation.

Suggested approach: add a `--managed` flag to `benchmark.py` that makes it call `setup.sh` before and teardown after each scenario within its own loop. This keeps all results in a single process, so the existing cross-scenario report generation works naturally without needing a new report-merging mode.

The Makefile target would then simplify to a single `benchmark.py --managed --scenarios 0 2 3 4` invocation instead of the shell loop.

I've updated the PR 5 spec in #491 to reflect this.

lioraron:
Two additional requirements for this PR based on spec discussion in #491:

**1. Full summary table in the HTML report**

`generate_html_report` needs to produce the one-row-per-scenario summary table specified in #491's Report section. The current report has a per-phase breakdown but is missing the top-level summary with:
- Interactive TTFT p99 (burst)
- Interactive ITL p99 (burst)
- Batch idle throughput (req/s)
- Batch SLO completion (actual vs target per job)
- Batch TTFT p50

All of this data is already collected or derivable from existing `ScenarioResult` fields — this is a presentation change in the HTML template, not a new metrics collection task.

**2. Managed orchestration (per earlier comment)**

As noted in my [previous comment](https://github.com/llm-d/llm-d-batch-gateway/pull/535#issuecomment-4870048950), `benchmark.py` should orchestrate setup/teardown per scenario internally (via a `--managed` flag or equivalent) rather than the Makefile shell loop calling `benchmark.py` per scenario separately. This is needed to produce the cross-scenario comparison report — without it, each scenario gets its own isolated `report.html` and the summary table above can't compare across scenarios.

The spec in #491 has been updated to reflect both of these (this PR is now "PR 4: GPU benchmark automation + summary table").

lioraron:
One more requirement: the batch JSONL input file is currently stored in a ConfigMap (`{name}-data`), which has a hard 1MB Kubernetes limit. With ISL ≥ 256 tokens (needed to stress TTFT so prefill takes real time and queuing is visible) and 1000 requests per job, the JSONL exceeds this limit.

Fix: replace the ConfigMap-based delivery with PVC-based storage — write the JSONL to a PersistentVolumeClaim (e.g., the existing `benchmark-results` PVC) and mount it into the batch-submit pod.

Updated in #491 under PR 4.

benchmarks/benchmark.py
```diff
@@ -416,7 +426,7 @@ def start_interactive_traffic(cfg, namespace):
       restartPolicy: Never
       containers:
         - name: guidellm
-          image: ghcr.io/vllm-project/guidellm:latest
+          image: ghcr.io/vllm-project/guidellm:v0.6.0
```
lioraron: This pins guidellm to `v0.6.0`, but the manifests on `main` (merged in #546) pin to `v0.6.1`. Should be consistent — change both occurrences (here and line 494) to `v0.6.1`.

`v0.6.0` has a known issue with `--disable-console-interactive` that was fixed in `v0.6.1`.

benchmarks/benchmark.py
```diff
@@ -779,12 +789,83 @@ def parse_scenario_results(results_dir):
...
+            log(f"  WARNING: Prometheus port-forward failed to start (exit={self._process.returncode})")
+            self._process = None
+            return False
+        log(f"  Prometheus port-forward started (localhost:{self.local_port} → {self.service})")
+        return True
+
+    def stop(self):
+        if self._process:
+            self._process.terminate()
+            self._process.wait(timeout=5)
```
lioraron: If the port-forward process ignores `SIGTERM`, `wait(timeout=5)` raises `subprocess.TimeoutExpired` and the process leaks. Add a `kill()` fallback:

```python
def stop(self):
    if self._process:
        self._process.terminate()
        try:
            self._process.wait(timeout=5)
        except subprocess.TimeoutExpired:
            self._process.kill()
            self._process.wait()
        self._process = None
```

lioraron:
The hardcoded password `benchmarkpw` in `benchmarks/setup.sh` was flagged by the downstream security scanner ([opendatahub-io/batch-gateway secret-scanning/3](https://github.com/opendatahub-io/batch-gateway/security/secret-scanning/3)).

Two occurrences:
- **Line 100**: `--set auth.password=benchmarkpw`
- **Line 109**: `postgresql://postgres:benchmarkpw@postgresql...`

Please replace with a variable (e.g. `BENCH_DB_PASSWORD`) defaulting to a placeholder, so the password isn't embedded as a string literal in the script.

lioraron:
The DCO check is failing — the last 3 commits are missing `Signed-off-by`:

- `3ab2b42` fix(benchmarks): GPU cluster compatibility fixes for end-to-end runs
- `5ba9779` fix(benchmarks): address PR review comments
- `664926f` fix(benchmarks): remove hardcoded PostgreSQL password from setup.sh

You can fix all of them with:

```bash
git rebase --signoff HEAD~3
git push --force-with-lease
```

kahilam:
Hi @lioraron lioraron This PR now covers both PR 4 (GPU benchmark automation) and PR 5 (Report completeness) from #491.

benchmarks/benchmark.py
```diff
@@ -1209,6 +1723,69 @@ def generate_html_report(cfg, results):
...
+        env=env,
+        check=False,
+    )
+    if result.returncode != 0:
+        log(f"  [managed] WARNING: setup.sh exited with code {result.returncode}")
+
+
+def _managed_teardown(args):
+    """Tear down all benchmark resources in the namespace."""
+    namespace = _NAMESPACE_OVERRIDE or "batch-bench"
```
lioraron: `_managed_teardown` falls back to `"batch-bench"` but `_managed_setup` uses `f"batch-bench-s{scenario}"`. Without `--namespace`, teardown targets a non-existent namespace and all resources leak.

The Makefile flow always passes `--namespace` so it works, but the standalone CLI is broken:
```
python3 benchmark.py --managed --context X --scenarios 0 2
# setup creates batch-bench-s0, batch-bench-s2
# teardown tries to clean "batch-bench" → no-op
```

Fix: accept `scenario` as a parameter (matching `_managed_setup`'s signature) and use the same `f"batch-bench-s{scenario}"` fallback.

benchmarks/benchmark.py
```diff
@@ -1312,9 +1903,20 @@ def run_scenario(cfg, scenario):
...
             log(f"  Flow control: saturation avg={flow_control_metrics.get('flow_control_saturation_avg', 0):.2f}")
 
+    # Collect GPU/infrastructure metrics for all scenarios
+    gpu_metrics = {}
+    if os.environ.get("PROMETHEUS_URL"):
+        end_time = datetime.datetime.utcfromtimestamp(time.time())
+        start_time = end_time - datetime.timedelta(seconds=cfg.cycles * (cfg.burst_seconds + cfg.idle_seconds))
+        gpu_metrics = collect_gpu_metrics(cfg.context, namespace, start_time, end_time)
+        if gpu_metrics:
+            log(f"  GPU: cache_usage avg={gpu_metrics.get('gpu_cache_usage_avg', 0):.1f}%, "
```
lioraron: `vllm:gpu_cache_usage_perc` returns a 0–1 ratio, but the log formats it without multiplying by 100. A cache usage of 85% logs as `0.8%` instead of `85.0%`.

The HTML chart correctly does `d.cache_avg * 100` — this is just the log line.

```python
log(f"  GPU: cache_usage avg={gpu_metrics.get('gpu_cache_usage_avg', 0) * 100:.1f}%, "
```

benchmarks/benchmark.py
```diff
@@ -1209,6 +1723,69 @@ def generate_html_report(cfg, results):
...
+    namespace = _NAMESPACE_OVERRIDE or f"batch-bench-s{scenario}"
+    ghcr_user = getattr(args, "ghcr_user", None) or os.environ.get("GHCR_USER", "")
+    ghcr_token = getattr(args, "ghcr_token", None) or os.environ.get("GHCR_TOKEN", "")
+    router_repo = getattr(args, "router_repo", None) or os.environ.get("ROUTER_REPO", "")
+    prometheus_release = (
+        getattr(args, "prometheus_release", None)
+        or os.environ.get("PROMETHEUS_RELEASE", "")
+    )
+
+    env = {
```
lioraron, line 1745: `PROMETHEUS_NAMESPACE` is missing from the env dict, even though `PROMETHEUS_RELEASE` is explicitly forwarded. If the user passes `--prometheus-namespace monitoring`, setup.sh won't see it — it'll use its own default `llm-d-monitoring`.

Add it alongside `PROMETHEUS_RELEASE`:
```python
"PROMETHEUS_NAMESPACE": args.prometheus_namespace,
```

benchmarks/setup.sh
```diff
@@ -308,17 +328,19 @@ else
...
+        if [ "${SCENARIO}" = "4" ]; then
+            FLOW_CONTROL_OVERLAY="-f ${SCRIPT_DIR}/helm-values/scenario-4-flow-control-overlay-router.yaml"
+        fi
         # Local repo mode (development override)
         log "  Using local repo: ROUTER_REPO=${ROUTER_REPO}"
         chart_dir="${ROUTER_REPO}/config/charts/llm-d-router-gateway"
         rm -f "${chart_dir}/Chart.lock"
         (cd "${chart_dir}" && helm dependency build >/dev/null 2>&1)
-        ${H} install "${GUIDE_NAME}" "${chart_dir}" \
+        ${H} upgrade --install "${GUIDE_NAME}" "${chart_dir}" \
```
lioraron, line 343: This was correctly changed to `upgrade --install`, but the OCI-mode path at line 376 still uses plain `helm install`. In `--managed` mode, if `_managed_teardown` fails to fully clean up the router release, the next setup.sh invocation will fail with `cannot re-use a name that is still in use`.

Apply the same change to line 376.

benchmarks/benchmark.py
```diff
@@ -272,17 +285,71 @@ def _batch_submit_script():
...
+              mountPath: /data
+      volumes:
+        - name: data
+          persistentVolumeClaim:
+            claimName: benchmark-results
+    """)
+    kubectl(["delete", "pod", helper_name, "--ignore-not-found", "--wait"],
+            cfg.context, namespace, check=False)
+    kubectl_apply(helper_yaml, cfg.context, namespace)
+    kubectl(["wait", "pod", helper_name, "--for=condition=Ready",
```
lioraron: If `kubectl wait` times out or `kubectl cp` fails on the third retry, `CalledProcessError` propagates and the `kubectl delete pod data-loader` at the end of this function is never reached — the busybox helper pod leaks (sleeping for 300s).

Wrap the wait-through-delete section in try/finally:
```python
try:
    kubectl(["wait", ...], ...)
    time.sleep(2)
    for name in job_names:
        ...
finally:
    kubectl(["delete", "pod", helper_name, "--ignore-not-found"], ...)
```

benchmarks/benchmark.py
```diff
@@ -1918,6 +2561,8 @@ def main():
     log(f"Metadata: {metadata_path}")
     log(f"Data:     {results_json_path}")
 
+    stop_prometheus_port_forward()
```
lioraron, line 2568: If any scenario raises or the user hits Ctrl+C, `stop_prometheus_port_forward()` is never called and the kubectl subprocess holds port 9090 open. The next run fails with `address already in use`.

Wrap the benchmark loop + cleanup in try/finally, or register `atexit.register(stop_prometheus_port_forward)` after starting it.

lioraron (approved):

PR #538, opened 2026-07-01
refactor: introduce result collector + improve error handling in `executeJob`

### LINKED ISSUE #537: enh: refactor job output writer + progress tracking
### Problem Statement

output writing and progress tracking are currently separate processes, but they usually go hand-in-hand; normally, as soon as we need to write back a result, we need also to track the progress

### Proposed Solution

we introduce a "result collector" component that updates the status of the execution + writes output. This in the future will allow to further decouple progress tracking and result collection from sending out requests for asynchronous processing.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

evacchi:
## Why is this PR needed?

introduce a result collector that both logs errors/status and updates job progress status; in the future we should be able to decouple sending a request from collecting the results for asynchronous processing.

## What does this PR do?

it refactors the outputWriter to resultCollector that _also_ handles (by delegating to the `executionProgress`) status updates.

## How was this tested?

## Checklist

## Related Issues

closes https://github.com/llm-d/llm-d-batch-gateway/issues/537, related #530, #529

### DIFF
--- a/internal/processor/worker/collector.go
+++ b/internal/processor/worker/collector.go
@@ -1,0 +1,162 @@
+/*
+Copyright 2026 The llm-d Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package worker
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"sync"
+
+	"github.com/go-logr/logr"
+
+	batch_types "github.com/llm-d/llm-d-batch-gateway/internal/shared/types"
+)
+
+// ResultItem is the outcome of a single inference request.
+// Produced by processModel goroutines, consumed by resultCollector.
+type ResultItem struct {
+	RequestID        string
+	CustomID         string
+	Response         *batch_types.ResponseData
+	Error            *OutputError
+	HadCapacityRetry bool
+	ModelID          string
+}
+
+// OutputError is the error structure written to JSONL output/error files.
+type OutputError struct {
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
+func (r *ResultItem) isSuccess() bool {
+	return r.Error == nil && r.Response != nil && r.Response.StatusCode == 200
+}
+
+func resultToOutputLine(r *ResultItem) *outputLine {
+	var outErr *outputError
+	if r.Error != nil {
+		outErr = &outputError{Code: r.Error.Code, Message: r.Error.Message}
+	}
+	return &outputLine{
+		ID:       r.RequestID,
+		CustomID: r.CustomID,
+		Response: r.Response,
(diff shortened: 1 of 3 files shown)


internal/processor/worker/executor.go
```diff
@@ -266,111 +247,72 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
...
 	for safeModelID, modelID := range modelMap.SafeToModel {
-		// Ordering guarantee: processModel returns → requestAbortFn → errCh send.
-		// This ensures the first real error reaches errCh before any context.Canceled
-		// from other models whose contexts were cancelled by requestAbortFn.
-		go func(safeModelID, modelID string) {
-			err := p.processModel(
+		go func() {
+			errCh <- p.processModel(
 				requestAbortCtx,
 				ctx,
```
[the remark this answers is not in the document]
evacchi: no longer true since go 1.22 -- this is fine and more readable.

internal/processor/worker/collector.go
```diff
@@ -0,0 +1,162 @@
...
+			continue
+		}
+		lineBytes = append(lineBytes, '\n')
+
+		isError := line.Error != nil
+		writer := c.outputWriter
+		if isError {
+			writer = c.errorWriter
+		}
+		if _, err := writer.Write(lineBytes); err != nil {
```
j-mok-dev, line 151: This aborts sibling dispatch, but it still swallows the actual output write failure. Because the collector does not surface this error back to `executeJob()`, a local output I/O failure here can still lead to a successful return with partial/truncated results.

evacchi: slightly modified now so that a write error is actually surfaced

internal/processor/worker/collector.go
```diff
@@ -0,0 +1,169 @@
...
+			c.logger.Error(err, "Failed to write output line", "customId", result.CustomID)
+			c.abort(err)
+			continue
+		}
+
+		c.progress.record(ctx, result.isSuccess())
```
j-mok-dev, line 157: This introduces a new coupling between result draining and progress publication. resultCollector is a single goroutine, so if progress.record() blocks on the status store, the collector stops draining results, collect() eventually backpressures model goroutines, and executeJob() can hang even though output writes are otherwise fine. In the previous design, progress updates were best-effort work done by worker goroutines, so a slow status-store call could stall one updater but not the entire result pipeline.

internal/processor/worker/collector.go
```diff
@@ -0,0 +1,169 @@
...
+	abortOnce    sync.Once
+
+	ch       chan *ResultItem
+	done     chan error
+	writeErr error
+}
+
+func newResultCollector(outputWriter, errorWriter *bufio.Writer, progress *executionProgress, logger logr.Logger, abortFn context.CancelFunc) *resultCollector {
+	if abortFn == nil {
+		panic("resultCollector: abortFn cannot be nil")
```
j-mok-dev, line 93: nit: I think it is better to avoid `panic` here. Our usual convention is to return an error (or treat nil as a no-op) rather than panic in production paths.
I noticed `executeJob` already substitutes a no-op when `requestAbortFn` is nil, so this may be unreachable today, but if a future caller hits this constructor directly, a panic would take down
the process. Returning an error (or documenting/handling nil explicitly) would make that failure mode safer.

evacchi:
this can also be closed as superseded by work in #580

lioraron:
@evacchi Is this superseded by #581 which is merged?
If so pls close this?

PR #550, opened 2026-07-03
fix: auto-skip sync e2e tests under async dispatch and harden async dispatcher

### LINKED ISSUE #547: fix: async dispatcher hardening — panic recovery, buffer config, deadline config, type assertion
### Context

Follow-up items from PR #458 review that were partially fixed or deferred.

### Items

**1. Panic recovery — `sync.Once` prevents dispatcher restart**

If `resultDispatcher.run()` panics (e.g., from a library bug or unexpected nil), `sync.Once` has already fired and `ensureStarted()` can never re-launch the goroutine. All subsequent `GetResult` calls hang indefinitely.

Fix: add `recover()` in `run()` that resets the `sync.Once` and re-launches, or replace `sync.Once` with a mutex-guarded started flag.

`pkg/clients/inference/async_inference_client_impl.go` (lines 48, 60-67)

**2. Bare type assertion violates project convention**

`key.(string)` in the Close/unregister path is a bare assertion (panics on failure). CLAUDE.md requires comma-ok form `v, ok := x.(T)`.

```go
c.pendingIDs.Range(func(key, _ any) bool {
    c.pool.dispatcher.unregister(key.(string))  // bare assertion
    return true
})
```

Fix: use `if k, ok := key.(string); ok { ... }`.

`pkg/clients/inference/async_inference_client_impl.go` (line ~216)

### LINKED ISSUE #549: Integration tests failed on 2026-07-03
## Root Cause

The dispatcher CI matrix entry (added in #458) reconfigures the **single processor deployment** from sync to async dispatch mode (`dispatchMode: async` via `processor-async-values.yaml`). It then runs `make test-e2e`, which executes **all** e2e tests (`go test ./...`), not just `TestDispatcher*`.

The standard e2e tests (`TestE2E`) assume sync dispatch — requests are forwarded directly to the vLLM sim and results are returned inline. Under async dispatch, requests are submitted to a Redis queue and results must be collected via `resultDispatcher`. This breaks the sync-path assumptions in multiple test groups.

## Failed Tests (9 failures, all in `TestE2E` — all `TestDispatcher*` tests passed)

| Test | Duration | Failure Mode |
|------|----------|-------------|
| `Batches/Cancel/TerminalBatchRejected` | 300s (timeout) | Batch stuck `in_progress` — requests submitted to async queue but never complete in time |
| `Batches/Expiration` | 12s | Requests get `batch_expired` / `result not collected before deadline` errors |
| `Batches/MultiModel` | 300s (timeout) | Batch stuck `in_progress` (completed=0, failed=1) — one model fails, rest never complete |
| `ProcessorGracefulShutdown/PodDeleteMidJob` | 23s | Batch reaches `completed` with 0 completed + 10 failed (status="completed" instead of expected "failed") |
| `ProcessorGracefulShutdown/RollingRestartOrphan` | 18s | Same: 0 completed + 10 failed, wrong terminal status |
| `OrphanRecovery/HardCrashInProgress` | 18s | Same pattern: 0 completed + all failed |
| `OrphanRecovery/CancellingOrphan` | 126s | `did not reach 1 completed requests within 2m0s` |
| `FlowControl/HeaderAndRetry/InferenceObjectiveHeader` | 72s | `inferenceObjective not found for model "sim-model" in ConfigMap` — async values overlay replaces modelGateways config, removing flow control settings. Batch never completes |
| `FlowControl/HeaderAndRetry/SLOHeader` | 0.05s | Nil pointer panic in `waitForBatchStatus` at `helpers_test.go:247` — cascading from InferenceObjectiveHeader failure |

## Why the Failures Happen

The `dev-deploy-dispatcher.sh` script runs:
```bash
helm upgrade batch-gateway ... --reset-values --values <current-minus-modelGateways> --values processor-async-values.yaml
```

This replaces the processor's `modelGateways` config with an async-only version that only has `sim-model` with `inferencePoolName: "sim-pool"` — no `inferenceObjective`, no flow control headers. The processor then routes all inference through the async dispatch path (submit to Redis, collect from `resultDispatcher`), which the standard tests don't account for.

## Fix

Add a `detectDispatcherDeployed(t)` skip guard to `TestE2E`, consistent with the existing auto-detection pattern used throughout the e2e suite (e.g., `detectGIEDeployed` for GIE tests). When the dispatcher matrix deploys async-processors and reconfigures the processor to async mode, `TestE2E` auto-skips itself — same way `TestDispatcher` auto-skips when no dispatcher is deployed.

This keeps `make test-e2e` fully auto-detecting with no need for `TEST_RUN` filters in CI.

## Context

- CI run: https://github.com/llm-d/llm-d-batch-gateway/actions/runs/28673930308
- Only the dispatcher matrix failed — all other 6 matrices passed
- All `TestDispatcher*` and `TestDispatcherOTelTraces` tests passed — the async dispatch code works correctly
- Introduced by #458 (e2e dispatcher integration)

[Original run logs](https://github.com/llm-d/llm-d-batch-gateway/actions/runs/28673930308)

lioraron:
## Summary

- **#549**: Add `detectDispatcherDeployed(t)` skip guard to `TestE2E` so sync-path e2e tests auto-skip when the processor is in async dispatch mode. Consistent with the existing auto-detection pattern (e.g., `detectGIEDeployed`). No CI workflow changes needed — `make test-e2e` remains fully self-detecting.
- **#547**: Harden the async dispatcher:
  - Add panic recovery in `resultDispatcher` via `processNext()` with deferred `recover()`, keeping the goroutine alive after panics (sync.Once would otherwise prevent restart)
  - Fix bare type assertion `key.(string)` in `asyncProducerClient.Close()` → comma-ok form
  - Wire `ResultPollTimeout` from `AsyncDispatchConfig` through to the result dispatcher (previously disconnected from config)

`DefaultDeadline` and `ResultBufferSize` are kept as internal constants — they are implementation details with sensible defaults, not values operators need to tune.

Closes #549
Closes #547

## Test plan

- [x] `make pre-commit` — all checks pass (fmt, vet, lint, unit tests, security scan)
- [x] `make test-regression` — API schema compatibility passes
- [x] `make test-integration` — integration tests pass
- [x] `ENABLE_DISPATCHER=true make dev-deploy` then `make test-e2e` — verify `TestDispatcher*` runs and `TestE2E` auto-skips
- [x] `make dev-deploy` (no dispatcher) then `make test-e2e` — verify `TestE2E` runs and `TestDispatcher*` auto-skips

### DIFF
--- a/pkg/clients/inference/async_inference_client_resolver.go
+++ b/pkg/clients/inference/async_inference_client_resolver.go
@@ -31,9 +31,11 @@ const asyncQueuePrefix = "llm-d-async:"
 
 // AsyncClientConfig holds the resolved configuration for async dispatch.
 type AsyncClientConfig struct {
-	RedisURL        string
-	Models          map[string]string // model name -> pool name
-	DefaultDeadline time.Duration     // fallback deadline when ctx has none; 0 defaults to 5m
+	RedisURL          string
+	Models            map[string]string // model name -> pool name
+	DefaultDeadline   time.Duration     // fallback deadline when ctx has none
+	ResultPollTimeout time.Duration     // per-poll timeout in the result dispatcher loop
+	ResultBufferSize  int               // per-job channel capacity for async results
 }
 
 // AsyncGatewayResolver routes models to per-job AsyncInferenceClient instances.
@@ -117,8 +119,14 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		}
 
 		poolLogger := logger.WithName("async-inference").WithValues("pool", poolName)
-		d := newResultDispatcher(p, poolLogger)
-		pool := &asyncPool{producer: p, dispatcher: d, logger: poolLogger, defaultDeadline: config.DefaultDeadline}
+		d := newResultDispatcher(p, poolLogger, config.ResultPollTimeout)
+		pool := &asyncPool{
+			producer:         p,
+			dispatcher:       d,
+			logger:           poolLogger,
+			defaultDeadline:  config.DefaultDeadline,
+			resultBufferSize: config.ResultBufferSize,
+		}
 
 		pools[model] = pool
 		closers = append(closers, d, p)
--- a/pkg/clients/inference/async_inference_client_resolver.go
+++ b/pkg/clients/inference/async_inference_client_resolver.go
@@ -33,9 +33,7 @@ const asyncQueuePrefix = "llm-d-async:"
 type AsyncClientConfig struct {
 	RedisURL          string
 	Models            map[string]string // model name -> pool name
-	DefaultDeadline   time.Duration     // fallback deadline when ctx has none
 	ResultPollTimeout time.Duration     // per-poll timeout in the result dispatcher loop
-	ResultBufferSize  int               // per-job channel capacity for async results
 }
 
 // AsyncGatewayResolver routes models to per-job AsyncInferenceClient instances.
@@ -121,11 +119,9 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		poolLogger := logger.WithName("async-inference").WithValues("pool", poolName)
 		d := newResultDispatcher(p, poolLogger, config.ResultPollTimeout)
 		pool := &asyncPool{
-			producer:         p,
-			dispatcher:       d,
-			logger:           poolLogger,
-			defaultDeadline:  config.DefaultDeadline,
-			resultBufferSize: config.ResultBufferSize,
+			producer:   p,
+			dispatcher: d,
+			logger:     poolLogger,
 		}
 
 		pools[model] = pool
--- a/pkg/clients/inference/async_inference_client_resolver.go
+++ b/pkg/clients/inference/async_inference_client_resolver.go
@@ -97,6 +97,11 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		poolToModel[poolName] = model
 	}
 
+	if config.ResultPollTimeout <= 0 {
+		_ = rdb.Close()
+		return nil, fmt.Errorf("ResultPollTimeout must be > 0")
+	}
+
 	pools := make(map[string]*asyncPool, len(config.Models))
 	var closers []io.Closer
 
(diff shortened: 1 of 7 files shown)


lioraron:
@yizhaodev @vishbhat @j-mok-dev Please review this PR when available?
 Thanks.

pkg/clients/inference/async_inference_client_resolver.go
```diff
@@ -117,8 +117,12 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		}
 
 		poolLogger := logger.WithName("async-inference").WithValues("pool", poolName)
-		d := newResultDispatcher(p, poolLogger)
-		pool := &asyncPool{producer: p, dispatcher: d, logger: poolLogger, defaultDeadline: config.DefaultDeadline}
+		d := newResultDispatcher(p, poolLogger, config.ResultPollTimeout)
```
vishbhat, line 125: nit: `NewAsyncResolver` is a public function but doesn't validate `ResultPollTimeout`. Before this PR the poll timeout was hardcoded to `time.Second`, so no external input could go wrong. Now that it's caller-supplied, a zero value causes `context.WithTimeout(ctx, 0)` to return an immediately-cancelled context on every iteration, turning the `for ctx.Err() == nil` loop into a CPU-burning spin loop. Production is protected by `validateAsyncDispatchConfig`, but the public API itself isn't.

**Suggestion:**
```go
if config.ResultPollTimeout <= 0 {
    return nil, fmt.Errorf("ResultPollTimeout must be > 0")
}

File: pkg/clients/inference/async_inference_client_test.go [hunk 2]
```diff
@@ -231,6 +231,61 @@ func TestAsyncProducerClient_Submit(t *testing.T) {
...
+	resp, err := client.GetResult(ctx)
+	if err != nil {
+		t.Fatalf("GetResult error: %v", err)
+	}
+	if resp.RequestID != "panic-1" {
+		t.Errorf("RequestID = %q, want %q", resp.RequestID, "panic-1")
+	}
+}
+
+func TestAsyncProducerClient_CloseHandlesNonStringKey(t *testing.T) {
```
vishbhat: nit: Per CLAUDE.md: _"Group tests for the same function under one `TestXxx` using `t.Run()`, not separate functions."_ `TestAsyncProducerClient_Submit` already has a `t.Run("close unregisters pending waiters", ...)` subtest at line 153. This should be another `t.Run` there (e.g., `t.Run("close handles non-string key without panic", ...)`) rather than a standalone function.

pkg/clients/inference/async_inference_client_resolver.go
lioraron: Done, thanks.

pkg/clients/inference/async_inference_client_test.go
lioraron: Done, thanks.

vishbhat (approved):

PR #554, opened 2026-07-06
enh: replace local input file copy with S3 ranged gets

### LINKED ISSUE #561: Eliminate local input file copy from executor to reduce emptyDir pressure and improve crash safety
## Problem

The preprocessor streams the full input file from S3 and writes a local copy (`input.jsonl`) to `emptyDir` so the executor can do random-access reads by byte offset during execution. This has two issues:

### Disk pressure

With [`numWorkers: 20`](https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L307) (the operational default) and the OpenAI max of 200 MB per input file, input copies alone can consume up to 4 GB of `emptyDir`. The [`sizeLimit` defaults to 10Gi](https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L237), so input copies can use 40% of the budget before accounting for output/error files growing during execution.

### Crash safety

Running out of `emptyDir` causes pod eviction, which takes out every in-flight job on that worker. Because startup recovery is `workdir`-based, eviction destroys partial outputs too — those jobs land as failed with nothing salvageable when the orphan reconciler notices. The input file copy is the largest unbounded consumer of that disk budget.

## Proposed solution

Add a `RetrieveRange` method to the `BatchFilesClient` interface that fetches a byte range from storage (using the S3 `Range` header). The executor reads each request line on demand from shared storage via ranged GETs instead of a local file copy. The preprocessor continues to stream the full file for validation and plan building, but no longer writes a local copy.

### Why this is safe performance-wise

The [global concurrency semaphore](https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/worker.go#L169) caps total in-flight inference requests at [100 across all workers](https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L313). Each goroutine does one S3 ranged GET (5-50ms) before an inference call (hundreds of ms to seconds). At any given moment, at most a handful of goroutines are in the S3 read phase — the rest are blocked on inference. S3 supports [5,500 GET/s per prefix](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html). The overhead is negligible for a batch system.

### Pre-sorted input files

The input JSONL already contains `body.model` and `body.messages` (including system prompts) in every line. The files API could sort lines by PrefixHash at upload time — it already parses each line for validation, so sorting adds negligible overhead. This makes plan entry offsets sequential in the stored file, eliminating the gap between file byte order and plan dispatch order.

### Partitioned paged readers

With a sorted file, the executor can partition entries into N contiguous offset ranges, each backed by a small paged reader (e.g., 4 MB buffer). Since entries within each partition are contiguous, the buffer slides forward monotonically with zero backward jumps.

This applies to both execution paths:
- **Sync path (`processModel`):** move `readRequestLine` before the goroutine spawn so reads are sequential per partition. Each partition scans forward through its range.
- **Async path (`processModelAsync`):** split the currently-sequential submit loop into N concurrent submit goroutines, each with its own partition and paged reader. This parallelizes the submit phase, which is the bottleneck for large async batches.

### Estimated numbers (200 MB file, 4 partitions, 4 MB buffer, 20 workers)

| Metric | Value |
|--------|-------|
| S3 GETs per job | ~50 (vs ~50K without paging) |
| Memory per worker | 16 MB |
| Memory across 20 workers | 320 MB |
| Concurrent S3 GETs per worker | ~1-4 (page fetches are infrequent — each page covers ~8K entries of inference) |

acardace:
## Summary

The preprocessor streams the input file from S3 and writes a full copy to local
disk so the executor can do random-access reads by byte offset. With 20 workers
at the OpenAI max of 200 MB per file, this consumes up to 4 GB of `emptyDir` for
input copies alone, plus the output/error files growing during execution.

This PR adds a `RetrieveRange` method to the `BatchFilesClient` interface that
fetches a byte range from storage (using the S3 `Range` header), and replaces all
local input file reads in the executor with ranged storage reads.

### Interface & implementations
- Add `RetrieveRange(ctx, fileName, folderName, offset, length)` to `BatchFilesClient`
- Implement in S3 (Range header), FS (ReadAt), retryclient, tracing, and mock

### Preprocessor
- Remove local `input.jsonl` write — the preprocessor still streams the full file
  for validation and plan building, but no longer writes a local copy
- Plan entry offsets are unchanged — they track byte positions in the stream,
  which match the remote file layout

### Executor
- Add `inputFileRef` struct holding remote storage coordinates (`storageName`, `folderName`)
- Add `resolveInputFileCoords` — resolves an `inputFileID` to storage coordinates via DB lookup (once per job)
- Convert `readRequestLine` from package-level function to `Processor` method that calls `RetrieveRange` internally
- Replace `*os.File` with `*inputFileRef` throughout `executeJob`, `processModel`, `processModelAsync`, `executeOneRequest`, `drainAndFinalize`, `drainUnprocessedRequests`
- `drainUnprocessedRequests` takes a separate `storageCtx` (`mainCtx`) for S3 reads since `progressCtx` (`requestAbortCtx`) may be cancelled during drain

### Removed
- `createLocalInputFile`, `jobInputFilePath`, `inputFileName` constant

## What stays on local disk
- Plan files (small, sequential reads)
- `output.jsonl` and `error.jsonl` (could be streamed to S3 as a follow-up)

## Testing
- All 1051 tests pass
- Regression tests pass
- New `TestRetrieveRange` tests for S3, FS, and retryclient implementations

Fixes https://github.com/llm-d/llm-d-batch-gateway/issues/561

### DIFF
--- a/internal/files_store/fs/client.go
+++ b/internal/files_store/fs/client.go
@@ -18,6 +18,7 @@ limitations under the License.
 package fs
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -211,6 +212,32 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
 		"path", relPath, "size", metadata.Size)
 
 	return file, metadata, nil
+}
+
+// RetrieveRange retrieves a byte range [offset, offset+length) from a file in the filesystem.
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	relPath, err := c.resolvePath(folderName, fileName)
+	if err != nil {
+		return nil, err
+	}
+
+	file, err := c.root.Open(relPath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	buf := make([]byte, length)
+	n, err := file.ReadAt(buf, offset)
+	if err != nil && (err != io.EOF || int64(n) != length) {
+		return nil, fmt.Errorf("read range at offset %d length %d: %w", offset, length, err)
+	}
+	buf = buf[:n]
+
+	logr.FromContextOrDiscard(ctx).V(logging.TRACE).Info("Range retrieved",
+		"path", relPath, "offset", offset, "length", length)
+
+	return io.NopCloser(bytes.NewReader(buf)), nil
 }
 
 // Delete deletes a file from the filesystem.
--- a/internal/files_store/retryclient/client.go
+++ b/internal/files_store/retryclient/client.go
@@ -115,6 +115,33 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
 	return rc, meta, err
 }
 
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	var rc io.ReadCloser
+	attempts, err := retry.Do(ctx, &c.cfg, func(attempt int) error {
+		if attempt > 1 {
+			if rc != nil {
+				_ = rc.Close()
+			}
+			recordRetry("retrieve_range", c.component)
+			logr.FromContextOrDiscard(ctx).Info("Retrying file retrieve range",
+				"file", fileName, "offset", offset, "length", length,
+				"attempt", attempt, "maxRetries", c.cfg.MaxRetries)
+		}
+
+		var retrieveErr error
+		rc, retrieveErr = c.inner.RetrieveRange(ctx, fileName, folderName, offset, length)
+		return retrieveErr
+	})
+	if err != nil {
+		if attempts > c.cfg.MaxRetries {
+			recordExhausted("retrieve_range", c.component)
+		}
+	} else {
+		recordSuccess("retrieve_range", c.component)
+	}
+	return rc, err
+}
+
 func (c *Client) Delete(ctx context.Context, fileName, folderName string) error {
 	attempts, err := retry.Do(ctx, &c.cfg, func(attempt int) error {
 		if attempt > 1 {
--- a/internal/files_store/tracing/tracing.go
+++ b/internal/files_store/tracing/tracing.go
@@ -105,6 +105,26 @@ func (r *tracedReadCloser) Close() error {
 	return err
 }
 
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	_, span := uotel.StartSpan(ctx, "storage.RetrieveRange")
+	span.SetAttributes(
+		c.backend,
+		attribute.String("storage.file_name", fileName),
+		attribute.String("storage.folder", folderName),
+		attribute.Int64("storage.offset", offset),
+		attribute.Int64("storage.length", length),
+	)
+
+	reader, err := c.inner.RetrieveRange(ctx, fileName, folderName, offset, length)
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, "retrieve range failed")
+		span.End()
+		return nil, err
+	}
+	return &tracedReadCloser{ReadCloser: reader, span: span}, nil
+}
+
 func (c *Client) Delete(ctx context.Context, fileName, folderName string) error {
 	ctx, span := uotel.StartSpan(ctx, "storage.Delete")
 	defer span.End()
--- a/internal/apiserver/file/file_handler_test.go
+++ b/internal/apiserver/file/file_handler_test.go
@@ -154,6 +154,10 @@ func (c *errStoreFileClient) Retrieve(ctx context.Context, fileName, folderName 
 	return nil, nil, errors.New("not implemented")
 }
 
+func (c *errStoreFileClient) RetrieveRange(_ context.Context, _, _ string, _ int64, _ int64) (io.ReadCloser, error) {
+	return nil, errors.New("not implemented")
+}
+
 func (c *errStoreFileClient) Delete(ctx context.Context, fileName, folderName string) error {
 	return nil
 }
--- a/internal/files_store/api/files.go
+++ b/internal/files_store/api/files.go
@@ -52,6 +52,9 @@ type BatchFilesClient interface {
 	// Retrieve retrieves a file from the files storage.
 	Retrieve(ctx context.Context, fileName, folderName string) (reader io.ReadCloser, fileMd *BatchFileMetadata, err error)
 
+	// RetrieveRange retrieves a byte range [offset, offset+length) from a file in storage.
+	RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error)
+
 	// Delete deletes the file in the specified location.
 	Delete(ctx context.Context, fileName, folderName string) (err error)
 }
--- a/internal/files_store/fs/client_test.go
+++ b/internal/files_store/fs/client_test.go
@@ -276,6 +276,41 @@ func TestRetrieve(t *testing.T) {
 
 }
 
+func TestRetrieveRange(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("retrieves byte range", func(t *testing.T) {
+		client := newTestClient(t)
+		content := []byte("hello world range test")
+
+		if _, err := client.Store(ctx, "range.txt", testFolder, 1024, 0, bytes.NewReader(content)); err != nil {
+			t.Fatalf("failed to store: %v", err)
+		}
+
+		rc, err := client.RetrieveRange(ctx, "range.txt", testFolder, 6, 5)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		defer rc.Close()
+
+		data, _ := io.ReadAll(rc)
+		if string(data) != "world" {
+			t.Errorf("expected %q, got %q", "world", string(data))
+		}
+	})
+
+	t.Run("returns error for non-existent file", func(t *testing.T) {
+		client := newTestClient(t)
+
+		_, err := client.RetrieveRange(ctx, "nonexistent.txt", testFolder, 0, 10)
+		if err != nil {
+			// Expect an error (file not found).
+			return
+		}
+		t.Error("expected error for non-existent file")
+	})
+}
+
 func TestDelete(t *testing.T) {
 	ctx := context.Background()
 
--- a/internal/files_store/mock/mock_files_client.go
+++ b/internal/files_store/mock/mock_files_client.go
@@ -145,7 +145,10 @@ func (m *MockBatchFilesClient) Retrieve(ctx context.Context, fileName, folderNam
 }
 
 // RetrieveRange retrieves a byte range [offset, offset+length) from a file.
-func (m *MockBatchFilesClient) RetrieveRange(_ context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+func (m *MockBatchFilesClient) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
 	filePath := filepath.Join(m.rootDir, folderName, fileName)
 
 	file, err := os.Open(filePath)
--- a/internal/files_store/mock/mock_files_client.go
+++ b/internal/files_store/mock/mock_files_client.go
@@ -19,6 +19,7 @@ package mock
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -141,6 +142,26 @@ func (m *MockBatchFilesClient) Retrieve(ctx context.Context, fileName, folderNam
 		Size:     fileInfo.Size(),
 		ModTime:  fileInfo.ModTime(),
 	}, nil
+}
+
+// RetrieveRange retrieves a byte range [offset, offset+length) from a file.
+func (m *MockBatchFilesClient) RetrieveRange(_ context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	filePath := filepath.Join(m.rootDir, folderName, fileName)
+
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file: %w", err)
+	}
+	defer file.Close()
+
+	buf := make([]byte, length)
+	n, err := file.ReadAt(buf, offset)
+	if err != nil && (err != io.EOF || int64(n) != length) {
+		return nil, fmt.Errorf("failed to read range at offset %d length %d: %w", offset, length, err)
+	}
+	buf = buf[:n]
+
+	return io.NopCloser(bytes.NewReader(buf)), nil
 }
 
 // List lists the files in the specified location.
--- a/internal/files_store/retryclient/client_test.go
+++ b/internal/files_store/retryclient/client_test.go
@@ -63,6 +63,14 @@ func (m *mockFilesClient) Delete(_ context.Context, _, _ string) error {
 	return nil
 }
 
+func (m *mockFilesClient) RetrieveRange(_ context.Context, _, _ string, _ int64, _ int64) (io.ReadCloser, error) {
+	m.retrieveCalls++
+	if m.retrieveCalls <= m.failUntil {
+		return nil, errors.New("transient retrieve range error")
+	}
+	return io.NopCloser(bytes.NewReader([]byte("data"))), nil
+}
+
 func (m *mockFilesClient) Close() error { return nil }
 
 func retryCfg() retry.Config {
@@ -190,6 +198,24 @@ func TestRetrieve_SucceedsAfterRetry(t *testing.T) {
 	defer rc.Close()
 	if meta.Size != 4 {
 		t.Fatalf("expected size 4, got %d", meta.Size)
+	}
+	if mock.retrieveCalls != 2 {
+		t.Fatalf("expected 2 retrieve calls, got %d", mock.retrieveCalls)
+	}
+}
+
+func TestRetrieveRange_SucceedsAfterRetry(t *testing.T) {
+	mock := &mockFilesClient{failUntil: 1}
+	c := New(mock, retryCfg(), "test")
+
+	rc, err := c.RetrieveRange(context.Background(), "f.txt", "folder", 0, 4)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	defer rc.Close()
+	data, _ := io.ReadAll(rc)
+	if string(data) != "data" {
+		t.Fatalf("expected %q, got %q", "data", string(data))
 	}
 	if mock.retrieveCalls != 2 {
 		t.Fatalf("expected 2 retrieve calls, got %d", mock.retrieveCalls)
--- a/internal/files_store/s3/client.go
+++ b/internal/files_store/s3/client.go
@@ -274,6 +274,31 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
 	return out.Body, metadata, nil
 }
 
+// RetrieveRange retrieves a byte range [offset, offset+length) from a file in S3.
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	key := c.buildKey(folderName, fileName)
+	rangeHeader := fmt.Sprintf("bytes=%d-%d", offset, offset+length-1)
+
+	out, err := c.s3Client.GetObject(ctx, &s3.GetObjectInput{
+		Bucket: aws.String(c.bucket),
+		Key:    aws.String(key),
+		Range:  aws.String(rangeHeader),
+	})
+	if err != nil {
+		var noSuchKey *types.NoSuchKey
+		var noSuchBucket *types.NoSuchBucket
+		if errors.As(err, &noSuchKey) || errors.As(err, &noSuchBucket) {
+			return nil, os.ErrNotExist
+		}
+		return nil, err
+	}
+
+	logr.FromContextOrDiscard(ctx).V(logging.TRACE).Info("Range retrieved",
+		"bucket", c.bucket, "key", key, "offset", offset, "length", length)
+
+	return out.Body, nil
+}
+
 // Delete deletes a file from S3.
 // The folderName parameter is used as a key prefix for tenant isolation.
 func (c *Client) Delete(ctx context.Context, fileName, folderName string) error {
--- a/internal/files_store/s3/client_test.go
+++ b/internal/files_store/s3/client_test.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"testing"
@@ -98,9 +99,20 @@ func (m *mockS3Client) GetObject(_ context.Context, params *s3.GetObjectInput, _
 	if !ok {
 		return nil, &types.NoSuchKey{}
 	}
+	data := obj.data
+	if params.Range != nil {
+		// Parse "bytes=start-end" range header.
+		var start, end int64
+		if _, err := fmt.Sscanf(*params.Range, "bytes=%d-%d", &start, &end); err == nil {
+			if end >= int64(len(data)) {
+				end = int64(len(data)) - 1
+			}
+			data = data[start : end+1]
+		}
+	}
 	return &s3.GetObjectOutput{
-		Body:          io.NopCloser(bytes.NewReader(obj.data)),
-		ContentLength: aws.Int64(int64(len(obj.data))),
+		Body:          io.NopCloser(bytes.NewReader(data)),
+		ContentLength: aws.Int64(int64(len(data))),
 		LastModified:  aws.Time(obj.lastModTime),
 	}, nil
 }
@@ -331,6 +343,41 @@ func TestRetrieve(t *testing.T) {
 		client := newTestClient(mock)
 
 		_, _, err := client.Retrieve(ctx, "nonexistent.txt", testFolderName)
+		if !errors.Is(err, os.ErrNotExist) {
+			t.Errorf("expected os.ErrNotExist, got %v", err)
+		}
+	})
+}
+
+func TestRetrieveRange(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("retrieves byte range", func(t *testing.T) {
+		mock := newMockS3Client()
+		client := newTestClient(mock)
+		content := []byte("hello world range test")
+
+		if _, err := client.Store(ctx, "range.txt", testBucketName, 1024, 0, bytes.NewReader(content)); err != nil {
+			t.Fatalf("failed to store: %v", err)
+		}
+
+		rc, err := client.RetrieveRange(ctx, "range.txt", testBucketName, 6, 5)
+		if err != nil {
+			t.Fatalf("expected no error, got %v", err)
+		}
+		defer rc.Close()
+
+		data, _ := io.ReadAll(rc)
+		if string(data) != "world" {
+			t.Errorf("expected %q, got %q", "world", string(data))
+		}
+	})
+
+	t.Run("returns error for non-existent file", func(t *testing.T) {
+		mock := newMockS3Client()
+		client := newTestClient(mock)
+
+		_, err := client.RetrieveRange(ctx, "nonexistent.txt", testBucketName, 0, 10)
 		if !errors.Is(err, os.ErrNotExist) {
 			t.Errorf("expected os.ErrNotExist, got %v", err)
 		}
--- a/internal/processor/worker/execute_pipeline.go
+++ b/internal/processor/worker/execute_pipeline.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/llm-d/llm-d-batch-gateway/pkg/clients/inference"
 
+	filesapi "github.com/llm-d/llm-d-batch-gateway/internal/files_store/api"
 	"github.com/llm-d/llm-d-batch-gateway/internal/processor/batchctx"
 	"github.com/llm-d/llm-d-batch-gateway/internal/processor/config"
 	"github.com/llm-d/llm-d-batch-gateway/internal/processor/pipeline"
@@ -48,6 +50,29 @@ func (p *Processor) executeJobAsync(ctx context.Context, params *jobExecutionPar
 		sloDeadline = params.task.SLO
 	}
 
+	var (
+		inputRef  *inputFileRef
+		inputFile *os.File
+	)
+	if params.jobInfo != nil && params.jobInfo.BatchJob != nil && params.jobInfo.BatchJob.InputFileID != "" {
+		ref, err := p.resolveInputFileCoords(ctx, params.jobInfo.BatchJob.InputFileID)
+		if err != nil {
+			return nil, fmt.Errorf("resolve input file coordinates: %w", err)
+		}
+		inputRef = ref
+	} else {
+		localInputPath := filepath.Join(jobRootDir, "input.jsonl")
+		if f, err := os.Open(localInputPath); err == nil {
+			inputFile = f
+			defer inputFile.Close()
+		}
+	}
+
+	var storage filesapi.BatchFilesClient
+	if p.files != nil {
+		storage = p.files.storage
+	}
+
 	// Setup pipeline.
 
 	tracker := pipeline.NewProgressTracker(
@@ -60,7 +85,9 @@ func (p *Processor) executeJobAsync(ctx context.Context, params *jobExecutionPar
 	tracker.AddFailed(modelMap.RejectedCount)
 
 	source := NewPlanFileSource(PlanFileSourceConfig{
-		InputFile:          files.input,
+		Storage:            storage,
+		InputRef:           inputRef,
+		InputFile:          inputFile,
 		PlansDir:           plansDir,
 		ModelMap:           modelMap,
 		Resolver:           p.inference,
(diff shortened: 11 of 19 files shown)


acardace:
@wseaton @vMaroon @evacchi

internal/files_store/fs/client.go
```diff
@@ -213,6 +214,30 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
...
+	}
+
+	file, err := c.root.Open(relPath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	buf := make([]byte, length)
+	if _, err := file.ReadAt(buf, offset); err != nil {
```
yizhaodev: there miight be an issue
`func readNormalizedLine` in `preprocessor.go` in will add `\n`, so the file length is N, but plan is N+1
then `file.ReadAt(buf, N+1)` will return EOF error?

lioraron:
Thanks for looking into disk usage optimization, Antonio. Before we proceed with the review, I want to share some concerns about the approach and suggest a different workflow.

**Process suggestion:** For changes of this scope (18 files, new interface method, rearchitecting the executor's I/O path), it would be better to start with an issue describing the problem and proposed design. That gives us a chance to discuss trade-offs and agree on an approach before investing in implementation. Happy to collaborate on the design there.

**On the problem statement:**

The PR describes a worst-case of 20 workers × 200MB = 4GB of `emptyDir` pressure. Looking at the current code:

- `NumWorkers` defaults to **1** — the 4GB scenario requires non-default configuration with 20 concurrent max-size batch jobs
- `cleanupJobArtifacts()` calls `os.RemoveAll(jobDir)` on every terminal path (success, cancel, expiry, error, recovery) — files don't accumulate; each worker holds at most one input file at a time
- Realistic disk pressure at defaults is ~200MB, not 4GB

The problem is real in theory at extreme scale, but it's not clear this is hitting anyone today.

**On the proposed solution:**

Replacing local `ReadAt` with per-request S3 range-GETs has significant trade-offs:

- **Latency:** Local `ReadAt` with OS page cache is <0.01ms. S3 range-GET is ~5–50ms. This latency is added to **every request** on the critical dispatch path
- **New failure mode:** Every batch request now depends on S3 being reachable at dispatch time. Network blips or S3 throttling become per-request failures instead of a one-time download at preprocessing
- **Scale:** A 1000-request batch = 1000 S3 GETs per job. With `Global: 100` concurrency, that's up to 100 concurrent S3 calls per worker. At scale this adds cost and pressure on S3 rate limits
- **Complexity:** Threading `inputFileRef` + `storageCtx` through the executor, drain, and finalize paths adds surface area for bugs (e.g., the separate `storageCtx` for drain)

**Simpler alternatives worth discussing:**

If disk pressure does become a real concern at high worker counts, there are lighter-weight options:

1. **Kubernetes `emptyDir.sizeLimit`** — let operators control disk allocation natively, no code change
2. **In-memory index** — the preprocessor already streams the full file for validation; it could retain parsed lines in a `[][]byte` keyed by plan entry, eliminating both disk _and_ S3 calls
3. **Memory-mapped file** — same random-access pattern as today, slightly lower memory overhead
4. **Configurable max concurrent input files** — a simple semaphore that bounds how many input copies exist at once, independent of `NumWorkers`

I'd suggest we close this PR, open an issue with the problem statement and these alternatives, and converge on a design before writing code. What do you think?

acardace:
> Thanks for looking into disk usage optimization, Antonio. Before we proceed with the review, I want to share some concerns about the approach and suggest a different workflow.
> 
> **Process suggestion:** For changes of this scope (18 files, new interface method, rearchitecting the executor's I/O path), it would be better to start with an issue describing the problem and proposed design. That gives us a chance to discuss trade-offs and agree on an approach before investing in implementation. Happy to collaborate on the design there.
> 
> **On the problem statement:**
> 
> The PR describes a worst-case of 20 workers × 200MB = 4GB of `emptyDir` pressure. Looking at the current code:
> 
> * `NumWorkers` defaults to **1** — the 4GB scenario requires non-default configuration with 20 concurrent max-size batch jobs
> * `cleanupJobArtifacts()` calls `os.RemoveAll(jobDir)` on every terminal path (success, cancel, expiry, error, recovery) — files don't accumulate; each worker holds at most one input file at a time
> * Realistic disk pressure at defaults is ~200MB, not 4GB

NumWorkers defaults to 1 in the Go config code (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/config/config.go#L300), but both the Helm chart (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L307) and the shipped config file (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/cmd/batch-processor/config.yaml#L4) override it to 20, which is the operational default that all deployments use. So the 20 × 200 MB = 4 GB calculation applies to any default deployment with max-size files. The emptyDir sizeLimit defaults to 10Gi (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L237), and input copies alone can consume 40% of that budget before accounting for output/error files growing during execution.

> 
> The problem is real in theory at extreme scale, but it's not clear this is hitting anyone today.
> 
> **On the proposed solution:**
> 
> Replacing local `ReadAt` with per-request S3 range-GETs has significant trade-offs:
> 
> * **Latency:** Local `ReadAt` with OS page cache is <0.01ms. S3 range-GET is ~5–50ms. This latency is added to **every request** on the critical dispatch path

Per-request latency overhead of S3 range GETs (5–50ms) is irrelevant for the batch gateway. Inference latency is hundreds of ms to seconds — the S3 overhead is noise. This is a batch system, not a real-time serving path.

> * **New failure mode:** Every batch request now depends on S3 being reachable at dispatch time. Network blips or S3 throttling become per-request failures instead of a one-time download at preprocessing

S3 is already a required dependency. A network blip during execution that prevents S3 range reads would equally prevent finalization from uploading results. The failure mode isn't new, it's the same dependency the system already has. Additionally, the retryclient wrapper handles transient errors with exponential backoff.

> * **Scale:** A 1000-request batch = 1000 S3 GETs per job. With `Global: 100` concurrency, that's up to 100 concurrent S3 calls per worker. At scale this adds cost and pressure on S3 rate limits

100 concurrent S3 GET requests is well within S3's capabilities. AWS documents 5,500 GET requests per second per prefix (https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html), and S3-compatible stores (MinIO, Ceph) handle similar throughput. 100 concurrent requests is not a concern.

> * **Complexity:** Threading `inputFileRef` + `storageCtx` through the executor, drain, and finalize paths adds surface area for bugs (e.g., the separate `storageCtx` for drain)

The multiple-context pattern is already used extensively throughout the executor. On main, there are 10 functions that take 2+ context parameters:
- executeJob (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L182) — 4 contexts (ctx, sloCtx, userCancelCtx, requestAbortCtx)
- processModel (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L405) — 4 contexts
- processModelAsync (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L546) - 4 contexts
- drainAndFinalize (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L704) - 4 contexts
- executeOneRequest (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L916) - 2 contexts
- writeResult (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L989) - 3 contexts
- preProcessJob (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/preprocessor.go#L47) - 3 contetxs
- finalizeJob (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/finalizer.go#L90) - 2 contexts
- runPollingLoop (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/worker.go#L243) - 2 contexts
Adding storageCtx to drainUnprocessedRequests is consistent with this existing pattern, not a new category of complexity.

> 
> **Simpler alternatives worth discussing:**
> 
> If disk pressure does become a real concern at high worker counts, there are lighter-weight options:
> 
> 1. **Kubernetes `emptyDir.sizeLimit`** — let operators control disk allocation natively, no code change

This is already set (10Gi default (https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L237)). It doesn't solve the problem, it just makes the pod get evicted when it exceeds the limit. The point is to not need the disk space at all.

> 2. **In-memory index** — the preprocessor already streams the full file for validation; it could retain parsed lines in a `[][]byte` keyed by plan entry, eliminating both disk _and_ S3 calls

This moves the same 200 MB × 20 workers = 4 GB problem from disk to RAM, which is worse, RAM is more expensive, and OOM kills are harder to recover from than disk pressure.

> 3. **Memory-mapped file** — same random-access pattern as today, slightly lower memory overhead

Still requires the file on disk. Doesn't reduce disk usage at all, it just changes how the kernel pages it.

> 4. **Configurable max concurrent input files** — a simple semaphore that bounds how many input copies exist at once, independent of `NumWorkers`

Adds throttling complexity to work around a resource problem that this PR eliminates entirely.

internal/files_store/fs/client.go
acardace: I think you're right, thank you! Will fix this.

wseaton:
+1 on the direction. I view this type of change necessary as a crash safety improvement, not just a disk usage optimization. Running out of emptyDir can cause pod eviction and takes out every in-flight job on that worker. Because startup recovery is workdir-based, eviction destroys the partial outputs too, so those jobs land as failed with nothing salvageable whenever the orphan reconciler notices. Dropping the biggest unbounded consumer of that disk budget, and reading from the source of truth instead of a local copy, is actually less moving parts from a correctness PoV.

This is also a step on the rung to more stateless workers in general. The less we can use temporary scratch space the better.

internal/processor/worker/executor.go
```diff
@@ -770,35 +766,28 @@ func (p *Processor) drainAndFinalize(
...
+	logger := logr.FromContextOrDiscard(progressCtx)
 
 	for _, entry := range entries {
 		customID := ""
-		if _, err := inputFile.ReadAt(buf[:entry.Length], entry.Offset); err == nil {
-			var req batch_types.Request
-			if err := json.Unmarshal(bytes.TrimSuffix(buf[:entry.Length], []byte{'\n'}), &req); err == nil {
-				customID = req.CustomID
-			}
+		req, _, _, readErr := p.readRequestLine(storageCtx, entry, inputRef, logger)
```
wseaton: If entries are known apriori contiguous here, we can batch them together and do one combined read as an optimization during the drain, correct?

acardace: Yes, very good point!

lioraron:
This discussion is a good example of why design issues before PRs matter — we're iterating on fundamental trade-offs in PR comments, which is harder to follow and means implementation work may need to be redone.

The PR does one S3 range GET per request in the batch. For `Global: 100` concurrency, that's up to 100 concurrent S3 calls sustained throughout execution. This applies not only to the drain path but also to `processModel` and `processModelAsync`.

There can be options to reduce the S3 calls volume by chunking and using buffers, e.g. for contiguous ranges or per model.

wseaton:
> The PR does one S3 range GET per request in the batch. For Global: 100 concurrency, that's up to 100 concurrent S3 calls sustained throughout execution. This applies not only to the drain path but also to processModel and processModelAsync.

100 sustained concurrent s3 calls is impossible in practice, no? There is a global rate limiter that affects the entire request lifecycle which is high hundreds of milliseconds to seconds of latency. The inference throughput would need to be several orders of magnitude faster to cause that level of S3 read contention

> There can be options to reduce the S3 calls volume by chunking and using buffers, e.g. for contiguous ranges or per model.

I think these optimizations are usually non-obvious without some initial implementation/PoC work, up front design is usually not the best place to hash those specifics out, experimentation and benchmarking will be required.

internal/processor/worker/executor.go
wseaton: Looking at this more I think we might be able to skip this read entirely if we start storing the custom_id in the plan entry and make it available here, since it's pre-calculated and then thrown away

lioraron:
> I think these optimizations are usually non-obvious without some initial implementation/PoC work, up front design is usually not the best place to hash those specifics out, experimentation and benchmarking will be required.

I agree, however these types of large refactoring projects that have high level trade-offs and options, should be documented and discussed first in an organized proposal, rather than immediate PR.

internal/processor/worker/executor.go
acardace: An issue is that the the plan entry is a fixed 16 bytes binary record, adding custom_id would break the fixed-size invariant. Honestly the plan files could be just moved to memory and custom_id be added. Maybe as a follow up?

acardace: and tbh we can do batched reads anyway (with a fixed-size in-memory buffer) for multiple entries even in the normal path of a job execution.

acardace:
> > I think these optimizations are usually non-obvious without some initial implementation/PoC work, up front design is usually not the best place to hash those specifics out, experimentation and benchmarking will be required.
> 
> I agree, however these types of large refactoring projects that have high level trade-offs and options, should be documented and discussed first in an organized proposal, rather than immediate PR.

I can create a github issue but I don't think this deserves a design document or anything like that. I think this PR solves an obvious issue. I've already answered point by point in https://github.com/llm-d/llm-d-batch-gateway/pull/554#issuecomment-4903259578. 

Also PRs about implementation specific things like these is usually where discussions happen.

internal/processor/worker/executor.go
```diff
@@ -595,7 +591,7 @@ func (p *Processor) processModelAsync(
 			break
 		}
 
-		req, batchReqID, parseErr, readErr := readRequestLine(inputFile, entry, logger)
+		req, batchReqID, parseErr, readErr := p.readRequestLine(requestAbortCtx, entry, inputRef, logger)
 		if readErr != nil {
 			return readErr
```
lioraron, line 596: Bug: context cancellation skips `drainAndFinalize`, leaving entries unaccounted.

`readRequestLine` now calls `RetrieveRange(requestAbortCtx, ...)`. When a sibling model cancels `requestAbortCtx` between the loop's context check (line 588) and this call, `RetrieveRange` returns `context.Canceled`. This `return readErr` exits `processModelAsync` immediately — `drainAndFinalize` is never called, so entries `[submitCount:]` are never recorded.

Result: `completed + failed < total`, the batch hangs forever waiting for entries that will never be accounted for.

The old `ReadAt` on a local `*os.File` was context-independent and immune to this race. The fix is to treat context cancellation errors from `readRequestLine` like the existing `requestAbortCtx.Err()` check above (break out of the loop and fall through to `drainAndFinalize`) rather than returning early.

internal/files_store/retryclient/client.go
```diff
@@ -115,6 +115,30 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
 	return rc, meta, err
 }
 
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	var rc io.ReadCloser
+	attempts, err := retry.Do(ctx, &c.cfg, func(attempt int) error {
```
lioraron, line 120: Bug: `RetrieveRange` leaks the `ReadCloser` from the failed attempt on retry.

The existing `Retrieve` method (line 96-98) has:
```go
if attempt > 1 {
    if rc != nil {
        _ = rc.Close()
    }
```

`RetrieveRange` is missing this cleanup. When a transient S3 error triggers a retry, the response body from the failed attempt is never closed — leaking an HTTP connection back to the S3 connection pool.

internal/files_store/fs/client.go
```diff
@@ -213,6 +214,30 @@ func (c *Client) Retrieve(ctx context.Context, fileName, folderName string) (io.
...
+		return nil, err
+	}
+
+	file, err := c.root.Open(relPath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	buf := make([]byte, length)
```
lioraron, line 230: Bug: `ReadAt` returns `(n, io.EOF)` when the read reaches exactly the end of file.

Per Go docs: *"ReadAt always returns a non-nil error when n < len(p). At end of file, that error is io.EOF."* When `offset+length` equals the file size exactly (which happens on the last entry of every batch), `ReadAt` returns `(length, io.EOF)` — the read succeeds (all bytes read) but this code treats any non-nil error as failure.

Fix: check for `io.EOF` specifically and ignore it when `n == length`:
```go
n, err := file.ReadAt(buf, offset)
if err != nil && !(err == io.EOF && n == int(length)) {
    return nil, fmt.Errorf(...)
}
buf = buf[:n]
```

internal/processor/worker/executor.go
```diff
@@ -281,31 +277,31 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
...
-					sloCtx,
-					userCancelCtx,
-					inputFile,
-					plansDir, safeModelID, modelID,
-					writers,
-					progress,
-					passThroughHeaders,
-					tenantID,
-				)
+			err = p.processModelAsync(
```
lioraron: The `if/else` block is dedented by one tab relative to its enclosing scope. The `if p.asyncInference != nil {` (line 279) is at 3 tabs, but the body `err = p.processModelAsync(` (line 280) drops to 2 tabs, and `} else {` (line 289) is at 2 tabs while the closing `}` (line 303) stays at 3 tabs. This will fail `gofmt`.

internal/files_store/tracing/tracing.go
```diff
@@ -105,6 +105,26 @@ func (r *tracedReadCloser) Close() error {
 	return err
 }
 
+func (c *Client) RetrieveRange(ctx context.Context, fileName, folderName string, offset int64, length int64) (io.ReadCloser, error) {
+	ctx, span := uotel.StartSpan(ctx, "storage.RetrieveRange")
```
lioraron: Tracing inconsistency: `defer span.End()` ends the span when `RetrieveRange` returns — before the caller reads from the `ReadCloser`.

The existing `Retrieve` method wraps the reader in `tracedReadCloser` which defers `span.End()` until `Close()`, capturing the actual data transfer time. `RetrieveRange` should do the same for consistency, otherwise S3 ranged-read latency is invisible in traces.

lioraron (commented):
@acardace Please create an issue with the details of the problem, solution options, and proposed solution?

internal/processor/worker/executor.go
```diff
@@ -886,11 +875,18 @@ func mergeInferenceHeaders(headers map[string]string, sloCtx context.Context, in
 // readRequestLine reads a single plan entry from the input file, parses it, and
 // generates a batch request ID. Returns the parsed request and batch request ID
 // on success, an outputLine on parse error, or a fatal error on I/O failure.
-func readRequestLine(inputFile *os.File, entry planEntry, logger logr.Logger) (*batch_types.Request, string, *outputLine, error) {
-	buf := make([]byte, entry.Length)
-	if _, err := inputFile.ReadAt(buf, entry.Offset); err != nil {
+func (p *Processor) readRequestLine(ctx context.Context, entry planEntry, inputRef *inputFileRef, logger logr.Logger) (*batch_types.Request, string, *outputLine, error) {
+	rc, err := p.files.storage.RetrieveRange(ctx, inputRef.storageName, inputRef.folderName, entry.Offset, int64(entry.Length))
```
lioraron: Performance: every call to `readRequestLine` issues a separate S3 ranged GET. This affects all code paths — the main async dispatch loop (`processModelAsync`), the sync path (`processModel`/`executeOneRequest`), and the drain path (`drainUnprocessedRequests`).

For a 100K-request batch, that's 100K sequential S3 round-trips in the normal execution path alone, plus potentially thousands more during drain on cancellation/expiry.

This should use batched/chunked reads — fetch a block of consecutive entries in one ranged GET into a buffer, then serve individual `readRequestLine` calls from that buffer. The old code's local `ReadAt` was effectively free; the S3 replacement needs to amortize network overhead.

internal/files_store/fs/client.go
acardace: This is now fixed.

internal/files_store/tracing/tracing.go
acardace: Fixed.

internal/processor/worker/executor.go
acardace: Fixed.

internal/files_store/fs/client.go
acardace: Fixed.

internal/files_store/retryclient/client.go
acardace: Fixed.

internal/processor/worker/executor.go
acardace: Fixed, I just switched readRequestLine to not receive a context and use context.Background() internally so that it behaves the same as before.

internal/processor/worker/executor.go
acardace: How are you getting these numbers?

Right now the code has semaphores which bound the total number of concurrent inference requests. There is a 100 cap on the number of requests that can happen in parallel at any given time across all workers, this means there can only be 100 S3 ranged Gets at any given time, this is far more than acceptable for me. Arguably we could push it further without issues.

https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/charts/batch-gateway/values.yaml#L313
https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/executor.go#L466

Also the goroutine making the inference request will spend only a handful of milliseconds on the S3 Gets, the actual inference request will take multiple seconds in the best case making this time (compared to doing FS reads) really negligible.

In theory there can be only 100 Gets in parallel, in practice I expect there's going to be ~10 at most at any time given that these are really quick compared to everything else the job's goroutine have to do to complete.

One more thing, the async path is completely sequential in that regard, there's 1 S3 get per worker at any given time, so even less of an issue.

lioraron: @acardace I agree the overhead is currently small relative to inference time. My concern is about the design pattern itself.

Per-line network I/O is a universally avoided pattern, even when each individual call is fast. Read-ahead is universally used when sequential access is known, not only because of the per-line-read cost, but also because per-line-read cost is _unnecessary_. The reason is that network round-trips carry fixed overhead (HTTP request/response framing, server-side request processing, S3 API call billing) that doesn't exist with local file I/O. The previous `ReadAt` on a local file got OS page-cache buffering for free — sequential reads were effectively memory reads. S3 has no equivalent; each ranged GET is a full HTTP request regardless of how small the payload is.

Also, "inference dominates" holds today but isn't guaranteed in the future — faster / smaller models, better hardware, or larger batch sizes can modify the ratio. I/O buffering removes the dependency on that assumption.

The fix is also not complex — fetching chunks of consecutive entries in a single ranged GET into a local buffer, then serving individual `readRequestLine` calls from that buffer. This brings the I/O cost profile back in line with the local-file behavior the code replaced.

This is about not introducing an unnecessary per-record network I/O pattern when a standard buffering approach avoids it.

acardace: > @acardace I agree the overhead is currently small relative to inference time. My concern is about the design pattern itself.
> 
> Per-line network I/O is a universally avoided pattern, even when each individual call is fast. Read-ahead is universally used when sequential access is known, not only because of the per-line-read cost, but also because per-line-read cost is _unnecessary_. The reason is that network round-trips carry fixed overhead (HTTP request/response framing, server-side request processing, S3 API call billing) that doesn't exist with local file I/O. The previous `ReadAt` on a local file got OS page-cache buffering for free — sequential reads were effectively memory reads. S3 has no equivalent; each ranged GET is a full HTTP request regardless of how small the payload is.

You can't do read-ahead right now with the current code design, entries are sorted by prefix-hash not by offset. I think we have to argue with what we have right now, this is a batch system, latency doesn't matter that much here, disruption of live requests and throughput does.

This change betters the crash resiliency story so IMHO it's much better than having huge fs consumption, especially when we're going to crank up the limits allowing more parallel jobs to be handled.

Read-ahead can be implemented in the future with a bigger design refactor. 

> 
> Also, "inference dominates" holds today but isn't guaranteed in the future — faster / smaller models, better hardware, or larger batch sizes can modify the ratio. I/O buffering removes the dependency on that assumption.

Again, let's deal with what we have right now, I'm not sure we'll quickly see (if ever) see LLM inference that takes ~10ms so that an S3 Get becomes the bottleneck. We're really arguing about something that likely takes 1% of the time slot of a request.

> The fix is also not complex — fetching chunks of consecutive entries in a single ranged GET into a local buffer, then serving individual `readRequestLine` calls from that buffer. This brings the I/O cost profile back in line with the local-file behavior the code replaced.
> 
> This is about not introducing an unnecessary per-record network I/O pattern when a standard buffering approach avoids it.

Again I think doing better in regards to crash resiliency and recovery is more important, using S3 as the single source of truth can help us resume jobs much more easily as well.

acardace:
@wseaton @lioraron just added a section in https://github.com/llm-d/llm-d-batch-gateway/issues/561 on solving the concurrent S3 calls potential issue (even if I don't think there is one right now).

j-mok-dev:
I took a close look at the behavior changes here. The plan-file path still builds plans by streaming the input once, so removing the local input copy by itself looks fine and I did not find a separate correctness issue there.

However, I do think the inline issue around RetrieveRange(context.Background()) should be fixed before this lands, because it changes the cancellation / SLO expiry / SIGTERM behavior in a worse direction than the old local ReadAt path.

I also share Lior's concern that moving per-request input access from local random reads to remote ranged reads introduces a potentially expensive hot-path trade-off. I understand the argument that this PR is primarily about reducing local disk pressure / crash risk, and I’m okay treating the performance side as follow-up work, but I do think that concern should be explicitly validated, either in this PR or in the follow-up issue.

internal/processor/worker/executor.go
```diff
@@ -886,11 +871,18 @@ func mergeInferenceHeaders(headers map[string]string, sloCtx context.Context, in
 // readRequestLine reads a single plan entry from the input file, parses it, and
 // generates a batch request ID. Returns the parsed request and batch request ID
 // on success, an outputLine on parse error, or a fatal error on I/O failure.
-func readRequestLine(inputFile *os.File, entry planEntry, logger logr.Logger) (*batch_types.Request, string, *outputLine, error) {
-	buf := make([]byte, entry.Length)
-	if _, err := inputFile.ReadAt(buf, entry.Offset); err != nil {
+func (p *Processor) readRequestLine(entry planEntry, inputRef *inputFileRef, logger logr.Logger) (*batch_types.Request, string, *outputLine, error) {
```
j-mok-dev, line 874: RetrieveRange is now remote I/O, but this is using context.Background() instead of a request-scoped context. That means cancel / SLO expiry / SIGTERM cannot interrupt a slow or stuck range read, so dispatched goroutines can stay blocked here holding semaphores and delay the errCancelled / errExpired / errShutdown paths. Before this change the input access was local ReadAt, so this can be a behavior regression rather than just an optimization concern.

acardace: That's a fair point, what about using the abortCtx to detect cancellation and have the worker go through draining in case there's a network error? Would that be ok?

acardace:
> I also share Lior's concern that moving per-request input access from local random reads to remote ranged reads introduces a potentially expensive hot-path trade-off.

I don't agree with this, as I said before this is not a component delivering real-time requests, the batch gateway most important metrics are throughput and the ability to not disrupt live traffic, having ranged S3 gets before dispatching an inference request doesn't degrade these key metrics, and it improves the resiliency of the component.

Also the time taken by an S3 request is orders of magnitude smaller than the inference request (we're talking about ~10-50ms vs seconds or even minutes) so I don't think the performance argument stands here, if this would be handling live traffic then it would be a different story altogether. I honestly don't think performance is impacted at all by this patch.

internal/processor/worker/executor.go
acardace: or a simple
```
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()
rc, err := p.files.storage.RetrieveRange(ctx, inputRef.storageName, inputRef.folderName, entry.Offset, int64(entry.Length))
```

might be enough?

j-mok-dev: `requestAbortCtx` so cancel / expiry / shutdown can interrupt a slow read.

j-mok-dev:
> > I also share Lior's concern that moving per-request input access from local random reads to remote ranged reads introduces a potentially expensive hot-path trade-off.
> 
> I don't agree with this, as I said before this is not a component delivering real-time requests, the batch gateway most important metrics are throughput and the ability to not disrupt live traffic, having ranged S3 gets before dispatching an inference request doesn't degrade these key metrics, and it improves the resiliency of the component.
> 
> Also the time taken by an S3 request is orders of magnitude smaller than the inference request (we're talking about ~10-50ms vs seconds or even minutes) so I don't think the performance argument stands here, if this would be handling live traffic then it would be a different story altogether. I honestly don't think performance is impacted at all by this patch.

That makes sense, and I agree this is not really a single-request latency concern. I’m more worried about the cost / request-volume trade-off of moving from one remote fetch plus local `ReadAt` to one remote ranged read per batch request, especially for very large batches. I’m okay not treating that as a blocker here, but I think it would be good to track and validate in a follow-up issue.

acardace:
Re opening as this is still relevant.

acardace:
@wseaton or @zdtsw can you re open it?

internal/processor/worker/executor.go
evacchi: I think still needs to be addressed +1 on abortCtx. BTW, related to the abundance of contexts: https://github.com/llm-d/llm-d-batch-gateway/pull/625

acardace: addressed.

evacchi (approved):

zdtsw:
can we get the pre-commit fixed? @acardace

lioraron (changes requested):
I'm not convinced that increasing the I/O frequency, latency, and error surface is justified.
Can this mechanism be added as a configurable option?

acardace:
> I'm not convinced that increasing the I/O frequency, latency, and error surface is justified. Can this mechanism be added as a configurable option?

This addresses a condition that cause the pod to crash because of the emptyDir being filled up, I think that's more important than I/O frequency and latency. Also S3 (even more if deployed locally) is not a bottleneck for this scenario.

lioraron:
> > I'm not convinced that increasing the I/O frequency, latency, and error surface is justified. Can this mechanism be added as a configurable option?
> 
> This addresses a condition that cause the pod to crash because of the emptyDir being filled up, I think that's more important than I/O frequency and latency. Also S3 (even more if deployed locally) is not a bottleneck for this scenario.

The price of increased I/O frequency, latency, and error surface will be paid always with this implementation. The probability of emptyDir issue is debatable. This proposed mechanism should be enhanced in terms of performance and made a configurable option.

acardace:
On the async path: `Produce` is a single goroutine iterating plan entries sequentially, and it blocks on the unbuffered `outgoingRequestCh <- *item` send after each S3 read. Even though `AsyncDispatcher` submits to a queue (not blocking on inference), it still consumes items one at a time from that channel, so `Produce` can't race ahead. The S3 read rate is bounded by the queue submit throughput, not by inference latency.

Even in the worst case where submits are instantaneous: a 10ms S3 read per entry yields ~100 items/sec per worker, 2,000/sec across 20 workers. S3 handles 5,500 GETs/sec per prefix. Not a bottleneck.

On the broader concerns:

The "I/O frequency, latency, and error surface will always be paid" framing assumes the cost is material. It isn't:
- Sync path: the global semaphore (default 100) gates how many requests are in-flight at once. Since `Produce` blocks on the channel send, the S3 read for entry N+1 can't happen until entry N is dispatched and a semaphore slot is acquired. Concurrent S3 GETs per worker ≈ 1 in steady state.
- Error surface: S3 is already a hard dependency, preprocessing streams from it, finalization uploads to it. A partition that breaks `RetrieveRange` equally breaks the existing `Retrieve` and `Store` calls. The retryclient handles transient failures. No new failure domain is introduced.

On emptyDir: with `numWorkers: 20` and `emptyDir.sizeLimit: 10Gi` (both shipped defaults), input copies alone consume up to 4 GB, 40% of the budget before a single output line is written. Hitting the limit causes pod eviction, which destroys every in-flight job's partial output since recovery is workdir-based. This isn't a theoretical edge case at default config.

On "make it configurable": that means maintaining, testing, and documenting two I/O paths indefinitely. The FS fallback (`inputFile *os.File`) already exists for local/test deployments, it's preserved in this PR. New deployments using shared storage get the ranged-read path, which is the correct default.

Read-ahead across plan entries is a real optimization worth doing, but it requires re-sorting entries by offset (they're currently sorted by PrefixHash for prefix cache grouping). That's a non-trivial follow-up, tracked in #561, not a reason to block this.

lioraron:
The implementation in this PR replaces a single input download with many remote storage operations - one per plan entry = up to 50,000 GETs per batch, considerably extends the opportunity to hit i/o errors, and makes per-job dispatch throughput dependent on range-read latency. This matters for fast inference and async submission paths, and it is particularly problematic after cancel or expiry: `Produce` deliberately continues serial range reads with a detached 30-second context so it can generate every error record. At 10ms/read, a 50,000-entry drain alone is about eight minutes.

This PR also eliminates the existing local-copy mechanism as a production option. A normal batch always has `input_file_id`, so it always resolves `inputFileRef` and uses `RetrieveRange`; the `*os.File` path is only reachable without an input file ID, which preprocessing rejects.

We can consider a simpler operational alternative: set the shipped default of `numWorkers` to 2 or 4, and set the `emptyDir` default size for the bounded input and output/error files. The current defaults for `numWorkers` and `emptyDir` look arbitrary. Scaling pods has standard methodologies, while scaling internal workers isn't standard.

PR #556, opened 2026-07-06
feat(benchmarks): sim mode flow control support (PR 6 of #491)

kahilam:
## Summary

PR 6 of #491: Enable scenario 4 validation on a local Kind cluster without GPU.

- **Scenario 4 sim mode** — Already implemented in previous PRs (`setup.sh` deploys GIE EPP with flow control, `EndpointPickerConfig`, and `InferenceObjective` CRDs). This PR fixes `helm install` → `upgrade --install` for idempotency.
- **Summary table SLO completion format** — Changed from raw `completed/total` counts to per-job actual-time/target-time with ✓/✗ indicators (e.g. `A: ✓ 8m/30m | B: ✓ 15m/2h | C: ✗ 35m/24h`)
- **Infrastructure metrics: requests_waiting chart** — Added `vllm:num_requests_waiting` time-series chart as the third infrastructure metric (queue pressure indicator)

## Changes

- `benchmarks/benchmark.py`: Added `get_per_job_progress()`, `_format_slo_completion()`, `_format_duration()` helpers; track `job_completion_times` in `monitor_scenario()`; added `requestsWaitingChart` canvas and JS rendering; updated narrative to include SLO pass/fail counts
- `benchmarks/setup.sh`: EPP `helm install` → `helm upgrade --install` for sim mode; PVCs use ReadWriteOnce in sim mode for Kind compatibility

## Test plan

- [x] Python syntax validation passes
- [x] Shell syntax validation passes (`bash -n`)
- [x] Unit tests for `_format_duration()` and `_format_slo_completion()` pass
- [x] `BENCHMARK_SCENARIO=4 make benchmark-local` on Kind cluster
- [x] Full GPU benchmark generates correct SLO format in report

Closes part of #491


kahilam:
Hi@lioraron ready for review. validated on Kind (sim mode) and GPU cluster (CoreWeave). All test plan items pass.

benchmarks/benchmark.py
```diff
@@ -1084,6 +1115,52 @@ def _aggregate_phases(phases, phase_filter=None):
...
 
 
+def _format_slo_completion(result):
+    """Format batch SLO as per-job actual/target with pass/fail indicators.
+
+    Returns e.g. 'A: ✓ 8m/30m<br>B: ✓ 15m/2h<br>C: ✓ 25m/24h'
+    """
+    job_labels = {"job-a": "A", "job-b": "B", "job-c": "C"}
+    windows = {"job-a": "30m", "job-b": "2h", "job-c": "24h"}
+    window_seconds = {"job-a": 1800, "job-b": 7200, "job-c": 86400}
```
lioraron: Duplicated constants: `window_seconds` is defined identically here and in `_generate_narrative()` (line 1204). The human-readable `windows` dict is a third partial duplicate, and `completion_windows` at line 350 is yet another. If any SLO window changes, the table and narrative will disagree.

Extract to a single module-level constant, e.g.:
```python
JOB_SLO_WINDOWS = {
    "job-a": {"label": "A", "display": "30m", "seconds": 1800},
    "job-b": {"label": "B", "display": "2h", "seconds": 7200},
    "job-c": {"label": "C", "display": "24h", "seconds": 86400},
}
```

benchmarks/benchmark.py
```diff
@@ -604,6 +605,30 @@ def get_batch_progress(context, namespace):
...
+                    break
+            jobs[name] = {"completed": job_completed, "total": job_total, "status": status}
+        except Exception:
+            pass
+    return jobs
+
+
 def get_current_phase(context, namespace):
     """Get current guidellm phase from job logs."""
     try:
```
lioraron, line 643: Bare `except Exception: pass` silently swallows parse errors. If a job's log format changes (e.g. `completed=5 of 10` instead of `completed=5/10`), `int()` raises `ValueError`, the exception is swallowed, and the job is silently omitted from `job_completion_times`. Since this runs in a polling loop every 10s, a persistent format mismatch permanently suppresses that job's SLO tracking with no indication.

At minimum log at debug level:
```python
except Exception as e:
    logger.debug(f"Failed to parse progress for {name}: {e}")
```

benchmarks/benchmark.py
```diff
@@ -1084,6 +1115,52 @@ def _aggregate_phases(phases, phase_filter=None):
...
+            passed = elapsed_s <= window_seconds[job_name]
+            indicator = "✓" if passed else "✗"
+            actual = _format_duration(elapsed_s)
+            parts.append(f"{label}: {indicator} {actual}/{target}")
+        else:
+            parts.append(f"{label}: — /{target}")
+
+    if not parts:
+        if result.batch_timeline:
+            final = result.batch_timeline[-1]
```
lioraron: Dead code: the `if not parts:` block is unreachable. The loop above iterates over `job_labels` (always 3 items) and unconditionally appends to `parts` on every iteration (both the `if` and `else` branches append). So `parts` is always length 3.

If the intent was to handle the case where `result.job_completion_times` is empty, that check should be done before the loop (the caller already handles this — `generate_html_report` only calls this function when `result.job_completion_times` is truthy). Remove this block.

benchmarks/setup.sh
```diff
@@ -133,18 +133,32 @@ ${K} -n "${NAMESPACE}" create secret generic batch-gateway-secrets \
...
 ${K} -n "${NAMESPACE}" apply -f - <<EOF
 apiVersion: v1
 kind: PersistentVolumeClaim
 metadata:
   name: batch-gateway-files
 spec:
-  accessModes: [ReadWriteMany]
+  accessModes: [${PVC_ACCESS_MODE}]
+  resources:
+    requests:
```
lioraron, line 149: The inline PVC replaces `apply -f manifests/results-pvc.yaml` but the manifest file is not deleted. It's now orphaned dead code — a future contributor might edit it expecting it to be applied.

Delete `benchmarks/manifests/results-pvc.yaml`.

lioraron (approved):
Thanks!

PR #559, opened 2026-07-07
fix(processor): make job dequeue and in-flight claim atomic

### LINKED ISSUE #562: Job pickup is not atomic: crash between dequeue and in-flight claim leaves job invisible
## Problem

Job pickup in the processor is two separate Redis operations: pop from the priority queue (`PQDequeue`), then write the in-flight claim (`InFlightSet`). A processor crash between the two leaves the job in neither store — invisible to the orphan reconciler's queue/in-flight cross-reference until a full GC cycle catches it via the DB fallback path.

## Proposed solution

Perform the pop and the claim in a single Lua script so a crash at any point leaves the job either queued or claimed, never in neither.

## Note on design issue #434

#434 ("Why PQDequeue and InFlightSet are not atomic") rejected the Lua script approach partly because scripts can't use blocking `BZMPOP`. That argument no longer holds: the blocking path has no callers — the only production call site (`poller.go`) passes `timeout=0`. With the blocking path dead, the atomicity objection evaporates and the two-step design's crash window is no longer justified.

Fixed by #559.

wseaton:
Fixes #562

Job pickup was two separate Redis operations: pop from the priority queue, then write the in-flight claim. A processor crash between the two left the job in neither store, invisible to the orphan reconciler.

This replaces the two-step pickup with a single Lua script (`PQDequeueAndClaim`) that pops the queue head and records the in-flight owner atomically. A crash at any point now leaves the job either queued or claimed.

Includes unit tests (miniredis) and a regression guard that races concurrent claimers against a reconciler-style observer to pin the invariant.

### DIFF
--- a/internal/database/redis/redis_pq_dequeue_claim.lua
+++ b/internal/database/redis/redis_pq_dequeue_claim.lua
@@ -18,7 +18,8 @@
 -- KEYS[2] = in-flight hash
 -- ARGV[1] = in-flight entry JSON (marshaled by the caller)
 -- Returns the popped member JSON, false when the queue is empty, or an error
--- if the popped member is corrupt.
+-- if the popped member is corrupt (the member stays dropped) or the claim
+-- write fails (the member is restored).
 
 local popped = redis.call("ZPOPMIN", KEYS[1])
 if #popped == 0 then
@@ -28,10 +29,15 @@ local member = popped[1]
 local score = popped[2]
 local ok, decoded = pcall(cjson.decode, member)
 if not ok or type(decoded) ~= "table" or type(decoded.id) ~= "string" or decoded.id == "" then
-    -- Redis scripts are not rolled back on error, so restore the popped member
-    -- before surfacing the corrupt queue entry to the caller.
-    redis.call("ZADD", KEYS[1], score, member)
-    return redis.error_reply("corrupt priority queue member")
+    -- Poison pill: restoring the member would put it back at the head of the
+    -- queue and starve every job behind it. Leave it dropped.
+    return redis.error_reply("corrupt priority queue member (dropped): " .. member)
 end
-redis.call("HSET", KEYS[2], decoded.id, ARGV[1])
+local hsetOk, hsetErr = pcall(redis.call, "HSET", KEYS[2], decoded.id, ARGV[1])
+if not hsetOk then
+    -- Script effects are not rolled back on error; restore the member so the
+    -- job is not lost between the queue and the in-flight hash.
+    redis.call("ZADD", KEYS[1], score, member)
+    return redis.error_reply("in-flight claim failed, member restored: " .. tostring(hsetErr))
+end
 return member
--- a/internal/database/redis/redis_pq_dequeue_claim.lua
+++ b/internal/database/redis/redis_pq_dequeue_claim.lua
@@ -17,18 +17,21 @@
 -- KEYS[1] = priority queue sorted set
 -- KEYS[2] = in-flight hash
 -- ARGV[1] = in-flight entry JSON (marshaled by the caller)
--- Returns the popped member JSON, or false when the queue is empty.
+-- Returns the popped member JSON, false when the queue is empty, or an error
+-- if the popped member is corrupt.
 
 local popped = redis.call("ZPOPMIN", KEYS[1])
 if #popped == 0 then
     return false
 end
 local member = popped[1]
+local score = popped[2]
 local ok, decoded = pcall(cjson.decode, member)
-if not ok or type(decoded) ~= "table" or decoded.id == nil then
-    -- Corrupt member: script effects are not rolled back on error, so raising
-    -- would eat the popped member. Return it unclaimed; the caller logs it.
-    return member
+if not ok or type(decoded) ~= "table" or type(decoded.id) ~= "string" or decoded.id == "" then
+    -- Redis scripts are not rolled back on error, so restore the popped member
+    -- before surfacing the corrupt queue entry to the caller.
+    redis.call("ZADD", KEYS[1], score, member)
+    return redis.error_reply("corrupt priority queue member")
 end
 redis.call("HSET", KEYS[2], decoded.id, ARGV[1])
 return member
--- a/internal/database/redis/redis_pq_dequeue_claim.lua
+++ b/internal/database/redis/redis_pq_dequeue_claim.lua
@@ -1,0 +1,34 @@
+-- Copyright 2026 The llm-d Authors
+
+-- Licensed under the Apache License, Version 2.0 (the "License");
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+
+--     http://www.apache.org/licenses/LICENSE-2.0
+
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+
+-- PQDequeueAndClaim: atomically pop the highest-priority (lowest score) member
+-- from the priority queue and record an in-flight claim for it.
+-- KEYS[1] = priority queue sorted set
+-- KEYS[2] = in-flight hash
+-- ARGV[1] = in-flight entry JSON (marshaled by the caller)
+-- Returns the popped member JSON, or false when the queue is empty.
+
+local popped = redis.call("ZPOPMIN", KEYS[1])
+if #popped == 0 then
+    return false
+end
+local member = popped[1]
+local ok, decoded = pcall(cjson.decode, member)
+if not ok or type(decoded) ~= "table" or decoded.id == nil then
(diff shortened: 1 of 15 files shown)


lioraron (commented):
Good initiative, thanks @wseaton.

A few suggestions before a deeper review:

1. Can you please open an issue describing the problem, the proposed solution, and associate the PR with the issue?

2. The aspect focused on in this PR was previously evaluated in design issue #434 - in the section "Why PQDequeue and InFlightSet are not atomic", and I agree that the evaluation in that paragraph should be overturned: An important justification for the Lua script approach is that `PQDequeue`'s blocking path (`BZMPop`) has no callers — the only production call site (poller.go:52) passes `timeout=0`. This justification should be captured in the issue requested above since it eliminates the main argument against using a Lua script (which can't use `BZMPOP`).

3. Consider modifying `PQDequeue` itself to perform the atomic dequeue+claim (via the Lua script), and removing the blocking path, instead of adding `PQDequeueAndClaim`? This keeps the interface clean and avoids accumulating dead code.

internal/database/redis/redis_pq_dequeue_claim.lua
```diff
@@ -0,0 +1,37 @@
...
+    return false
+end
+local member = popped[1]
+local score = popped[2]
+local ok, decoded = pcall(cjson.decode, member)
+if not ok or type(decoded) ~= "table" or type(decoded.id) ~= "string" or decoded.id == "" then
+    -- Redis scripts are not rolled back on error, so restore the popped member
+    -- before surfacing the corrupt queue entry to the caller.
+    redis.call("ZADD", KEYS[1], score, member)
+    return redis.error_reply("corrupt priority queue member")
```
lioraron: **Bug — corrupt member creates a livelock (poison pill)**

Restoring the corrupt member with `ZADD` at the same score makes it the head of the queue again. Every subsequent `PQDequeueAndClaim` call pops the same corrupt member, fails validation, restores it, and returns an error. Valid jobs behind it are starved indefinitely.

The old `PQDequeue`/`ZMPop` path was safer: it left the corrupt member permanently removed after the pop, so processing continued with the next job.

**Suggested fix:** drop the corrupt member instead of restoring it. Log the corruption (via the error reply) and let it stay removed:
```lua
-- don't ZADD it back; return the error so the Go side logs it
return redis.error_reply("corrupt priority queue member (dropped): " .. member)
```

internal/database/redis/redis_pq_dequeue_claim.lua
```diff
@@ -0,0 +1,37 @@
...
+end
+redis.call("HSET", KEYS[2], decoded.id, ARGV[1])
+return member
```
lioraron, line 43: **Bug — no rollback of ZPOPMIN if HSET fails**

`redis.call()` aborts the script on error, but ZPOPMIN has already executed and Redis Lua scripts don't transact — the job vanishes from both the queue and the in-flight hash.

The corrupt-member path above correctly restores via ZADD before returning the error, but this HSET failure path has no equivalent guard.

In the old code, `InFlightSet` failure was non-fatal — the worker logged it and continued processing the job. Here the job becomes temporarily invisible until the GC reconciler detects the orphan and re-enqueues it.

**Suggested fix:** use `pcall` for HSET and restore the member on failure:
```lua
local hsetOk, hsetErr = pcall(redis.call, "HSET", KEYS[2], decoded.id, ARGV[1])
if not hsetOk then
    redis.call("ZADD", KEYS[1], score, member)
    return redis.error_reply("HSET failed, member restored: " .. tostring(hsetErr))
end
```

internal/database/redis/redis_queue.go
```diff
@@ -195,6 +196,55 @@ func (c *ExchangeDBClientRedis) PQDequeue(ctx context.Context, timeout time.Dura
...
+		return nil, fmt.Errorf("failed to marshal in-flight entry: %w", err)
+	}
+
+	cctx, ccancel := context.WithTimeout(ctx, c.timeout)
+	defer ccancel()
+	raw, err := redisScriptPQDequeueClaim.Run(cctx, c.redisClient,
+		[]string{priorityQueueKeyName, inFlightKeyName}, entryData).Result()
+	if err != nil {
+		// The script returns false on an empty queue, which go-redis maps to Nil.
+		if errors.Is(err, goredis.Nil) {
```
lioraron, line 159: **Regression — error logging and health check dropped**

The old `PQDequeue` (lines 155–162 on main) logged unexpected Redis errors at ERROR level and called `redisClientChecker.Check()` to detect failover to a read-only replica:
```go
logger.Error(err, "PQDequeue: B/ZMPop failed")
cerr := c.redisClientChecker.Check(ctx)
```

This new method just wraps and returns the error. The polling loop in `worker.go` silently swallows it (`if task == nil || err != nil { continue }`). Result: if Redis becomes unreachable or fails over, the processor appears healthy (no crash, no log) but processes zero jobs indefinitely.

**Suggested fix:** add the same error logging and `redisClientChecker.Check()` call before returning, matching `PQDequeue`'s behavior.

internal/database/redis/redis_queue_claim_test.go
```diff
@@ -0,0 +1,212 @@
...
+				task, err := exch.PQDequeueAndClaim(ctx, "proc-1")
+				if err == nil {
+					t.Fatalf("PQDequeueAndClaim() err=nil, want error (task=%v)", task)
+				}
+
+				members, err := minirds.ZMembers(queueKey)
+				if err != nil {
+					t.Fatalf("ZMembers() err=%v", err)
+				}
+				if len(members) != 1 || members[0] != tc.member {
```
lioraron: **Missing test — valid job behind corrupt member is unreachable (livelock)**

This test asserts the corrupt member is restored to the queue, but doesn't verify what happens on the *next* call. Add a subcase that enqueues a valid job at a higher score behind the corrupt one, calls `PQDequeueAndClaim` twice, and asserts the valid job is returned on the second call.

With the current ZADD-restore behavior, this test would fail — the corrupt member is always re-popped first — which directly exposes the poison pill from the Lua script comment above.

internal/database/redis/redis_queue_claim_test.go
```diff
@@ -0,0 +1,212 @@
...
+				if err != nil {
+					t.Fatalf("InFlightGetAll() err=%v", err)
+				}
+				if len(entries) != 0 {
+					t.Fatalf("corrupt member produced in-flight entries: %v", entries)
+				}
+			})
+		}
+	})
+}
```
lioraron, line 259: **Missing test — HSET failure after ZPOPMIN loses the job**

There's no test for what happens when the in-flight HSET fails after ZPOPMIN has already popped the member. You can trigger this in miniredis by setting the in-flight key to a non-hash type before the call:
```go
t.Run("HSET failure does not lose job", func(t *testing.T) {
    minirds, exch := newExchangeClient(t)
    ctx := context.Background()
    // ... enqueue a valid job ...
    // Sabotage the in-flight key so HSET fails with WRONGTYPE
    minirds.Set("llmd_batch:inflight", "not-a-hash")
    task, err := exch.PQDequeueAndClaim(ctx, "proc-1")
    // Should error, but the job must still be in the queue
    // (currently it's lost)
})
```
This test would fail today, directly validating the `pcall` fix suggested on the Lua script.

wseaton:
Addressed the full review:

1.  opened #562, linked via `Fixes`. It captures the overturning of the #434 evaluation — the blocking `BZMPOP` path had no callers (the poller always passed `timeout=0`), which removes the main argument against the Lua script.
2. rather than folding the claim into `PQDequeue`, I removed `PQDequeue` entirely and kept `PQDequeueAndClaim`, the name reflects that dequeue now also writes ownership. The dead blocking path, `ZMPop`/`BZMPop` usage, and the timeout-semantics tests went with it. A side benefit: the queue tests now also run against miniredis (`ZPOPMIN` is supported where `ZMPOP` isn't); verified against real Redis 7 as well.
3. All four inline findings fixed with tests

acardace:
@wseaton just a note, when this is fixed we should also fix the garbage collector which right now tries to detect stale jobs (which won't exist anymore) every time it fires. Probably we have to rethink the recovery mechanism.

internal/database/redis/redis_queue.go
```diff
@@ -130,69 +130,60 @@ func (c *ExchangeDBClientRedis) PQDelete(ctx context.Context, item *db_api.Batch
...
-		if time.Since(c.idleLogLast) >= c.idleLogFreq {
-			logger.Info("PQDequeue: no items")
-			c.idleLogLast = time.Now()
+		// The polling loop treats dequeue errors as "no work" and keeps
+		// spinning, so log here and probe the connection to surface Redis
+		// outages and read-only failovers that would otherwise be silent.
+		logger.Error(err, "PQDequeueAndClaim: script failed")
+		if cerr := c.redisClientChecker.Check(ctx); cerr != nil {
+			logger.Error(cerr, "PQDequeueAndClaim: ClientCheck failed")
 		}
```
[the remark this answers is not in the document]
wseaton: this is reasonable IMHO

wseaton:
Willing to close in favor of https://github.com/llm-d/llm-d-batch-gateway/issues/572

PR #560, opened 2026-07-07
fix(benchmarks): send x-gateway-inference-objective header for interactive traffic

lioraron:
## Summary

- Interactive guidellm traffic was missing the `x-gateway-inference-objective` header, so in scenario 4 (flow control) requests landed in the default priority band instead of the priority-100 `interactive-default` band — the router couldn't distinguish interactive from batch
- Switch guidellm from `--target` to `--backend` JSON config with `extras.headers` to send `x-gateway-inference-objective: interactive-default` on all interactive requests (harmless in scenarios without flow control)

## Test plan

- [x] `BENCHMARK_SCENARIO=0 make benchmark-local` — verify guidellm still runs correctly with the new `--backend` JSON format
- [x] `BENCHMARK_SCENARIO=4 make benchmark-local` — verify interactive traffic gets priority-100 treatment (check Router flow control metrics show interactive requests in the high-priority band)

### DIFF
--- a/benchmarks/benchmark.py
+++ b/benchmarks/benchmark.py
@@ -460,11 +460,7 @@ def start_interactive_traffic(cfg, namespace):
         f"burst@{cfg.burst_rate}/s for {cfg.burst_seconds}s, "
         f"idle@{cfg.idle_rate}/s for {cfg.idle_seconds}s")
 
-    backend_json = json.dumps({
-        "kind": "openai_http",
-        "target": cfg.target,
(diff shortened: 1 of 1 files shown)


yizhaodev (approved):

lioraron:
@yizhaodev Thanks for raising this!

@kahilam After this is merged, let's rerun scenario 4? In the latest results it appears that scenario 4 doesn't protect interactive performance better than scenario 3. Maybe this addition will help.

PR #566, opened 2026-07-08
fix(s3): use single bucket with key prefix instead of per-tenant buckets

### LINKED ISSUE #565: S3 backend uses per-tenant buckets, hitting AWS bucket limit at scale
## Problem

The S3 storage backend currently uses the `folderName` parameter (derived from tenant ID via SHA256 hash) as the **S3 bucket name**. This means every tenant gets its own S3 bucket, auto-created on first use.

**AWS S3 hard limits:**
- Default: **100 buckets** per account
- After support request: **1000 buckets** max per account
- These are hard limits that cannot be increased further

This means the system breaks once tenant count exceeds 1000 (or 100 without limit increase).

yizhaodev:
## Why is this PR needed?

The S3 backend uses `folderName` (tenant ID hash) as the S3 **bucket name**, creating one bucket per tenant. AWS S3 limits accounts to 100 buckets by default (1000 max after support request), so the system breaks once tenant count exceeds this hard limit.

Fixes #565

## What does this PR do?

Switches the S3 backend from per-tenant buckets to a **single configured bucket** with tenant isolation via S3 key prefixes.
- before: bucket=`t-<hash>`, key=`prefix/fileName`
- after: bucket=`config.bucket`, key=`prefix/t-<hash>/fileName`

The PR include:
- Adds required `bucket` field to S3 Config
- Simplifies `ensureBucket` — called once at startup instead of per-operation
- Updates Helm values/templates and local dev configs with new `bucket` field
- No changes to FS backend

## How was this tested?

- [x] Unit tests added/updated/verified (1057 pass, 0 fail)

## Checklist

## Related Issues

Fixes #565

### DIFF
--- a/internal/files_store/s3/client.go
+++ b/internal/files_store/s3/client.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -56,13 +55,13 @@ type uploaderAPI interface {
 }
 
 // Client implements api.BatchFilesClient using S3 storage.
-// The folderName parameter in Store/Retrieve/Delete is used as the S3 bucket name.
+// All objects are stored in a single configured bucket. The folderName parameter
+// in Store/Retrieve/Delete is used as part of the S3 key for tenant isolation.
 type Client struct {
-	s3Client         s3API
-	uploader         uploaderAPI
-	prefix           string
-	autoCreateBucket bool
-	knownBuckets     sync.Map // caches bucket names that have been verified to exist
+	s3Client s3API
+	uploader uploaderAPI
+	prefix   string
+	bucket   string
 }
 
 var _ api.BatchFilesClient = (*Client)(nil)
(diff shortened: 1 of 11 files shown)


lioraron (commented):
The single-bucket approach correctly addresses the S3 100-bucket limit.

internal/files_store/s3/client.go
```diff
@@ -56,20 +55,21 @@ type uploaderAPI interface {
 }
 
 // Client implements api.BatchFilesClient using S3 storage.
```
lioraron, line 57: Now that `folderName` is used as a key prefix (not a bucket name), the comment in `internal/util/com/common.go:74-88` is stale — it references "S3 bucket naming requirements (63 characters max)" and "S3 maximum" as the rationale for the 63-char limit. Since S3 object keys allow up to 1024 bytes, the bucket-naming justification no longer applies. Worth updating the comments in `GetFolderNameByTenantID` to reflect the new semantics (e.g., "filesystem path length constraints" or just drop the S3-specific rationale).

lioraron (approved):

internal/files_store/s3/client.go
yizhaodev: comments updated

PR #570, opened 2026-07-09
test: Allow testAPIKey to be configured via TEST_API_KEY env var

mynhardtburger:
## Why is this PR needed?

To enable testing from outside of a the cluster using a provided API key

## What does this PR do?

Allows the test api key to be configured via the TEST_API_KEY env var when executing the e2e tests.
Because the testApiserverObsURL is only accessible from inside the cluster, make the test's readiness check optional.

## How was this tested?

## Checklist

- [ ] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)


yizhaodev (approved):

test/e2e/e2e_test.go
```diff
@@ -109,7 +110,12 @@ func TestE2E(t *testing.T) {
 	testExchangeClientType = detectExchangeClientType(t)
 	t.Logf("DB client type: %s, exchange client type: %s", testDBClientType, testExchangeClientType)
 
-	waitForReady(t, testApiserverObsURL, 30*time.Second)
+	if resp, err := http.Get(testApiserverObsURL + "/ready"); err == nil {
+		resp.Body.Close()
+		waitForReady(t, testApiserverObsURL, 30*time.Second)
+	} else {
+		t.Logf("observability endpoint %s not reachable, skipping readiness gate: %v", testApiserverObsURL, err)
```
yizhaodev: @mynhardtburger how about add a flag for external testing
testSkipReadiness = getEnvOrDefault("TEST_SKIP_READINESS", "false")

test/e2e/e2e_test.go
```diff
@@ -109,7 +110,12 @@ func TestE2E(t *testing.T) {
...
+	}
```
lioraron, line 129: Skipping the readiness gate entirely when the obs URL is unreachable means tests run with no readiness guarantee — the apiserver may not be ready yet, causing flaky failures.

The obs endpoint is cluster-internal, but `testApiserverURL` (the data endpoint) IS reachable from outside. Consider probing that instead:

```go
waitForReady(t, testApiserverURL, 30*time.Second)
```

This way tests always wait for the server to be ready, regardless of where they run from. You can additionally probe the obs URL when it's reachable.

test/e2e/e2e_test.go
```diff
@@ -109,7 +110,12 @@ func TestE2E(t *testing.T) {
 	testExchangeClientType = detectExchangeClientType(t)
 	t.Logf("DB client type: %s, exchange client type: %s", testDBClientType, testExchangeClientType)
 
-	waitForReady(t, testApiserverObsURL, 30*time.Second)
+	if resp, err := http.Get(testApiserverObsURL + "/ready"); err == nil {
```
lioraron, line 124: When the obs URL IS reachable, this single `http.Get` succeeds and then `waitForReady` is called on line 116 — which does the exact same `http.Get` again as its first iteration. The initial probe is redundant in the happy path. If you switch to probing `testApiserverURL` as the primary readiness gate (see other comment), this duplication goes away.

lioraron (commented):
Review comments on the readiness gate change.

j-mok-dev:
I think this skip may be broader than necessary. `TestDispatcher` already handles the "dispatcher not deployed" case, so could we keep the common API / batch lifecycle tests running and only skip the specific subtests that are incompatible with async dispatch?

test/e2e/e2e_test.go
mynhardtburger: Added a `waitForServerUp` helper which tests for any response (not specifically an OK response like `waitForReady`) and used it to query `testApiserverURL`.

test/e2e/e2e_test.go
mynhardtburger: Duplicate GET call avoided by testing the response code on the initial call used to determine the reach-ability. If not OK then only poll via waitForReady.

mynhardtburger:
> I think this skip may be broader than necessary. `TestDispatcher` already handles the "dispatcher not deployed" case, so could we keep the common API / batch lifecycle tests running and only skip the specific subtests that are incompatible with async dispatch?

@j-mok-dev :
Added `skipIf` to skip subtests at a more granular level. I'm not very familiar with which tests are truly incompatible with async dispatch, so a second set of eyes would be appreciated.

j-mok-dev:
> > 
> 
> @j-mok-dev : Added `skipIf` to skip subtests at a more granular level. I'm not very familiar with which tests are truly incompatible with async dispatch, so a second set of eyes would be appreciated.

The skipped tests look right for async.

Two Batches subtests still look async-incompatible with the current dispatcher deploy (only `sim-model` is registered, and there is no sync PerModelMaxConcurrency saturation):
- `Batches/MultiModel`: uses `sim-model-b`, which is stripped in `processor-async-values.yaml`
- `Batches/Expiration`: depends on a blocker batch saturating sync dispatch slots
I'd skip those two as well when `dispatcherDeployed` is true.

mynhardtburger:
> > > 
> > 
> > 
> > @j-mok-dev : Added `skipIf` to skip subtests at a more granular level. I'm not very familiar with which tests are truly incompatible with async dispatch, so a second set of eyes would be appreciated.
> 
> The skipped tests look right for async.
> 
> Two Batches subtests still look async-incompatible with the current dispatcher deploy (only `sim-model` is registered, and there is no sync PerModelMaxConcurrency saturation):
> 
>     * `Batches/MultiModel`: uses `sim-model-b`, which is stripped in `processor-async-values.yaml`
> 
>     * `Batches/Expiration`: depends on a blocker batch saturating sync dispatch slots
>       I'd skip those two as well when `dispatcherDeployed` is true.

I've made the `Batches/MultiModel` and `Batches/Expiration` subtests also conditional on `testDispatcherDeployed`

j-mok-dev (approved):
LGTM

vishbhat (approved):

PR #571, opened 2026-07-09
feat(otel): add fine-grained processor pipeline spans

### LINKED ISSUE #489: [Feature]: [Observability] Add fine-grained OpenTelemetry spans for the processor pipeline (ingestion, execution, per-model dispatch)
### Problem Statement

The processor currently emits a single coarse `process-batch` span covering the entire job lifecycle, plus detached spans for finalize / re-enqueue / cancel paths.

The main work stages inside `process-batch` — ingestion/plan-build and execution/dispatch — have **no child spans**, so a trace cannot show where time is spent within a batch or which model's dispatch is slow.

### Proposed Solution

1. `ingest-and-plan` — wrap the ingestion/plan-build in `preprocessor.go` (around the existing `planBuildStart` block). Attributes: input line count, per-model entry counts, size bucket.
2. `execute-job` — wrap `executeJob`. Attributes: number of models, total requests, global concurrency ceiling.
3. `process-model` — per-model child span in `processModel`. Attributes: `gen_ai.request.model` / model name, in-flight count, request count, success/failure counts.
4. (optional, sampled) `execute-request` — very high volume; only add behind a ratio sampler or skip in favor of metrics, to avoid trace explosion on large batches.

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

madhugoutham:
## Why is this PR needed?

The processor has a single coarse `process-batch` span. The main work stages (ingestion and execution) have no child spans, so traces cannot show where time is spent within a batch or which model's dispatch is slow.

## What does this PR do?

Adds three child spans to `process-batch`:
- `ingest-and-plan`: covers ingestion/plan-build with attributes (line count, model count, rejected count, size bucket)
- `execute-job`: covers execution/dispatch with attributes (model count, total requests)
- `process-model`: per-model child of `execute-job` with attributes (model name, request
count), covering both sync and async dispatch paths

Non-terminal errors (system failures) are recorded with `RecordError` + `SetStatus(codes.Error)` on all spans. Expected terminal states (SLO expiry, user cancel, shutdown) are not marked as errors, matching the existing `process-batch` convention. Attribute keys use constants in `internal/util/otel/otel.go` and `semconv.GenAiRequestModel` from the OTel semantic conventions package.

## How was this tested?

- Verified span parentage with in-memory exporter (process-model nests under execute-job, not process-batch)

## Checklist

## Related Issues

Fixes #489


internal/processor/worker/job_runner.go
```diff
@@ -168,7 +168,14 @@ func (p *Processor) runJob(ctx context.Context, params *jobExecutionParams) {
...
 
 	// ingestion: pre-process job (rejects unregistered-model requests early)
-	if err := p.preProcessJob(ctx, sloCtx, userCancelCtx, params.jobInfo); err != nil {
+	ingestCtx, ingestSpan := uotel.StartSpan(ctx, "ingest-and-plan")
+	err = p.preProcessJob(ingestCtx, sloCtx, userCancelCtx, params.jobInfo)
+	if err != nil && !errors.Is(err, errExpired) && !errors.Is(err, errCancelled) && !errors.Is(err, errShutdown) {
+		ingestSpan.RecordError(err)
+		ingestSpan.SetStatus(codes.Error, "pre-process failed")
+	}
+	ingestSpan.End()
```
lioraron, line 177: The same error is now recorded on both `ingestSpan` (line 173-174) and the parent `span` (lines 175-176). Same for `execSpan` vs `span` in the execution block below (lines 209-210 vs 204-205). This produces duplicate error events in traces.

Since the child spans now carry the error, consider removing the `span.RecordError` / `span.SetStatus` from the parent for these two cases — the parent span already shows the error via its child. Or if you intentionally want errors at both levels (common in OTel), add a brief comment noting it's deliberate.

lioraron (commented):
Good work; a few items below.

internal/processor/worker/executor.go
```diff
@@ -180,6 +186,10 @@ func (ep *executionProgress) counts() *openai.BatchRequestCounts {
 // SLO expiry or SIGTERM. Its sole purpose is to let the drain phase distinguish user cancel from
 // SLO expiry.
 func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx context.Context, params *jobExecutionParams) (*openai.BatchRequestCounts, error) {
```
lioraron, line 188: This `trace.ContextWithSpan` injection is correct but non-obvious — it grafts the execute-job span onto the `requestAbortCtx` cancellation tree so that `process-model` spans become children of `execute-job` rather than `process-batch`.

The comment above explains the *what*, but consider also noting *why* this is needed: `requestAbortCtx` is derived from `sloCtx` (not from `ctx`/execCtx), so it doesn't naturally carry the execute-job span. Something like:

```go
// requestAbortCtx is derived from sloCtx (not ctx), so it doesn't carry
// the execute-job span. Graft it so process-model spans nest correctly.
```

internal/processor/worker/executor.go
```diff
@@ -279,10 +294,13 @@ func (p *Processor) executeJob(ctx, sloCtx, userCancelCtx, requestAbortCtx conte
 		// This ensures the first real error reaches errCh before any context.Canceled
 		// from other models whose contexts were cancelled by requestAbortFn.
 		go func(safeModelID, modelID string) {
+			modelCtx, modelSpan := uotel.StartSpan(requestAbortCtx, "process-model",
+				trace.WithAttributes(semconv.GenAiRequestModel(modelID)),
+			)
```
lioraron, line 299: The `process-model` span is started from `requestAbortCtx` and the returned `modelCtx` is passed as the `requestAbortCtx` parameter to `processModel`/`processModelAsync`. This works correctly — `modelCtx` inherits cancellation from `requestAbortCtx` via the OTel `Start` wrapper, and the `SetAttr(requestAbortCtx, ...)` calls inside those functions (lines 448, 590) correctly target the `process-model` span.

One edge to verify: if `requestAbortCtx` is already cancelled when `StartSpan` is called (e.g., a sibling model called `requestAbortFn`), the span is still created and ended normally — but the goroutine will exit immediately via `requestAbortCtx.Err()` checks in the dispatch loop. This is fine, just noting for completeness.

internal/util/otel/otel.go
```diff
@@ -47,6 +47,11 @@ const (
 	AttrRequestTotal     = "batch.request.total"
 	AttrRequestCompleted = "batch.request.completed"
 	AttrRequestFailed    = "batch.request.failed"
+	AttrModelCount       = "batch.model.count"
+	AttrRequestCount     = "batch.request.count"
+	AttrInputLineCount   = "batch.input.line_count"
+	AttrRejectedCount    = "batch.input.rejected_count"
```
lioraron, line 53: `AttrRequestTotal` ("batch.request.total") is now used on two different span levels with slightly different semantics:
- On the `process-batch` span: final completed+failed count (set after execution)
- On the `execute-job` span (new): total input line count from `modelMap.LineCount` (set before execution)

These happen to be the same value, but the semantic is different ("planned" vs "completed"). Consider using a distinct attribute like `AttrInputRequestTotal` for the execute-job span to avoid confusion when reading traces, or just reuse `AttrInputLineCount` since it's already defined and means the same thing here.

internal/processor/worker/job_runner.go
madhugoutham: Removed `span.RecordError `/ `span.SetStatus` from the parent process-batch span for both the ingest and execute error paths the child spans (ingestSpan, execSpan) now own the error recording. Parent still records errors for things that happen outside the child spans (status transition, finalization).

internal/processor/worker/executor.go
madhugoutham: Agreed updated the comment to explain it

internal/util/otel/otel.go
madhugoutham: swapped to `AttrInputLineCount` on the execute-job span since it represents planned input lines, not final completed+failed counts. `AttrRequestTotal` now stays exclusively on the parent process-batch span (via `setRequestCountAttrs`)

lioraron (approved):

Issue #557, opened 2026-07-07
enh: `processModelAsync` is not processing asynchronously


evacchi:
### Problem Statement

`processModelAsync` (`internal/processor/worker/executor.go:546`) was introduced in PR #458 as the async dispatch path for llm-d-async integration. 

The methods glues together `llm-d-batch-gateway` with `llm-d-async`, hence _the name_; however, despite this name, it executes **synchronously** from the worker's perspective: the worker blocks until all results arrive, holding its worker semaphore token the entire time.

Thus, with `NumWorkers=N`, at most N batch jobs can be in-flight, and each worker is **idle-waiting** on results for the entire inference duration. While `llm-d-async` _actually processes requests asynchronously_ the processor-side concurrency model is identical to the sync path with one worker being blocked per job.

### Related comments

This was identified during review by @acardace ([comment](https://github.com/llm-d/llm-d-batch-gateway/pull/458#discussion_r2185052244)):

> Doesn't this make it synchronous? Shouldn't we have a different worker coming from a different pool waiting for the results and producing the output files?
>
> My worry is that this basically is the same as the sync counterpart as it holds up the worker which keeps the semaphore token until this is done, effectively this counters the benefits of using async IIUC because `runJob` doesn't decrease the semaphore until `executeJob` is done.

And by @yizhaodev ([comment](https://github.com/llm-d/llm-d-batch-gateway/pull/458#discussion_r2170994805)):

> It seems `AsyncInferenceClient.Generate()` will block until result received? So the number of in-flight async requests is capped by the processor's worker concurrency limit, we won't really benefit from async mode?

Both observations are correct. The flow is:

1. **`acquire()`** — worker takes a `p.tokens` semaphore slot (worker.go:418)
2. **`runJob()`** — holds the slot for the entire job lifetime; `release()` is deferred (job_runner.go:80)
3. **`executeJob()` → `processModelAsync()`** — Phase 1 submits all requests to Redis (fast), but Phase 2 blocks calling `GetResult()` in a loop until every result comes back (executor.go:658–680)
4. **`release()`** — only fires after `runJob` returns

### Root cause

The `InferenceClient` interface was designed around synchronous HTTP request/response:

```go
Generate(ctx context.Context, req *GenerateRequest) (*GenerateResponse, *ClientError)
```

[...]

PR #573, opened 2026-07-10
fix: size async result channel to plan length

madhugoutham:
## Why is this PR needed?

Under the async submit-then-collect path, results can arrive before collect starts reading. The per-job results channel used a fixed buffer of 100. When more than 100 results arrived first, `resultDispatcher` dropped them (`default:` + "Dropping result"). Those results were already removed from Redis and from the waiter map, so collect waited until abort and the job looked like it expired. 

## What does this PR do?

Sizes the per-job results channel to the plan length (`len(entries)`) via `ClientForWithCapacity`, so submit-before-collect cannot fill a fixed buffer of 100. Keeps the non-blocking send and drop only as a defensive fallback, so one full channel does not stall the shared dispatcher for other jobs on the same pool.

## How was this tested?

Unit and integration: `go test ./pkg/clients/inference/ ./internal/processor/worker/ -count=1`, `go test ./... -count=1`, `go test ./test/regression/... -count=1`, `go test -tags=integration ./... -count=1`.

Regression coverage: `TestResultDispatcher_SubmitBeforeCollectBuffer` shows undersized buffer drops results, and buffer sized to the batch delivers all.

## Checklist

## Related Issues

None. Standalone correctness fix.

### DIFF
--- a/internal/processor/worker/executor.go
+++ b/internal/processor/worker/executor.go
@@ -566,7 +566,10 @@ func (p *Processor) processModelAsync(
 
 	logger.V(logging.INFO).Info("Processing requests for model (async)", "numEntries", len(entries))
 
-	asyncClient := p.asyncInference.ClientFor(modelID)
+	// Size the per-job results buffer to the plan length so results that
+	// arrive during submit-before-collect cannot fill a fixed buffer of 100
+	// and be silently dropped by the shared resultDispatcher.
+	asyncClient := p.asyncInference.ClientForWithCapacity(modelID, len(entries))
 	if asyncClient == nil {
 		logger.V(logging.INFO).Info("No async client for model, draining as model_not_found")
 		p.drainUnprocessedRequests(
--- a/pkg/clients/inference/async_inference_client_impl.go
+++ b/pkg/clients/inference/async_inference_client_impl.go
@@ -106,6 +106,10 @@ func (d *resultDispatcher) processNext(ctx context.Context) {
 		RequestID: result.ID,
 		Response:  []byte(result.Payload),
 	}
+	// Non-blocking send keeps the shared dispatcher from stalling when one
+	// job's channel is full. Callers must size the per-job buffer to cover
+	// submit-before-collect (see ClientForWithCapacity); this default branch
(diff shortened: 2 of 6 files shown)


internal/processor/worker/executor.go
```diff
@@ -566,7 +566,10 @@ func (p *Processor) processModelAsync(
 
 	logger.V(logging.INFO).Info("Processing requests for model (async)", "numEntries", len(entries))
 
-	asyncClient := p.asyncInference.ClientFor(modelID)
+	// Size the per-job results buffer to the plan length so results that
+	// arrive during submit-before-collect cannot fill a fixed buffer of 100
+	// and be silently dropped by the shared resultDispatcher.
+	asyncClient := p.asyncInference.ClientForWithCapacity(modelID, len(entries))
```
j-mok-dev, line 572: I don't think `len(entries)` is a safe bound here.
`processModelAsync()` still does full submit before collect, so this change allows one job to buffer every result until collect starts. In the current async path that means we're no longer limited by the old fixed buffer; we're now effectively scaling the per-job in-memory queue with batch size. Since the dispatcher enqueues full responses rather than lightweight metadata, a large batch can turn this correctness fix into a significant memory-growth / OOM risk.
I think the safer fix is to keep the result channel bounded and overlap submit + collect so responses are drained and written as they arrive.

yizhaodev: agree

if fs/s3 is down causing the channel to be full, I would rather drop message than oom in processor pod, 

instead, we can consider adding metric/log for dropped message for devops visibility.

madhugoutham: Thanks @j-mok-dev @yizhaodev . Validated the concern. processModelAsync submits the complete plan before collecting results, so sizing the channel to len(entries) is not a safe memory bound. I agree that the right direction is bounded buffering with collection overlapping submission. That overlaps the broader JobRunner/collector work tracked in #557, so I’m closing this PR rather than add a competing partial redesign here. I’ll align on scope before opening a follow-up.

madhugoutham:
Closing this PR to revisit the approach.

PR #574, opened 2026-07-10
test(e2e): add trailing newline edge case for batch input

yizhaodev:
## What does this PR do?

E2E test verifying input JSONL files with trailing newlines (one or two) produce correct request counts and don't cause parse errors.

## How was this tested?

## Checklist

## Related Issues

None. Standalone correctness fix.

### DIFF
--- a/test/e2e/batches_test.go
+++ b/test/e2e/batches_test.go
@@ -42,6 +42,7 @@ func testBatches(t *testing.T) {
 		t.Run("IdempotentRetry", doTestCancelIdempotentRetry)
 		t.Run("TerminalBatchRejected", doTestCancelTerminalBatchRejected)
 	})
+	t.Run("TrailingNewline", doTestBatchTrailingNewline)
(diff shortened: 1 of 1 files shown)


test/e2e/batches_test.go
```diff
@@ -303,6 +304,45 @@ func doTestBatchSharedInputFile(t *testing.T) {
 	}
 }
 
+// doTestBatchTrailingNewline verifies that trailing newlines in input files
```
j-mok-dev: this test is asserting behavior that the current implementation does not provide yet. `preProcessJob()` still passes blank lines to `extractAndValidateLine()`, so an input ending with `\n\n` becomes an empty JSON record and should fail ingestion rather than complete with `total=2`. Could you either include the parser-side fix in this PR or narrow the test to the currently supported case?

yizhaodev: good point, changed to test with only 1 trailing newline
let's handle more edge cases in the future

lioraron (approved):

PR #579, opened 2026-07-13
feat(processor): cancel pending async requests on batch abort (#543)

### LINKED ISSUE #543: [Feature]: Implement async dispatch cancellation in batch-processor
### Problem Statement

# Implement async dispatch cancellation in batch-processor

## Summary

When a batch job is cancelled while using async dispatch mode (llm-d-async), the batch-processor must correctly cancel in-flight inference requests. This requires handling three distinct cancellation cases depending on the job's lifecycle stage.

## Background

In sync dispatch mode, cancellation is straightforward: the processor aborts the HTTP request context, and the single in-flight request is immediately terminated. In async dispatch mode, requests are decoupled — they sit in a Redis queue and are processed independently by the dispatcher. The processor must coordinate cancellation across this boundary.

Depends on:
- llm-d-incubation/llm-d-async#299 (CancelRequests API)
- llm-d-incubation/llm-d-async#298 (Structured ResultMessage for cancel signaling)

### Case 1: Job still queued (not yet picked up by processor)

**Condition:** The batch job is in `validating` or `in_progress` status but the processor has not started its submit loop yet.

**Current behavior:** The apiserver sets `cancelling` status. When the processor picks it up, `requestAbortCtx` is already cancelled, so no requests are ever submitted.

**Action needed:** None. This already works correctly for both sync and async modes.

---

### Case 2: Job in-progress, submit loop is running

**Condition:** The processor is actively iterating over the input file and submitting requests to the dispatcher. A cancel signal arrives mid-loop.

**Current behavior:** `requestAbortCtx` is cancelled → the submit loop checks `ctx.Err()` on each iteration and breaks early. Remaining (unsent) requests are never submitted.

**Action needed:** None. The existing `requestAbortCtx` mechanism already stops the submit loop for both sync and async paths.

---

### Case 3: Requests already submitted to dispatcher ⚠️

**Condition:** Some or all requests have already been submitted to the dispatcher's Redis queue. The cancel signal arrives after submission.

**Current behavior (broken):** The processor stops the submit loop (Case 2) and then enters the drain/collect phase. It collects results for all pending requests and marks them as `batch_expired`. But:
- Requests sitting in the dispatcher queue **continue to be dispatched and consume inference compute**
- Results are marked `batch_expired` instead of `batch_cancelled`
- There is no signal to the dispatcher to stop processing these requests

**This is what we need to fix.**

### Step 1: Call CancelRequests on the async producer

After the submit loop is aborted, call the new `CancelRequests` API (llm-d-incubation/llm-d-async#299) with the list of request IDs that were already submitted:

```go
// After submit loop breaks due to cancel
if cancelRequested && p.asyncInference != nil {
    submittedIDs := collectSubmittedRequestIDs(pendingRequests)
    if err := p.asyncInference.CancelRequests(ctx, submittedIDs); err != nil {
        logger.Error(err, "failed to cancel requests in dispatcher")
    }
}
```

### Step 2: Collect results and mark as cancelled

Continue collecting results from the result queue (the dispatcher may have already processed some before receiving the cancel). For each result received:

- If the result indicates cancellation (via structured ResultMessage from #298: `StatusCode == 0, Error == "cancelled"`): write to error file as `batch_cancelled`
- If the result is a normal response that arrived before cancel took effect: **still override as `batch_cancelled`** — this aligns with sync path behavior where all responses are discarded on cancel

```go
// In the collect phase after cancel
for _, result := range collectedResults {
    writeResult(result, statusBatchCancelled) // override regardless of actual response
}
```

### Step 3: Fix status label — use `batch_cancelled` not `batch_expired`

The current drain logic uses `batch_expired` for all pending requests after abort. When the abort reason is cancellation (not expiration), the status should be `batch_cancelled` to match the OpenAI Batch API spec.

## Sync vs Async parity

| Behavior | Sync mode | Async mode (proposed) |
|----------|-----------|----------------------|
| Stop sending new requests | ✅ ctx cancel aborts HTTP | ✅ submit loop breaks on ctx.Err() |
| Cancel in-flight requests | ✅ HTTP context cancelled | ✅ CancelRequests API call |
| Result handling on cancel | Override with `batch_cancelled` | Override with `batch_cancelled` |
| Compute waste | Minimal (single request) | Reduced (cancel set prevents dispatch) |

## Implementation plan

1. **Wait for llm-d-async#299** — `CancelRequests` API availability
2. **Wait for llm-d-async#298** — Structured ResultMessage (to distinguish cancel from error)
3. **Add `CancelRequests` call** in `processModelAsync` after submit loop abort
4. **Fix drain status** from `batch_expired` → `batch_cancelled` when cancel is the abort reason
5. **Add integration test** with dispatcher that verifies cancelled requests don't produce inference results

## Related

- llm-d-incubation/llm-d-async#298
- llm-d-incubation/llm-d-async#299
- PR #458 (async dispatch integration — initial implementation)

CC: @lioraron

j-mok-dev:
## Why is this PR needed?

When a batch job is aborted (SLO expiry, user cancellation, or model failure) while using async dispatch, requests already submitted to the dispatcher sit in the queue until they timeout on their own. This wastes dispatcher and inference resources on work whose results will be discarded.

Additionally, uncollected pending requests were always classified as`batch_expired` regardless of the actual abort reason, producing misleading error codes in the output file.

## What does this PR do?

**Cancel dispatcher-side pending requests on abort:**

After the submit/collect loop breaks, the processor calls the dispatcher's `CancelRequests` API to mark still-pending request IDs as cancelled before they are dispatched. This is best-effort — uses a detached 5s timeout because `requestAbortCtx` is typically already cancelled at this point. Already-dispatched/in-flight requests are not 
affected; the existing `Close()` path handles waiter cleanup.

**Classify uncollected requests with the correct error code:**

Introduces `uncollectedPendingError`, which mirrors `drainAndFinalize`'s abort-reason priority to select the appropriate error code (`batch_expired`, `batch_cancelled`, `batch_failed`) for submitted-but-uncollected requests. Previously these were all hardcoded as `batch_expired`.

**Other changes:**
- Add `Cancel(ctx context.Context) error` to `AsyncInferenceClient` interface, implemented via `producer.CancelRequests`
- Bump `llm-d-async` dependencies from v0.7.2 to v0.7.3 (adds `CancelRequests` API)
- Fix `dev-deploy-dispatcher.sh`: fallback to `docker save | ctr import` when `kind load docker-image` fails on multi-platform OCI images

## How was this tested?

  - `SLOExpiryDuringCollect`: verifies `Cancel()` is called and pending
    requests get `batch_expired` error code
  - `Cancel` unit tests on `asyncProducerClient`: verifies Redis
    cancellation markers are set for pending IDs, and no-op for empty
    pending set
  - `BatchCancel` E2E test: submits a batch with mixed fast/slow
    requests, cancels mid-flight, verifies completed requests are
    recorded and remaining are marked `batch_cancelled`
  - Full `make dev-deploy` + `make test-e2e` pass

## Checklist

## Related Issues

Fixes #543


internal/processor/worker/executor.go
```diff
@@ -679,11 +679,23 @@ func (p *Processor) processModelAsync(
 		delete(pending, resp.RequestID)
 	}
 
+	// Best-effort: tell the dispatcher to drop still-pending requests before
+	// dispatch. Use a detached timeout — requestAbortCtx is often already
+	// cancelled when we reach this path.
+	if len(pending) > 0 {
+		cancelCtx, cancelFn := context.WithTimeout(context.Background(), 5*time.Second)
+		if err := asyncClient.Cancel(cancelCtx); err != nil {
```
lioraron, line 687: The Cancel error path is correctly handled here (log and continue — the drain loop maintains accounting regardless). However, none of the three new test cases (`CancelDuringCollect`, `UserCancelDuringCollect`, `SLOExpiryDuringCollect`) exercise this path — every `cancelFn` mock returns `nil`.

Consider adding a test variant where `cancelFn` returns an error, verifying that:
- The drain loop still writes the correct number of error lines
- `completed + failed == total` still holds
- The correct sentinel error is returned

This would guard against a regression if someone accidentally changes this to return early on Cancel failure.

internal/processor/worker/executor.go
```diff
@@ -699,6 +711,29 @@ func (p *Processor) processModelAsync(
 		inputFile, entries[submitCount:], writers, progress, modelErr, logger, len(entries), 0)
 }
 
+// uncollectedPendingError selects the error code/message for submitted-but-
+// uncollected async requests, matching drainAndFinalize's abort-reason order.
+func uncollectedPendingError(
+	sloCtx, userCancelCtx, mainCtx context.Context,
+	modelErr error,
```
lioraron, line 719: Nit: parameter order here is `(sloCtx, userCancelCtx, mainCtx, modelErr)` while `drainAndFinalize` (line 704) uses `(requestAbortCtx, mainCtx, sloCtx, userCancelCtx, ..., modelErr)`. Since both functions implement the same abort-reason priority switch, matching the context parameter order would reduce cognitive load when comparing them.

internal/processor/worker/executor.go
j-mok-dev: thanks. good catch! updated.

internal/processor/worker/executor.go
j-mok-dev: thanks! updated.

lioraron (approved):

Issue #580, opened 2026-07-14
enh: refactor request dispatch and collect with channel-based concurrency


evacchi:
### Problem Statement

Current design for the "sync" flow makes it harder to integrate a truly async processing flow. 

### Proposed Solution

We propose an alternative implementation for the async flow that can be also nicely retro-fit to work in the sync scenario.

- We propose to roll this out first to the new `async` flow
- Then we will extend this to the `sync` flow

Each job is sent to a `JobExecutor`. The JobExecutor assembles a pipeline of "actors" that exchange messages over channels.

- `ProgressTracker`: tracks request completion counts and pushes throttled updates to the status store
- `RequestSource`: produces `RequestItems` (inference requests)
  - Currently there is an implementation that follows the original impl, reading from a "Plan File" (`PlanFileSource`)
- A `PendingRequests` registry keeps track of outgoing, unfulfilled requests (optional, used for async processing -- see later)
- The `RequestDispatcher` reads `RequestItem` and may directly or indirectly produce a `ResultItem` (see below)
- The `ResultCollector` collects results per-job, reading from the job's "result" channel.

---

### Dispatchers

In the most basic case, a `DirectDispatcher`:
- receives `RequestItem` 
- relays them directly to an "InferenceClient",
- (optionally) sends the `ResultItem` to its result channel 

`RequestDispatcher`s can be chained:

```go
// RequestDispatcher handles a request — either by processing it directly
// (e.g. calling inference) or by delegating to another dispatcher
// (e.g. routing, microbatching). Composable as a chain.
type RequestDispatcher interface {
	// Run starts the message loop, reads from the request channel, may write directly or indirectly to the result channel.
	// A dispatcher does not write to the result channel if requests are handled asynchronously.
	// Returns when either the request channel is closed or the context is Done/Canceled etc.
	Run(ctx context.Context, requestCh <-chan RequestItem, resultCh chan<- ResultItem) error
}
```

---

### Async Dispatcher

```
Source ──[requestCh]──> AsyncDispatcher ──> queue (fire-and-forget)
                             │
                        pending.Store(req-id)

ResultBroadcaster(s) ──[resultCh]──> Collector
        │                                │
   SharedClient                     pending.Resolve(req-id)
   .GetResult()                     .Record()
                                         │
                                      Tracker
```

[...]

PR #581, opened 2026-07-14
enh: refactor async integration to channel-based pipeline

evacchi:
## Why is this PR needed?

We refactor `executeJob()` to accommodate an async, channel/gouroutine-based processing pipeline that will work better in the async dispatcher case, and can will fit nicely for sync processing as well.

## What does this PR do?

Replaces the `executeJob()` with `executeJobAsync()` for async processing, using a channel-based pipeline in a new `internal/processor/pipeline` package. The old execution path (`executeJob()` / `processModel()` ) is preserved; `processModelAsync()` is dropped, in favor of the new flow.

---

### Actors

| Actor | Responsibility |
|-------|---------------|
| **`RequestSource`** | Reads plan files, parses JSONL, assembles headers (SLO, fairness, inference objective), produces `RequestItem` values |
| **`RequestDispatcher`** | Routes requests and produces results. Composable as a delegate chain (useful in the sync case). Implementation is currently `AsyncDispatcher` (queue-backed) |
| **`ResultCollector`** | Reads results, writes JSONL to output/error files, records progress via tracker |
| **`ProgressTracker`** | Atomic counters + background ticker for throttled status-store pushes |
| **`JobExecutor`** | Orchestrator — creates channels, starts actors via errgroup, waits for cascade shutdown |

### Dispatcher chain (sync)

The `DirectDispatcher` sends requests to the HTTP clients as they come in without holdback.

It is designed to be the leaf in the dispatch pipeline: usually a gate of some sort will intercept requests and buffer them to rate limit. 

The DirectDispatcher sends requests, receives the responses and sends them out on the result channel for collection.

### Dispatcher (async)

```
AsyncDispatcher → queue (fire-and-forget)
                        │
ResultBroadcaster ← SharedClient.GetResult() → resultCh → Collector
```

The `AsyncDispatcher` 
1. submits to the queue and returns, and flags the ID in a per-job `PendingRequests` map
2. subscribes all messages for the corresponding result queue from a `ResultBroadcaster`

The `ResultBroadcaster` broadcasts responses from a given queue to all its subscribers (it DOES NOT filter per model ID). 

The collector reads from the result channel and matches against a  per-job `PendingRequests` map.

e.g., AsyncDispatcher sends out request ID X for Model/Pool A; then it subscribes for all results for model A; the collector filters out all results for model A that do not match request ID X.

### Pending requests

A `MutexMap`-based `PendingRequests` tracks in-flight requests. The dispatcher stores entries before dispatching; the collector resolves them when results arrive. Results with `CustomID` already set (cancels, inline errors) bypass the map. Results without `CustomID` (async inference) are enriched from the map. Unknown results (broadcast from another job) are silently dropped.

### Cascade shutdown

Each dispatcher closes `resultCh` when done → collector drains and returns → tracker is cancelled and pushes final counts. For cancellation, dispatchers drain remaining `requestCh` items as cancelled results before closing `resultCh`.

### Cancellation

The executor passes a single context (`requestAbortCtx`). The pipeline doesn't know *why* it was stopped. After `Execute` returns, the caller (`executeJobV2`) inspects `sloCtx`, `userCancelCtx`, and `mainCtx` to choose the terminal status (`errExpired`, `errCancelled`, `errShutdown`).

## How was this tested?

## Checklist

## Related Issues

First part of #580, supersedes #529, #528 and related.


internal/processor/worker/broadcasters.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (r *broadcasterRegistry) Run(ctx context.Context) {
+	for _, b := range r.broadcasters {
+		r.wg.Add(1)
+		go func() {
+			defer r.wg.Done()
+			b.Run(ctx)
+		}()
+	}
+}
```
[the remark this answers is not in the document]
evacchi: range variables have not been reused since go 1.22

internal/processor/pipeline/pending.go
```diff
@@ -0,0 +1,68 @@
...
+		if _, ok := p.m.LoadAndDelete(result.RequestID); ok {
+			if p.count.Add(-1) == 0 {
+				p.once.Do(func() { close(p.done) })
+			}
+		}
+		return true
+	}
+	if msg, ok := p.m.LoadAndDelete(result.RequestID); ok {
+		if p.count.Add(-1) == 0 {
+			p.once.Do(func() { close(p.done) })
```
j-mok-dev: I think this `done` channel is acting like a one-shot latch, but `PendingRequests` is not monotonic here: new requests can still be `Store()`d after the count temporarily reaches zero. In the async pipeline, an early result can resolve the first request and close `done`, then later submissions can increment the count again. After that, `Wait()` will return immediately even though there are still pending requests.

internal/processor/pipeline/pending.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// Wait blocks until all pending entries are resolved or ctx is cancelled.
+func (p *PendingRequests) Wait(ctx context.Context) {
+	if p.count.Load() == 0 {
+		return
+	}
+	select {
+	case <-ctx.Done():
+	case <-p.done:
```
j-mok-dev: I think `Wait()` can return too early here. `done` is closed permanently the first time the pending count reaches zero, but this pending set is not monotonic: in the async pipeline, earlier requests can resolve while later requests are still being submitted. In that interleaving, `Store()` can add new pending requests after `done` was already closed, and this branch will still return immediately even though unresolved requests remain.

internal/processor/pipeline/result_router.go
```diff
@@ -0,0 +1,150 @@
...
+		select {
+		case <-ctx.Done():
+			return
+
+		case msg, ok := <-incomingCh:
+			if !ok {
+				return
+			}
+			if msg.err != nil {
+				b.logger.Error(msg.err, "GetResult failed, stopping broadcaster")
```
j-mok-dev: Because this broadcaster is processor-wide shared state, returning here means one transient `GetResult()` error can kill result collection for all future jobs on this model until restart. I think this path should probably log and continue retrying instead of exiting the goroutine.

internal/processor/pipeline/collector.go
```diff
@@ -0,0 +1,116 @@
...
+// Both dispatchers close the channel when done, so this always terminates.
+// Returns ctx.Err() if the context was cancelled during draining.
+func (c *ResultCollector) Drain(ctx context.Context, resultCh <-chan ResultItem) error {
+	var firstErr error
+	for msg := range resultCh {
+		if !c.pending.Resolve(&msg) {
+			continue
+		}
+		if err := c.Receive(msg); err != nil && firstErr == nil {
+			firstErr = err
```
j-mok-dev, line 63: This remembers the first collector error but keeps draining while the rest of the pipeline can continue submitting requests. If local result persistence is already broken, should this abort the pipeline rather than allowing more async work whose results may never be recorded?

internal/processor/worker/source_planfile.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (s *PlanFileSource) readEntry(entry planEntry, modelID string) (*pipeline.RequestItem, error) {
+	buf := make([]byte, entry.Length)
+	if _, err := s.inputFile.ReadAt(buf, entry.Offset); err != nil {
+		return nil, fmt.Errorf("read input at offset %d: %w", entry.Offset, err)
+	}
+
+	trimmed := bytes.TrimSuffix(buf, []byte{'\n'})
+	var req batch_types.Request
+	if err := json.Unmarshal(trimmed, &req); err != nil {
```
j-mok-dev, line 97: In the current `readRequestLine()` / `executeOneRequest()` flow, a malformed input line becomes a request-level error entry. Here we just log and drop it, so async can silently lose a line instead of preserving it in `error.jsonl` and request counts.

internal/processor/pipeline/result_router.go
```diff
@@ -0,0 +1,150 @@
...
+	var body map[string]any
+	if err := json.Unmarshal(resp.Response, &body); err != nil {
+		logger.Error(err, "Failed to unmarshal async response", "requestID", resp.RequestID)
+		result.Error = &OutputError{
+			Code:    "parse_error",
+			Message: fmt.Sprintf("response body could not be parsed: %v", err),
+		}
+		return result
+	}
+	result.Response = &batch_types.ResponseData{
```
j-mok-dev, line 152: The new async path seems to drop some observability that exists in the sync execution flow: request error accounting, token usage extraction, and per-request inflight/execution timing. If that loss is intentional it probably needs to be called out, otherwise this looks like an async-only metrics regression.

internal/processor/worker/source_planfile.go
```diff
@@ -0,0 +1,144 @@
...
+		s.logger.Error(err, "Failed to parse request line, skipping")
+		return nil, nil
```
lioraron: **Regression: unparseable request lines silently dropped.**

When `json.Unmarshal` fails, returning `(nil, nil)` causes `Produce` to `continue` — the request vanishes from the pipeline with no error entry and no progress count. This means `completed + failed < total` for any batch with malformed lines.

The old `readRequestLine` (still in executor.go:889-905) returned a parse-error `*outputLine`, which was written to the error JSONL and counted as a failure. That invariant (`output_lines + error_lines == total_requests`) is relied on by the OpenAI batch API contract.

Consider returning a `RequestItem` with error metadata so it flows through the pipeline and the collector writes it to the error file:
```go
if err := json.Unmarshal(trimmed, &req); err != nil {
    s.logger.Error(err, "Failed to parse request line")
    errItem := &pipeline.RequestItem{
        RequestID: fmt.Sprintf("batch_req_%s", uuid.NewString()),
    }
    return errItem.Error("parse_error", fmt.Sprintf("failed to parse request line: %v", err)), nil
}
```
(Adjusting the return type of `readEntry` to return `*ResultItem` when appropriate, or emitting the error result directly.)

internal/processor/pipeline/dispatcher_async.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+	}
+
+	// Drain remaining requests as cancelled (if loop broke early).
+	for msg := range requestCh {
+		resultCh <- *msg.Canceled()
+	}
+
+	// Wait for all pending results to be resolved by the collector.
+	d.pending.Wait(ctx)
```
lioraron, line 88: **Regression: submitted-but-uncollected requests lost on cancellation.**

When `ctx` is cancelled, `Wait` returns immediately, leaving any requests that were submitted to the async queue but not yet collected with no corresponding output or error entry. The dispatcher then proceeds to `Unsubscribe` and `close(resultCh)` without draining them.

The old `processModelAsync` (removed in this PR) explicitly drained pending requests as errors:
```go
// Drain submitted-but-uncollected requests as errors so that
// output_lines + error_lines == total_requests.
for _, pr := range pending {
    out := newErrorOutputLine(pr.batchReqID, pr.customID,
        string(batch_types.ErrCodeBatchExpired), "result not collected before deadline")
    // ... write to error file, record failure
}
```

Without this, on SLO expiry or user cancel, `output_lines + error_lines < total_requests`, which violates the batch API contract.

Consider adding a drain step after `Wait` returns due to context cancellation — iterate unresolved entries in the pending map and send error results to `resultCh` before closing it.

internal/processor/pipeline/result_router.go
```diff
@@ -0,0 +1,150 @@
...
+		for {
+			resp, err := b.client.GetResult(ctx)
+			if err != nil {
+				if ctx.Err() != nil {
+					return
+				}
+				if errors.Is(err, context.DeadlineExceeded) {
+					continue
+				}
+				incomingCh <- resultMsg{err: err}
```
lioraron: **Regression: broadcaster permanently stops on transient errors.**

Any non-timeout `GetResult` error (e.g., transient Redis connection blip) causes the inner goroutine to return, which stops the broadcaster permanently. Since broadcasters are created once per model in `broadcasterRegistry` and shared across all jobs for the processor's lifetime, a single transient error kills async result collection for that model across all current and future jobs.

The old `resultDispatcher.run()` had a retry loop:
```go
func (d *resultDispatcher) run(ctx context.Context) {
    for ctx.Err() == nil {
        d.processNext(ctx)  // returns on error, but loop retries
    }
}
```

Consider adding retry-with-backoff logic here. Only stop the inner goroutine when `ctx.Err() != nil`.

internal/processor/pipeline/dispatcher_async.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Unsubscribe blocks until any in-progress broadcaster send completes
+	// (MutexMap.Delete waits for Range to release the lock), so closing
```
lioraron: This comment is factually incorrect. The broadcaster uses `subscribers.Keys()` (result_router.go:108), not `Range()`. `Keys()` acquires the lock, copies keys into a slice, and releases the lock *before* iteration. So `Delete()` (called by `Unsubscribe`) does NOT block waiting for sends to complete — there's a race window where the broadcaster can send to a channel after it's been unsubscribed and closed.

The actual safety is provided by `safeChannelSend`'s panic recovery, not by `MutexMap` locking semantics.

Either:
1. Correct the comment to describe the actual safety mechanism (panic recovery + the fact that `pending.Wait` ensures all expected results are already resolved before this runs), or
2. Switch the broadcast loop from `Keys()` to `Range()` so the guarantee described here actually holds (and `safeChannelSend` becomes unnecessary).

internal/processor/pipeline/executor.go
```diff
@@ -0,0 +1,70 @@
...
+	pending    *PendingRequests
+	collector  *ResultCollector
+	tracker    *ProgressTracker
+	logger     logr.Logger
+}
+
+type JobExecutorConfig struct {
+	Source     RequestSource
+	Dispatcher RequestDispatcher
+	Pending    *PendingRequests
```
lioraron: `Pending *PendingRequests` is stored in both the config and executor structs but never referenced in `Execute()`. No call site passes it either. Dead code — consider removing.

internal/processor/pipeline/progress_tracker.go
```diff
@@ -0,0 +1,115 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/go-logr/logr"
+
+	"github.com/llm-d/llm-d-batch-gateway/internal/shared/openai"
+)
+
+var progressUpdateInterval = time.Second
```
lioraron: Mutable package-level variable. CLAUDE.md bans mutable globals except for compiled regexps, SQL schemas, and error sentinels.

The `interval` field on `ProgressTracker` (line 30) already exists as the right injection point. Consider accepting it as a constructor parameter and removing this package-level var:
```go
func NewProgressTracker(total int64, updater ProgressUpdater, jobID string, interval time.Duration, logger logr.Logger) *ProgressTracker {
    return &ProgressTracker{
        total:    total,
        updater:  updater,
        jobID:    jobID,
        interval: interval,
        logger:   logger,
    }
}
```

internal/processor/pipeline/async_test.go
```diff
@@ -0,0 +1,296 @@
...
+		time.Sleep(50 * time.Millisecond)
+		for i := 0; i < 3; i++ {
+			client.deliver(items[i].RequestID, map[string]any{"ok": true})
+		}
+		time.Sleep(50 * time.Millisecond)
+		cancel()
+	}()
+
+	_, err := executor.Execute(ctx)
+	_ = err
```
lioraron: Discarding the error with `_ = err` means this test can't detect regressions like panics or data corruption. At minimum, assert that the error is `context.Canceled`.

Also, line 289 computes `total := outputLines + errorLines` but never asserts it equals `len(items)`. This is exactly the invariant that would catch the pending-drain regression (Finding 2 above). Consider:
```go
if total != len(items) {
    t.Errorf("total output+error lines = %d, want %d (all requests accounted for)", total, len(items))
}
```

internal/processor/pipeline/main_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	os.Exit(m.Run())
+}
+
+func makeItems(n int, model string) []RequestItem {
+	items := make([]RequestItem, n)
+	for i := range items {
+		items[i] = RequestItem{
+			RequestID: "req-" + model + "-" + string(rune('0'+i)),
```
lioraron: `string(rune('0'+i))` produces non-digit characters for i >= 10 (`:` for 10, `;` for 11, etc.). Currently safe since the max caller uses n=10 (indices 0-9), but this is a latent bug. Consider:
```go
RequestID: fmt.Sprintf("req-%s-%d", model, i),
CustomID:  fmt.Sprintf("c-%d", i),
```

internal/processor/worker/source_planfile.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/llm-d/llm-d-batch-gateway/internal/processor/config"
+	"github.com/llm-d/llm-d-batch-gateway/internal/processor/pipeline"
+	batch_types "github.com/llm-d/llm-d-batch-gateway/internal/shared/types"
+	"github.com/llm-d/llm-d-batch-gateway/pkg/clients/inference"
+)
+
+// PlanFileSource reads plan files and input JSONL to produce RequestItems.
+// After Produce returns, call Unproduced() to get the count of entries
+// that were not produced (e.g. due to context cancellation).
+type PlanFileSource struct {
```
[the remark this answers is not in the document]
evacchi: done

internal/processor/pipeline/dispatcher_async.go
evacchi: yes, outdated comment!

internal/processor/pipeline/pending.go
evacchi: should be addressed now. different approach: we know the job count from the start so we count down instead and we use a buffered channel instead (send `struct{}{}` on done with `select`) so we won't block

internal/processor/pipeline/pending.go
evacchi: see above

internal/processor/pipeline/result_router.go
evacchi: should be addressed

internal/processor/pipeline/collector.go
evacchi: should be addressed now

internal/processor/worker/source_planfile.go
evacchi: should be addressed nowx

internal/processor/pipeline/result_router.go
evacchi: observability metrics reintroduced, thanks for noticing!

internal/processor/worker/source_planfile.go
evacchi: Thanks for noticing, should be addressed now. We add in fact an `Error` field in `RequestItem` to track parse errors and propagate them in the pipeline

internal/processor/pipeline/dispatcher_async.go
evacchi: should be addressed now!

internal/processor/pipeline/result_router.go
evacchi: added backoff!

internal/processor/pipeline/executor.go
evacchi: good eye!

internal/processor/pipeline/progress_tracker.go
evacchi: good point!

internal/processor/pipeline/async_test.go
evacchi: fixed

internal/processor/pipeline/main_test.go
evacchi: fair, fixed

evacchi:
rebased + added support to cancellation, verified e2e tests still pass
following up with PR for "direct" dispatching (i.e. without llm-d-async)

internal/processor/pipeline/collector.go
j-mok-dev: The collector now skips further writes after the first persistence failure, but the rest of the pipeline can still keep submitting async work until the source/dispatcher finish. That still means we may enqueue remote work even after local result persistence is already broken.

If we want persistence failure to be terminal for the job, I think we still need to abort the pipeline promptly here rather than only returning the error after drain completes.

internal/processor/worker/source_planfile.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			return fmt.Errorf("read plan for model %s: %w", modelID, err)
+		}
+
+		for _, entry := range entries {
+			item, err := s.readEntry(entry, modelID)
+			if err != nil {
+				return err
+			}
+			outgoingRequestCh <- *item
```
j-mok-dev, line 82: `Produce()` does a plain blocking send here and does not observe `ctx.Done()` while handing items to the dispatcher. So if the downstream side stops reading or is cancelled, this source can still block indefinitely on the channel send instead of exiting promptly with the request context.
I think this send should be wrapped in a `select` on `outgoingRequestCh <- *item` and `ctx.Done()` so the source respects cancellation during shutdown.

internal/processor/pipeline/collector.go
evacchi: true, but the input file has a limited size and eventually we will reach the end, we can bubble up the signal to the top, though; it depends on how critical it is to short-circuit -- I'd argue, not that critical, we shouldn't actually process when we reach the error case, we just drain+skip so it's relatively quick.

internal/processor/pipeline/request_item.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (r *RequestItem) Error(code, message string) *ResultItem {
+	return &ResultItem{
+		RequestID: r.RequestID,
+		CustomID:  r.CustomID,
+		ModelID:   r.ModelID,
+		Error:     &OutputError{Code: code, Message: message},
+	}
+}
```
lioraron, line 54: **Bug: inflight metrics leak on all error/cancel paths**

`Error()` does not propagate `SubmittedAt` to the `ResultItem`. In the dispatcher, `IncProcessorInflightRequests` / `IncModelInflightRequests` are called *before* `Submit()`, but the collector only decrements when `msg.SubmittedAt.IsZero()` is false.

This means metrics leak on: submit failures, cancellations, and `DrainUnresolved` — basically every error path. Only successful broadcaster results (where `Resolve` copies `SubmittedAt` in the third branch) correctly decrement. Over time, this permanently inflates inflight counters.

Fix:
```go
func (r *RequestItem) Error(code, message string) *ResultItem {
	return &ResultItem{
		RequestID:   r.RequestID,
		CustomID:    r.CustomID,
		ModelID:     r.ModelID,
		SubmittedAt: r.SubmittedAt,
		Error:       &OutputError{Code: code, Message: message},
	}
}
```

internal/processor/pipeline/pending.go
```diff
@@ -0,0 +1,104 @@
...
+// Resolve enriches a result with request metadata. Returns true if the
+// result is accepted: either it already has metadata (cancels, inline errors)
+// or it was found in the pending map (async inference results).
+// Returns false only for broadcast results that belong to another job.
+func (p *PendingRequests) Resolve(result *ResultItem) bool {
+	if result.Error != nil {
+		if _, ok := p.m.LoadAndDelete(result.RequestID); ok {
+			p.decrement()
+		}
+		return true
```
lioraron: **Bug: cross-job error contamination from broadcaster**

When `result.Error != nil`, this returns `true` unconditionally — even if the `RequestID` was NOT in the pending map. Broadcaster error results (nil body or bad JSON from `asyncResult`) have `Error` set but empty `CustomID`. If two jobs share a broadcaster, Job B accepts and writes Job A's error results to its own error file.

Inline errors from the dispatcher always have `CustomID` set (via `msg.Error()`), so we can distinguish:

```go
if result.Error != nil {
	if _, ok := p.m.LoadAndDelete(result.RequestID); ok {
		p.decrement()
		return true
	}
	// Accept inline errors (have CustomID), reject broadcast errors for other jobs.
	return result.CustomID != ""
}
```

internal/processor/worker/execute_pipeline.go
```diff
@@ -0,0 +1,191 @@
...
+		Logger:             logger,
+	})
+
+	// The dispatcher forwards requests for processing.
+	pending := pipeline.NewPendingRequests(modelMap.LineCount)
+
+	// Broadcasters propagate asynchronous responses to subscribers.
+	broadcasters := p.broadcasters.forModels(modelMap)
+
+	// The async dispatcher forwards requests to the async dispatcher
```
lioraron: Copy-paste error: "The async dispatcher forwards requests to the async dispatcher" references itself.

Should be something like: "The async dispatcher submits requests to the async queue and subscribes the result channel to the per-model broadcasters."

internal/processor/pipeline/dispatcher.go
```diff
@@ -0,0 +1,13 @@
+package pipeline
+
+import "context"
+
+// RequestDispatcher handles a request — either by processing it directly
+// (e.g. calling inference) or by delegating to another dispatcher
+// (e.g. routing, microbatching). Composable as a chain.
+type RequestDispatcher interface {
+	// Run starts the message loop, reads from the request channel, may write directly or indirectly to the result channel.
+	// A dispatcher does not write to the result channel if requests are handled asynchronously.
```
lioraron: This is inaccurate — `AsyncDispatcher.Run()` writes directly to `resultCh` for cancel errors, parse errors, model-not-found errors, submit errors, and `DrainUnresolved` entries. Only successful inference results arrive indirectly via broadcaster subscriptions.

Suggestion:
```go
// Run starts the message loop, reads from the request channel, and writes results to the result channel.
// An async dispatcher writes error results directly; successful inference results arrive via broadcaster subscriptions.
// Returns when either the request channel is closed or the context is Done/Canceled etc.
```

internal/processor/pipeline/result_router.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+		}
+	}
+}
+
+// safeChannelSend sends to `ch` and handles the panic if the channel was closed concurrently
+// (we received an unsubscription request). It is safe to just ignore the panic.
+func (b *ResultBroadcaster) safeChannelSend(result ResultItem, ch chan<- ResultItem) {
+	defer func() {
+		if r := recover(); r != nil {
```
lioraron, line 123: `recover()` catches *any* panic, not just "send on closed channel". An unrelated nil-pointer dereference or index-out-of-range in the broadcast path would be silently swallowed with Info-level logging, making future debugging very difficult.

Consider checking the panic type:
```go
defer func() {
	if r := recover(); r != nil {
		if msg, ok := r.(string); ok && msg == "send on closed channel" {
			b.logger.Info("Broadcast send recovered (subscriber unsubscribed)",
				"requestID", result.RequestID)
		} else {
			panic(r) // re-panic for unexpected failures
		}
	}
}()
```

.gitignore
```diff
@@ -12,3 +12,5 @@ benchmarks/results/*
 .dispatcher-port-forward.pid
 .dispatcher-sim-port-forward.pid
 *.DS_Store
+.idea/
+
```
lioraron: Pre-commit CI failure: there's a trailing blank line after this. The `end-of-file-fixer` hook wants the file to end with exactly one newline — remove the extra empty line.

internal/processor/pipeline/result_router.go
evacchi: I'll do it for extra safety (`recover()` is on the slow path anyway), but the only line in this function is:

```
	ch <- result
```

so that's all we're protecting here

internal/processor/worker/source_planfile.go
evacchi: right, but as commented above https://github.com/llm-d/llm-d-batch-gateway/pull/581#discussion_r3581967080, the downstream side (the collector) always drains the channel, so this should be fine.

lioraron (approved):
Thanks!

Issue #572, opened 2026-07-10
Atomic job pickup and crash recovery via Postgres-native queue


acardace:
### Problem

In `runPollingLoop` ([worker.go:254-268](https://github.com/llm-d/llm-d-batch-gateway/blob/cc34fa9/internal/processor/worker/worker.go#L254)), there is a window between `PQDequeue` (removing the job from the Redis queue) and `InFlightSet` (recording it in the Redis in-flight hash) where a crash leaves the job invisible — it is neither in the queue nor in the in-flight table. Only the GC reconciler (default 60-minute cycle) can detect it by cross-referencing non-terminal DB records against the queue and in-flight hash.

More broadly, job ownership is spread across three data stores (Postgres/Redis DB for status, Redis sorted set for the queue, Redis hash for in-flight tracking) with no transactional guarantee tying them together. Every crash boundary between these stores is a potential inconsistency that the reconciler must paper over after the fact.

Recovery latency is also a concern: a crashed job waits up to one full reconciler interval before being detected and re-enqueued or failed. For SLO-sensitive batches this delay can exhaust the SLO, turning a recoverable job into an expired one.

### Proposed solution

Use the Postgres DB as the sole source of truth for job ownership and queuing. Replace the Redis priority queue with a Postgres-native queue using `SELECT ... FOR UPDATE SKIP LOCKED`, and replace the Redis in-flight hash with a `processor_id` column on the batch item. Job dequeue and claim become a single atomic transaction — no gap, no distributed bookkeeping.

Switch the processor from a Deployment to a StatefulSet so each replica has a stable identity (`processor-0`, `processor-1`, ...). This enables immediate crash recovery: a restarted processor queries the DB for jobs it previously owned and resumes them without waiting for the GC reconciler.

### Schema changes

Add three columns to `batch_items`:

```sql
ALTER TABLE batch_items ADD COLUMN processor_id TEXT;
ALTER TABLE batch_items ADD COLUMN priority BIGINT;
ALTER TABLE batch_items ADD COLUMN epoch BIGINT NOT NULL DEFAULT 0;

CREATE INDEX idx_batch_items_queue
  ON batch_items (priority ASC)
  WHERE status = 'validating' AND processor_id IS NULL;

CREATE INDEX idx_batch_items_processor
  ON batch_items (processor_id)
  WHERE processor_id IS NOT NULL;
```

- `processor_id` records which processor pod owns an in-progress job.
[...]

PR #582, opened 2026-07-14
feat: replace Redis queue with Postgres-native atomic job ownership

acardace:
## Summary

Replaces the Redis-based priority queue and in-flight tracking with a Postgres-native queue using `SELECT ... FOR UPDATE SKIP LOCKED`. Job dequeue and ownership become a single atomic transaction — no gap, no distributed bookkeeping.

**Design**: #572

### Postgres-native queue
- Add `processor_id`, `priority`, and `epoch` columns to `batch_items`
- `PQDequeue`: atomic CTE that claims jobs with `FOR UPDATE SKIP LOCKED` and sets `processor_id`
- `PQEnqueue`: re-enqueue with guards against resurrecting terminal or already-queued jobs
- `PQDelete`: atomic cancel-from-queue with `cancelled_at` timestamp
- Partial indexes for queue (`priority ASC WHERE processor_id IS NULL`) and recovery (`processor_id WHERE NOT NULL`)

### Processor: Deployment → StatefulSet
- Stable pod identity (`processor-0`, `processor-1`, ...) via `os.Hostname()`
- Headless service for StatefulSet DNS

### Job recovery
- `recoverOwnedJobs`: on startup, processor queries DB for non-terminal jobs by `processor_id` and recovers them — no GC dependency for same-pod restarts
- Removed `recoverStaleJobs` (filesystem scan) — superseded by DB-based recovery
- Recovery is phase-aware: finalizing → complete uploads, cancelling → cancelled, in_progress → re-enqueue or fail based on SLO

### GC: event-driven orphan recovery
- K8s pod watcher (informer) detects processor pod changes
- Reconciler re-enqueues orphans with valid SLO, fails expired ones, transitions `cancelling` orphans to `cancelled`
- `HasProcessorID` filter — reconciler only fetches owned jobs, not the entire queue

### Epoch-based fencing
- Fencing token incremented on every ownership change (dequeue, re-enqueue, GC reclaim)
- Processor writes include `WHERE epoch = N` — a zombie whose job was reclaimed cannot overwrite the new owner's state
- GC terminal transitions bump epoch via `BumpEpoch` flag

### Cancel handler
- `PQDelete` sets `cancelled` + `cancelled_at` atomically; handler re-reads and returns current state
- In-progress cancel uses CAS (`expectedStatus`) to prevent overwriting terminal status

### Complexity reduction
- **Removed**: Redis `BatchDBClient`, `FileDBClient`, `InFlightClient`, heartbeat goroutine
- **Kept**: Redis for real-time progress counts, cancel events, exchange
- Net change: -567 lines

## Testing
- Unit tests for queue, pod watcher, reconciler, recovery, epoch fencing, CAS cancel
- E2e tests updated for StatefulSet self-recovery semantics
- All unit and e2e tests passing


acardace:
@hexfusion @evacchi @wseaton

BUGS
```diff
@@ -0,0 +1,3 @@
+TODO:
```
wseaton, line 1: Is this to be committed, or just a planning doc?

j-mok-dev:
I agree that this PR is trying to solve a real correctness problem, but I have a concern about the architectural direction it takes.

This design collapses queue ownership directly into the Postgres batch store (`processor_id`, `priority`, `epoch` on `batch_items`). Once we do that, the queue is no longer an independently pluggable component. Queue semantics become part of the SQL row model.

I do not think this is the direction we want. One of the goals of this project is to keep storage roles independently replaceable. We have already been moving in the direction of making Redis optional by providing alternative backend implementations, not by collapsing multiple roles into a single storage model. If we merge this design, we effectively give up queue-level pluggability and make SQL-backed ownership semantics the only realistic model.

The better direction, in my view, is to separate delivery from ownership:

- keep the queue as a pluggable delivery mechanism
- introduce a separate claim / lease / fencing abstraction for ownership correctness
- implement that coordination layer in Postgres first if needed
- require processors to claim work after dequeue and only execute on successful claim
- carry the fencing token through later writes, recovery, and finalization

That would preserve the correctness properties this PR is aiming for without destroying the queue abstraction.

wseaton:
> I do not think this is the direction we want. One of the goals of this project is to keep storage roles independently replaceable. We have already been moving in the direction of making Redis optional by providing alternative backend implementations, not by collapsing multiple roles into a single storage model. If we merge this design, we effectively give up queue-level pluggability and make SQL-backed ownership semantics the only realistic model.

As a follow-up, maybe it's better to focus instead on decoupling job level ownership (low TPS) and transport (higher TPS) into two distinct interfaces, if we break up `BatchPriorityQueueClient` we can keep two seperate queue implementations but fix the original correctness issues that motivated this PR. 

Ownership can stay in an RDBMS (we already require one) and then transport can stay pluggable. I think I agree with all the other points made

j-mok-dev:
> > I do not think this is the direction we want. One of the goals of this project is to keep storage roles independently replaceable. We have already been moving in the direction of making Redis optional by providing alternative backend implementations, not by collapsing multiple roles into a single storage model. If we merge this design, we effectively give up queue-level pluggability and make SQL-backed ownership semantics the only realistic model.
> 
> As a follow-up, maybe it's better to focus instead on decoupling job level ownership (low TPS) and transport (higher TPS) into two distinct interfaces, if we break up `BatchPriorityQueueClient` we can keep two seperate queue implementations but fix the original correctness issues that motivated this PR.
> 
> Ownership can stay in an RDBMS (we already require one) and then transport can stay pluggable. I think I agree with all the other points made

Yes, I agree. This makes the direction much clearer.
My concern was not “ownership in Postgres,” but “ownership and transport collapsing into the same abstraction.” If we separate those concerns, we can preserve the correctness guarantees without giving up transport-level pluggability.
I think this is the right direction to move in. 👍

lioraron:
I agree with @j-mok-dev and @wseaton - added a comment in the design issue https://github.com/llm-d/llm-d-batch-gateway/issues/572#issuecomment-4973194353

acardace:
Comment to remove stale label.

internal/database/postgresql/batch_schema.sql
```diff
@@ -12,14 +12,29 @@
...
+    id            TEXT PRIMARY KEY,
+    tenant_id     TEXT NOT NULL,
+    expiry        BIGINT,
+    tags          JSONB,
+    spec          JSONB,
+    status        JSONB,
+    processor_id  TEXT,
+    priority      BIGINT,
+    epoch         BIGINT NOT NULL DEFAULT 0
 );
```
evacchi, line 25: possibly not relevant, but this won't update the table if it exist.
If we do care, then we might want to change this to something like:

```
CREATE TABLE IF NOT EXISTS batch_items (
... same as earlier ...
)
....
ALTER TABLE ... ADD COLUMN IF NOT EXISTS --> add this
```

internal/processor/worker/recovery.go
```diff
@@ -60,55 +59,47 @@ type recoveryResult struct {
...
-func (p *Processor) recoverStaleJobs(ctx context.Context) {
+// Runs once at startup before the polling loop.
+func (p *Processor) recoverOwnedJobs(ctx context.Context) {
 	logger := logr.FromContextOrDiscard(ctx)
 
-	dirs, err := p.discoverStaleJobDirs()
+	items, _, _, err := p.batchDB.DBGet(ctx, &db.BatchQuery{
+		ProcessorID: p.processorID,
+		NonTerminal: true,
+	}, true, 0, 1000)
```
evacchi: this is fetching the first 1000 jobs. do we need to paginate?

internal/database/postgresql/batch_queue.go
```diff
@@ -0,0 +1,219 @@
...
+			    epoch = epoch + 1
+			WHERE id = $1
+			  AND processor_id IS NOT NULL
+			  AND `+nonTerminalCondition+`
+		)
+		-- TODO: the processor polling loop (worker.go) could LISTEN on this channel
+		-- to wake up immediately instead of waiting for the next poll interval.
+		SELECT pg_notify('batch_jobs_available', '')`,
+		jobPriority.ID,
+	)
```
evacchi: ok take this with some grain of salt because it's a bot's claim.

It looks like during CreateBatch, DBStore creates new rows with:

```
status = validating
processor_id = NULL
```

this is eligible for PQDequeue to claim. A worker can claim it in the interval before PQEnqueue runs.
PQEnqueue then sees the non-null processor_id and treats that row as a candidate for re-enqueue, clearing its ownership and returning it to the queue.

double-check this for safety

evacchi (commented):
some nits and maybe a potential race condition that you may want to verify by adding an ad-hoc test.

there is also a BUGS file that we probably want to delete or convert into issues

internal/database/postgresql/batch_queue.go
acardace: Yes that checks out and it is now fixed, I've just removed the second enqueue which was useless. Thank you.

internal/database/postgresql/batch_schema.sql
acardace: Fixed, thank you.

internal/processor/worker/recovery.go
acardace: Yes, fixed.

BUGS
evacchi: @acardace don't forget to rm this / add issues instead

evacchi (approved):
🫡

internal/apiserver/batch/batch_handler.go
```diff
@@ -580,21 +540,19 @@ func (c *BatchAPIHandler) CancelBatch(w http.ResponseWriter, r *http.Request) {
 		return
```
wseaton, line 547:
```diff
-	if err := c.clients.BatchDB.DBUpdate(ctx, dbItem, nil); err != nil {
+	dbItem.Epoch = item.Epoch
+	if err := c.clients.BatchDB.DBUpdate(ctx, dbItem, item.Status); err != nil {
+		if errors.Is(err, api.ErrConflict) {
+			apiErr := openai.NewAPIError(http.StatusConflict, "", "batch changed state during cancel, retry", nil)
+			common.WriteAPIError(w, r, apiErr)
+			return
+		}
 		logger.Error(err, "failed to update batch in database")
 		common.WriteInternalServerError(w, r)
 		return
 	}
```

We need to catch `api.ErrConflict` here and use the item's own epoch for a proper CAS

internal/gc/reconciler/reconciler.go
```diff
@@ -32,41 +32,44 @@ import (
...
 type Reconciler struct {
 	batchDB         db.BatchDBClient
 	queue           db.BatchPriorityQueueClient
-	inflight        db.InFlightClient
 	interval        time.Duration
 	dryRun          bool
 	onCycleComplete func(*Result)
+
+	mu             sync.RWMutex
+	liveProcessors map[string]bool
```
wseaton, line 64: We should track whether a snapshot has arrived and treat unknown differently from empty

`liveProcessors` starts empty would mean every owned job is an orphan, so a GC start while the StatefulSet is unstable, or a bad STS name, selector, or RBAC, re-enqueues everything and fences out every live processor on each tick

```suggestion
	liveProcessors map[string]bool
	snapshotReady  bool
```

internal/gc/reconciler/reconciler.go
```diff
@@ -77,30 +80,48 @@ func NewReconciler(
...
 }
 
-// RunLoop runs the reconciler in a continuous loop at the configured interval.
+// SetLiveProcessors updates the set of currently alive processor pod names.
+// Called by the pod watcher on add/delete events.
+func (r *Reconciler) SetLiveProcessors(processors map[string]bool) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.liveProcessors = processors
+}
```
wseaton, line 111:
```suggestion
	r.liveProcessors = processors
	r.snapshotReady = true
}

func (r *Reconciler) hasSnapshot() bool {
	r.mu.RLock()
	defer r.mu.RUnlock()
	return r.snapshotReady
}
```

internal/gc/reconciler/reconciler.go
```diff
@@ -143,45 +163,11 @@ func (r *Reconciler) run(ctx context.Context) {
...
-			}
-
-			if entry, ok := inflightEntries[job.ID]; ok {
-				lastSeen := time.Unix(entry.LastSeen, 0)
-				if lastSeen.After(stalenessThreshold) {
-					continue
-				}
-			}
-
+	for _, job := range jobs {
```
wseaton, line 179: We should skip the cycle until the watcher has reported once. Ideally this sits before the fetch

```suggestion
	if !r.hasSnapshot() {
		logger.Info("Reconciler: no live processor snapshot yet, skipping cycle")
		return
	}

	for _, job := range jobs {
```

internal/database/postgresql/batch_schema.sql
```diff
@@ -12,14 +12,34 @@
...
 
+-- Schema migration for existing tables from previous versions.
+ALTER TABLE batch_items ADD COLUMN IF NOT EXISTS processor_id TEXT;
+ALTER TABLE batch_items ADD COLUMN IF NOT EXISTS priority BIGINT;
+ALTER TABLE batch_items ADD COLUMN IF NOT EXISTS epoch BIGINT NOT NULL DEFAULT 0;
```
wseaton, line 30:
```suggestion
ALTER TABLE batch_items ADD COLUMN IF NOT EXISTS epoch BIGINT NOT NULL DEFAULT 0;

-- Rows written before the queue columns existed carry the SLO in a tag and
-- have no owner. Restore the queue order from the tag and hand in-flight rows
-- to a sentinel owner so the reconciler reclaims them on its first cycle.
UPDATE batch_items
   SET priority = (tags->>'slo_unix_micro')::bigint
 WHERE priority IS NULL AND tags ? 'slo_unix_micro';
UPDATE batch_items
   SET processor_id = 'pre-migration'
 WHERE processor_id IS NULL
   AND status::jsonb->>'status' IN ('in_progress', 'finalizing', 'cancelling');
```

Fable 5.1 found this one, related to migration/upgrade safety: 

Existing rows come out of the ALTERs with `priority` and `processor_id` NULL. A pre-upgrade `validating` row gets claimed by `PQDequeue`'s UPDATE and then the scan of NULL `priority` into `int64` fails, so the row is claimed and lost until the pod restarts. A pre-upgrade `in_progress`/`finalizing` row has no owner, is not `validating`, and so is invisible to both the queue and the reconciler forever.

internal/database/postgresql/batch_queue.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return fmt.Errorf("PQEnqueue: %w", err)
+	}
+	return nil
```
wseaton:
```suggestion
	result, err := c.pool.Exec(ctx,
		`WITH re_enqueued AS (
			UPDATE batch_items
			SET processor_id = NULL,
			    status = jsonb_set(status, '{status}', '"validating"'),
			    epoch = epoch + 1
			WHERE id = $1
			  AND processor_id IS NOT NULL
			  AND `+nonTerminalCondition+`
			RETURNING id
		)
		-- TODO: the processor polling loop (worker.go) could LISTEN on this channel
		-- to wake up immediately instead of waiting for the next poll interval.
		SELECT pg_notify('batch_jobs_available', '') FROM re_enqueued`,
		jobPriority.ID,
	)
	if err != nil {
		return fmt.Errorf("PQEnqueue: %w", err)
	}
	if result.RowsAffected() == 0 {
		return fmt.Errorf("PQEnqueue: %s is not owned or already terminal: %w", jobPriority.ID, api.ErrConflict)
	}
	return nil
```

When the guard matches nothing (job already terminal, or already back in the queue) this returns nil, so the reconciler counts a re-enqueue and the worker logs one for a no-op. Return the rows affected as a conflict so callers can tell. Moving the notify into the CTE result also stops us waking processors when nothing changed.

internal/database/postgresql/file_db.go
```diff
@@ -106,7 +106,7 @@ func (c *PostgresFileDBClient) DBGet(
 		items[i] = &api.FileItem{
 			BaseIndexes:  *indexes[i],
 			BaseContents: *contents[i],
-			Purpose:      extras[i][colPurpose],
+			Purpose:      extras[i][colPurpose].(string),
 		}
```
wseaton, line 111:
```suggestion
		purpose, _ := extras[i][colPurpose].(string)
		items[i] = &api.FileItem{
			BaseIndexes:  *indexes[i],
			BaseContents: *contents[i],
			Purpose:      purpose,
		}
```

nit: `purpose` is nullable in `file_schema.sql`, so a bare assertion panics the apiserver on list. Comma-ok form per the repo rules.

wseaton:
Also running the full fault tolerance simulation from #653 on this

wseaton (changes requested):

internal/processor/worker/recovery.go
```diff
@@ -60,55 +59,57 @@ type recoveryResult struct {
...
-	dirs, err := p.discoverStaleJobDirs()
-	if err != nil {
-		logger.Error(err, "Startup recovery: failed to scan workdir")
-		return
+	const pageSize = 1000
+	var items []*db.BatchItem
+	cursor := 0
+	for {
+		batch, nextCursor, more, err := p.batchDB.DBGet(ctx, &db.BatchQuery{
+			ProcessorID: p.processorID,
```
wseaton: Claim owned jobs in one UPDATE that bumps `epoch` and a new `recovery_attempts` counter, recover at the returned epoch, and past a small budget go straight to `recoverWithFailed`. Reset the counter in `PQDequeue`.

```sql
UPDATE batch_items
   SET epoch = epoch + 1, recovery_attempts = recovery_attempts + 1
 WHERE processor_id = $1 AND <non-terminal>
RETURNING id, epoch, recovery_attempts
```

Two bugs here:

- Recovery writes at the old epoch, so a zombie and its same-named replacement both pass the fence (`TestRecoverOwnedJobsFencesPreviousEpoch`).
- A job that crashes its processor during recovery loops forever: same-named pod, same job, same crash. Sim scenarios F5/F6 went from `failed` on main to pinned in `finalizing` with the pod in CrashLoopBackOff, and the reconciler never reclaims because the StatefulSet is never stable.

Evidence: https://gist.github.com/wseaton/f4011ab07deedfc49f1f3e8ba9c593bc

wseaton: Claim owned jobs in one UPDATE that bumps `epoch` and a new `recovery_attempts` counter, recover at the returned epoch, and past a small budget go straight to `recoverWithFailed`. Reset the counter in `PQDequeue`.

```sql
UPDATE batch_items
   SET epoch = epoch + 1, recovery_attempts = recovery_attempts + 1
 WHERE processor_id = $1 AND <non-terminal>
RETURNING id, epoch, recovery_attempts
```

Two bugs:
- Recovery writes at the old epoch, so a zombie and its same-named replacement both pass the fence (`TestRecoverOwnedJobsFencesPreviousEpoch`).
- A job that crashes its processor during recovery loops forever. Sim scenarios F5/F6 went from `failed` on main to pinned in `finalizing` with the pod in CrashLoopBackOff, the reconciler never reclaims because the StatefulSet is never stable.

Tested patch (`gist-fix-recovery-budget.patch`) and evidence: https://gist.github.com/wseaton/f4011ab07deedfc49f1f3e8ba9c593bc

BUGS
acardace: that's just a mistake, will fix it.

acardace: They were personal notes.

acardace:
Thanks @wseaton, I think I have addressed everything you pointed out in the review. PTAL

wseaton (approved):
LGTM, thank you for the hard work on this change!

The simulation harness from #653 shows several correctness bugs fixed from this PR, namely:

- Terminal overwrite. A processor's completed write could clobber a terminal status the reconciler had already written. Ownership is now fenced, so a stale owner's write is rejected.
- Worker crash stranding the job. A job whose processor died mid-run was failed with all completed work discarded. It is now re-enqueued and finishes with its output.
- Cancel revert. Cancelling a queued batch had a window where the batch was out of the queue but still validating, and the reconciler re-enqueued it behind the user's back. Cancel-from-queue is now one atomic row transition.

PR #584, opened 2026-07-15
enh: refactor sync flow to channel-based pipeline

### LINKED ISSUE #580: enh: refactor request dispatch and collect with channel-based concurrency

evacchi:
Note: Optionally, the last commit also deletes the old `executeJob()` (for now deprecate and call executeJobAsync()` to keep the diff smaller) and all related dead code. Test cases should be already migrated to equivalent cases in the new flow. We can also do the deletion in a separate PR.

---

Current design for the "sync" flow makes it harder to integrate a truly async processing flow. 

In PR #581 we introducef a fully-async pipeline for llm-d-async integration, while we mention that the same design can be adapted for the sync case as well (where HTTP requests are handled directly).

The unified design makes it easier to handle both cases and should be overall more composable, this PR migrates to the same design the "sync" flow as well.

```
Source ──[requestCh]──> Optional Gate ──> DirectDispatcher ──[resultCh]──> Collector
                            │                   │                              │
                       e.g AIMD         spawn goroutine                .Record()
                                        call Generate()                     │
                                        send result                     Tracker

```

## What does this PR do?

This PR introduces `AIMDDispatcher` and `DirectDispatcher` and it refactors some cross-cutting concerns into `PreDispatcher` (publishing metrics, filtering out parse errors)

```

...──[requestCh]──> PreProcessor ──> Optional Gate ──> DirectDispatcher ──[resultCh]──>...    
                                          │                   │                     
                                     e.g AIMD         spawn goroutine                
                                                         call Generate()             
                                                         send result                 
```

## How was this tested?

- added test coverage, verified existing e2e show the same behavior, 
- being verified on a real workload on a real cluster.

## Checklist

## Related Issues

Related to #581.
Closes #580 once both are merged.


internal/processor/worker/execute_pipeline.go
```diff
@@ -0,0 +1,226 @@
...
+			pipeline.AIMDDispatcherStage(models, p.cfg.Concurrency.Global, logger),
+			pipeline.DirectDispatcherStage(p.inference, logger),
+		)
+	default:
+		models := buildAIMDModels(modelMap, p.inference, p.endpointLimits)
+		return pipeline.Pipe(
+			pipeline.PreDispatcherStage(),
+			pipeline.AIMDDispatcherStage(models, p.cfg.Concurrency.Global, logger),
+			pipeline.DirectDispatcherStage(p.inference, logger),
+		)
```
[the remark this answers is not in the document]
evacchi: intentional, the AIMD component also manages general dispatch limits -- we can reconsider how to handle this in the future.

evacchi:
rebased

evacchi:
the original version of this PR also included a DSL for piping `Dispatchers`:


```go
	switch {
	case p.asyncInference != nil:
		broadcasters := p.broadcasters.forModels(modelMap)
		return pipeline.Pipe(
			pipeline.PreDispatcherStage(),
			pipeline.AsyncDispatcherStage(p.asyncInference, broadcasters, pending, logger),
		)
	case p.cfg.Concurrency.AIMD.Enabled:
		models := buildAIMDModels(modelMap, p.inference, p.endpointLimits)
		return pipeline.Pipe(
			pipeline.PreDispatcherStage(),
			pipeline.AIMDDispatcherStage(models, p.cfg.Concurrency.Global, logger),
			pipeline.DirectDispatcherStage(p.inference, logger),
		)
	default:
		return pipeline.Pipe(
			pipeline.PreDispatcherStage(),
			pipeline.DirectDispatcherStage(p.inference, logger),
		)
	}
```

It does not add much to readability and it can be reintroduced later

internal/processor/pipeline/dispatcher_direct.go
```diff
@@ -0,0 +1,176 @@
...
+		handleClientError(&result, msg.ModelID, clientErr, logger)
+	case resp == nil:
+		result.Error = &OutputError{Code: string(httpclient.ErrCategoryServer), Message: "inference returned no response"}
+	default:
+		handleSuccess(&result, msg.ModelID, resp, logger)
+	}
+
+	isSuccess := result.Error == nil && result.Response != nil && result.Response.StatusCode == 200
+	if !isSuccess {
+		metrics.RecordRequestError(msg.ModelID)
```
lioraron: Bug: `RecordRequestError` is double-counted in the sync flow.

`buildResult()` calls it here for every non-200 result, and then `collector.go:120` calls it again for the same result in `Receive()`. The result flows through the pipeline chain to the collector, so both fire → error rate metrics are 2x inflated for all sync-mode jobs.

Fix: remove this call from `buildResult()` — the collector already handles `RecordRequestError` for both sync and async flows.

internal/processor/pipeline/dispatcher_direct.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	}
+
+	result.Response = &batch_types.ResponseData{
+		StatusCode: 200,
+		RequestID:  resp.RequestID,
+		Body:       body,
+	}
+
+	recordTokenUsage(body, modelID, logger)
```
lioraron: Bug: `recordTokenUsage` is double-counted in the sync flow — same pattern as `RecordRequestError`.

`handleSuccess()` calls it here, and then `collector.go:110` calls it again for the same successful result. Both fire for every 200 response → token usage metrics (prompt_tokens, completion_tokens) are 2x inflated for sync-mode jobs.

Fix: remove this call from `handleSuccess()` — the collector handles it for both flows.

internal/processor/pipeline/dispatcher_direct.go
```diff
@@ -0,0 +1,176 @@
...
+	result := buildResult(msg, resp, clientErr, d.logger)
+	if ctx.Err() != nil && result.Error != nil && result.Response == nil {
+		code, message := cancelCode(ctx)
+		result.Error = &OutputError{Code: code, Message: message}
+	}
+	resultCh <- result
+}
+
+func buildResult(msg RequestItem, resp *inference.GenerateResponse, clientErr *inference.ClientError, logger logr.Logger) ResultItem {
+	result := ResultItem{
```
lioraron, line 82: Fragile implicit contract: `buildResult()` intentionally does NOT copy `SubmittedAt` from `RequestItem` to `ResultItem`. This is what prevents the collector from double-decrementing inflight metrics (the collector guards Dec calls with `!msg.SubmittedAt.IsZero()`). But it's undocumented — if someone later adds `SubmittedAt: msg.SubmittedAt` here, inflight metrics will silently double-decrement.

Worth adding a short comment explaining why `SubmittedAt` is not copied (DirectDispatcher handles its own Dec/Duration in `submitRequest()`, so the collector must skip them).

internal/processor/pipeline/dispatcher_aimd.go
```diff
@@ -0,0 +1,149 @@
...
+
+var _ RequestDispatcher = (*AIMDDispatcher)(nil)
+
+func NewAIMDDispatcher(
+	next RequestDispatcher,
+	models map[string]*EndpointAIMD,
+	globalLimit int,
+	logger logr.Logger,
+) *AIMDDispatcher {
+	globalSem, _ := semaphore.New(globalLimit, nil)
```
lioraron: `semaphore.New` returns `(nil, ErrCap)` when `capacity <= 0`. The error is discarded here, so a misconfigured `globalLimit` (e.g. `concurrency.global: 0`) would leave `globalSem` nil → panic at `acquireSlot`.

Config validation may ensure this doesn't happen today, but a defensive check would prevent a cryptic nil-pointer panic:
```go
globalSem, err := semaphore.New(globalLimit, nil)
if err != nil {
    // return error or log.Fatal with a clear message about the config field
}
```

lioraron:
**source_planfile.go:82** — `Produce` ignores context cancellation

After context cancellation (SLO expiry, user cancel), the `Produce` loop continues doing `ReadAt` + `json.Unmarshal` for every remaining plan entry before sending it into the channel. PreDispatcher will drain everything and convert to cancel errors, so there's no deadlock, but for a large job with many remaining requests this delays shutdown with unnecessary I/O.

Consider adding a ctx check:
```go
select {
case outgoingRequestCh <- *item:
case <-ctx.Done():
    return nil
}
```

lioraron:
**collector.go:104** — `RecordModelRequestExecutionDuration` fires for requests that never ran inference

This fires for any result with non-zero `SubmittedAt` — including cancel errors from `AIMDDispatcher.acquireSlot` failures. Those requests never ran inference; the recorded duration is just AIMD semaphore queue wait time, which pollutes the execution duration histogram.

Consider guarding:
```go
if msg.Error == nil {
    metrics.RecordModelRequestExecutionDuration(time.Since(msg.SubmittedAt), msg.ModelID)
}
```
(The Dec calls should remain unconditional since PreDispatcher incremented for all submitted requests.)

internal/processor/pipeline/collector.go
```diff
@@ -70,7 +70,10 @@ func (c *ResultCollector) Drain(ctx context.Context, resultCh <-chan ResultItem)
 	if firstErr != nil {
 		return firstErr
 	}
-	return ctx.Err()
+	if ctx.Err() != nil {
+		c.logger.Info("Context cancelled after drain completed", "err", ctx.Err())
+	}
+	return nil
```
j-mok-dev, line 83: If `Receive()` has already hit a write error, the collector keeps draining until dispatch finishes and only returns that error at the end. In the unified pipeline that means we can continue issuing/completing backend work after output persistence has already failed. Should this fail fast and cancel the pipeline instead, since at that point we can no longer durably record subsequent results?

internal/processor/pipeline/dispatcher_aimd.go
evacchi: ok!

internal/processor/pipeline/dispatcher_direct.go
evacchi: added a test to verify, and moved all logic related to metrics in the collector, as suggested below: now disapatcher_direct no longer deals with metrics (and no more double counting!)

internal/processor/pipeline/dispatcher_direct.go
evacchi: done!

internal/processor/pipeline/dispatcher_direct.go
evacchi: recordTokenUsage is only called in handleSuccess() and we removed it (see above)

internal/processor/pipeline/collector.go
evacchi: this is intentional and, at this time, it is also safe. Adding this comment on `Drain()` for clarity:

```go
// On a write error, Drain continues reading but skips further writes. This
// is intentional: the dispatcher sends results to resultCh and blocks if
// nobody reads; returning early would deadlock the pipeline. The error
// is returned after the channel closes, causing executeJobAsync to route
// the job to handleFailed (which uploads whatever partial output is on disk).
func (c *ResultCollector) Drain(ctx context.Context, resultCh <-chan ResultItem) error {
```

we can investigate mechanisms to return early, but I think we can do this safely in a follow up, maybe we can create an issue

evacchi:
> **source_planfile.go:82** — `Produce` ignores context cancellation
> 
> After context cancellation (SLO expiry, user cancel), the `Produce` loop continues doing `ReadAt` + `json.Unmarshal` for every remaining plan entry before sending it into the channel.

as discussed above, draining instead of immediate close is intentional, but indeed I/O and `ReadAt+json.Unmarshal` can be skipped so that draining happens fast, and no I/O wastes our time. I'll add this behavior + test.

We can also revisit this Produce + Drain behavior in a follow up and investigate short-circuiting!

internal/processor/pipeline/collector.go
j-mok-dev: Makes sense to me. In the current pipeline shape, returning early from the collector would risk blocking producers that are still sending on `resultCh`, so I agree this is better handled as a follow-up.
I think the right follow-up is to cancel dispatch on the first persistence failure while still draining results until the channel closes, rather than trying to make the collector return immediately. I'll create the issue. thanks.

j-mok-dev (approved):

internal/processor/worker/source_planfile.go
```diff
@@ -75,6 +79,13 @@ func (s *PlanFileSource) Produce(ctx context.Context, outgoingRequestCh chan<- p
 		}
 
 		for _, entry := range entries {
+			if ctx.Err() != nil {
+				outgoingRequestCh <- pipeline.RequestItem{
+					RequestID: fmt.Sprintf("batch_req_%s", uuid.NewString()),
```
lioraron: Bug: the minimal item has empty `CustomID`, which causes `Resolve()` to silently drop it.

The collector's `Resolve()` checks: `result.Error != nil` → true → `LoadAndDelete(RequestID)` → miss (never stored) → `result.CustomID != ""` → **false** → returns false → item dropped.

These cancelled items are never written to the error file and never tracked by the progress tracker. On a large job cancelled at 50%, `completed + failed < total` in the final batch counts.

Fix: set `CustomID` on the minimal item (same pattern as parse errors at line 113):
```go
reqID := fmt.Sprintf("batch_req_%s", uuid.NewString())
outgoingRequestCh <- pipeline.RequestItem{
    RequestID: reqID,
    CustomID:  reqID,
    ModelID:   modelID,
}
```

lioraron:
Follow-up on [this earlier comment](https://github.com/llm-d/llm-d-batch-gateway/pull/584#issuecomment-4993547350) about `collector.go` `RecordModelRequestExecutionDuration`:

With the double-counting fix (`746b1cd`), `buildResult` now copies `SubmittedAt` — so the collector records execution duration for **all** results uniformly, including cancel errors from `AIMDDispatcher.acquireSlot` failures where no inference ran. The recorded duration for those is just AIMD semaphore queue wait time.

This is lower priority and could be a follow-up, but worth tracking — the execution duration histogram would be polluted during cancellation/expiry scenarios.

internal/processor/worker/source_planfile.go
evacchi: excellent! fixing. Yet again let's add a follow up about the RequestItem's lifecycle/contract, these seems like things that should be strictly encoded by methods/constructors/predicates

evacchi: done

lioraron (approved):

PR #590, opened 2026-07-16
fix(processor): propagate async ResultMessage HTTP status to batch

### LINKED ISSUE #589: Async dispatch does not propagate HTTP status code from async-processor result
## Problem

When using async dispatch mode (`dispatchMode: async`), the batch-gateway processor does not propagate the HTTP status code returned by the async-processor. This causes authorization failures (HTTP 403) to be reported as `parse_error` instead of preserving the original status code.

### Expected behavior

When an async-processor receives a 403 from the inference gateway (e.g., due to unauthorized credentials), the batch output should contain `"status_code": 403`, matching the sync dispatch behavior.

### Actual behavior

The batch output contains:
```json
{"error":{"code":"parse_error","message":"response body could not be parsed: unexpected end of JSON input"}}
```

## Root cause

Two issues in the async result handling chain:

### 1. `GetResult()` discards `StatusCode`

`pkg/clients/inference/async_shared_client.go:80-83`:

```go
return &GenerateResponse{
    RequestID: result.ID,
    Response:  []byte(result.Payload),
    // result.StatusCode is discarded
}, nil
```

The `GenerateResponse` struct has no `StatusCode` field, so `ResultMessage.StatusCode` from the llm-d-async API is lost.

### 2. `asyncResult()` hardcodes `StatusCode: 200`

`internal/processor/pipeline/result_router.go:152-153`:

```go
result.Response = &batch_types.ResponseData{
    StatusCode: 200, // always 200, regardless of actual HTTP status
    RequestID:  resp.RequestID,
    Body:       body,
}
```

### Full request flow

```
batch-gateway processor → Redis queue (with user's Authorization header)
→ async-processor → Internal Gateway (AuthPolicy) → 403
→ async-processor writes ResultMessage{StatusCode: 403, Payload: ""} to Redis
→ batch-gateway GetResult() reads result, discards StatusCode  ← Bug 1
→ asyncResult() tries json.Unmarshal on empty body → parse_error  ← Bug 2
```

The async-processor side is correct — `NewHTTPResult()` in `llm-d-async/api` properly sets `StatusCode: 403`. The status code is lost on the batch-gateway consumer side.

## Suggested fix

1. Add `StatusCode int` field to `GenerateResponse`
2. In `GetResult()`, propagate `result.StatusCode` to `GenerateResponse.StatusCode`
3. In `asyncResult()`, use the actual `StatusCode` instead of hardcoding 200; for non-2xx responses with unparseable body, still preserve the status code in the output

j-mok-dev:
## Why is this PR needed?
When `dispatchMode: async`, HTTP error statuses from the async-processor (e.g. 403 with an empty body) were lost on the batch-gateway consumer side. Batch output then reported `parse_error` instead of preserving `status_code: 403`, which broke sync/async parity for authorization failures.
Upstream llm-d-async already returns `ResultMessage.StatusCode` / error fields (llm-d/llm-d-async#298). This PR consumes those fields.
## What does this PR do?
- Add `StatusCode`, `ErrorCode`, and `ErrorMessage` to `GenerateResponse`
- Propagate those fields from `ResultMessage` in `GetResult()`
- Update `asyncResult()` to:
  - use the real HTTP status instead of hardcoding 200
  - preserve non-2xx status even when the body is empty or non-JSON
  - map non-HTTP failures (`StatusCode == 0` + error metadata) to output errors
  - treat legacy payloads with no status as HTTP 200
- Unit coverage for status/error propagation
- Dispatcher e2e (`HTTPErrorStatusPreserved`) using a consumer-less `sim-pool-inject` pool so a synthetic `ResultMessage{StatusCode: 403, Payload: ""}` can be injected without racing the async-processor (Kind has no AuthPolicy)
## How was this tested?
## Checklist
## Related Issues
Fixes #589

### DIFF
--- a/internal/processor/pipeline/result_router.go
+++ b/internal/processor/pipeline/result_router.go
@@ -136,21 +136,59 @@ func asyncResult(resp *inference.GenerateResponse, logger logr.Logger) ResultIte
 	result := ResultItem{
 		RequestID: resp.RequestID,
 	}
-	if resp.Response == nil {
-		result.Error = &OutputError{Code: "server_error", Message: "async response has no body"}
+
+	if resp.IsNonHTTPFailure() {
+		code := resp.ErrorCode
+		if code == "" {
+			code = "server_error"
+		}
+		msg := resp.ErrorMessage
+		if msg == "" {
+			msg = "async request failed"
+		}
+		result.Error = &OutputError{Code: code, Message: msg}
 		return result
 	}
-	var body map[string]any
-	if err := json.Unmarshal(resp.Response, &body); err != nil {
-		logger.Error(err, "Failed to unmarshal async response", "requestID", resp.RequestID)
-		result.Error = &OutputError{
-			Code:    "parse_error",
-			Message: fmt.Sprintf("response body could not be parsed: %v", err),
+
+	statusCode := resp.StatusCode
+	if statusCode == 0 {
+		// Legacy ResultMessage with only Payload — treat as HTTP 200.
+		statusCode = 200
+	}
+
+	if len(resp.Response) == 0 {
+		if statusCode >= 200 && statusCode < 300 {
+			result.Error = &OutputError{Code: "server_error", Message: "async response has no body"}
+			return result
+		}
+		// Non-2xx with empty body (e.g. 403 from auth): preserve status code.
+		result.Response = &batch_types.ResponseData{
+			StatusCode: statusCode,
+			RequestID:  resp.RequestID,
+			Body:       map[string]any{},
 		}
 		return result
 	}
+
+	var body map[string]any
+	if err := json.Unmarshal(resp.Response, &body); err != nil {
+		if statusCode >= 200 && statusCode < 300 {
+			logger.Error(err, "Failed to unmarshal async response", "requestID", resp.RequestID)
+			result.Error = &OutputError{
+				Code:    "parse_error",
+				Message: fmt.Sprintf("response body could not be parsed: %v", err),
+			}
+			return result
+		}
+		// Non-2xx with unparseable body: still preserve the HTTP status.
+		body = map[string]any{
+			"error": map[string]any{
+				"message": string(resp.Response),
+			},
+		}
+	}
 	result.Response = &batch_types.ResponseData{
-		StatusCode: 200,
+		StatusCode: statusCode,
 		RequestID:  resp.RequestID,
 		Body:       body,
 	}
--- a/internal/processor/pipeline/async_test.go
+++ b/internal/processor/pipeline/async_test.go
@@ -55,8 +55,9 @@ func (c *fakeAsyncClient) Close() error { return nil }
 func (c *fakeAsyncClient) deliver(requestID string, body map[string]any) {
 	resp, _ := json.Marshal(body)
 	c.results <- &inference.GenerateResponse{
-		RequestID: requestID,
-		Response:  resp,
+		RequestID:  requestID,
+		Response:   resp,
+		StatusCode: 200,
 	}
 }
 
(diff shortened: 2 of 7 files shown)


scripts/dev-deploy-dispatcher.sh
```diff
@@ -11,10 +11,11 @@ fi
 # ── Configuration ────────────────────────────────────────────────────────────
 KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-batch-gateway-dev}"
 DISPATCHER_RELEASE="${DISPATCHER_RELEASE:-dispatcher}"
-DISPATCHER_VERSION="${DISPATCHER_VERSION:-v0.7.3}"
+_ASYNC_MOD_VERSION=$(grep 'llm-d-async/api' "${REPO_ROOT}/go.mod" | awk '{print $2}')
+DISPATCHER_VERSION="${DISPATCHER_VERSION:-${_ASYNC_MOD_VERSION}}"
 DISPATCHER_IMAGE="${DISPATCHER_IMAGE:-ghcr.io/llm-d-incubation/llm-d-async:${DISPATCHER_VERSION}}"
```
yizhaodev, line 15: chart and image not changed after graduation?

j-mok-dev: The v0.7.4 container image hasn't been published yet — the [release workflow failed](https://github.com/llm-d/llm-d-async/actions/runs/29516135464) The chart OCI path at `ghcr.io/llm-d/charts/async-processor` also returns 404 currently.
Once the upstream image is available at `ghcr.io/llm-d/llm-d-async:v0.7.4`, I'll update the script to point there. Keeping it at the working v0.7.3 incubation path for now so dev-deploy and e2e don't break.

yizhaodev: it seems chart has moved
```
# helm show chart oci://ghcr.io/llm-d-incubation/charts/async-processor --version 0.7.4
Error: failed to perform "FetchReference" on source: ghcr.io/llm-d-incubation/charts/async-processor:0.7.4: not found

# helm show chart oci://ghcr.io/llm-d/charts/async-processor --version 0.7.4
Pulled: ghcr.io/llm-d/charts/async-processor:0.7.4
Digest: sha256:34023f82d2a40f245b1f1ebebfaeaa5e721190582d8016594fa6a6ff7989e195
apiVersion: v2
appVersion: v0.7.4
description: Helm chart for Async-Processor (AP) - Asynchronous Processor for LLM
  inference workloads
name: async-processor
type: application
version: 0.7.4
```

internal/processor/pipeline/result_router.go
```diff
@@ -136,21 +136,60 @@ func asyncResult(resp *inference.GenerateResponse, logger logr.Logger) ResultIte
 	result := ResultItem{
 		RequestID: resp.RequestID,
 	}
-	if resp.Response == nil {
-		result.Error = &OutputError{Code: "server_error", Message: "async response has no body"}
+
+	// Non-HTTP failure: StatusCode == 0 with error metadata from llm-d-async.
+	if resp.StatusCode == 0 && (resp.ErrorCode != "" || resp.ErrorMessage != "") {
```
evacchi: maybe let's turn this into a predicate on `ResponseItem`

evacchi (approved):
only a nit, other than that 👍

scripts/dev-deploy-dispatcher.sh
yizhaodev: maybe stay with 0.7.3 and we merge this
once they fixed 0.7.4, we upgrade

evacchi: yes if we can stay on 0.7.3 (pre-rename?) the diff should be smaller as swell

j-mok-dev: Thanks @evacchi  @yizhaodev  I've dropped the v0.7.4 upgrade commit from this branch.
This PR now only contains the status-code propagation fix + e2e test. Will open a follow-up PR for the module migration once the upstream image is published.

yizhaodev (approved):

internal/processor/pipeline/result_router.go
j-mok-dev: Thanks! added `IsNonHTTPFailure()` on `GenerateResponse` and switched the callsite to use it.

PR #591, opened 2026-07-16
fix(pipeline): decrement inflight metrics before write-failure guard

madhugoutham:
## Why is this PR needed?

After a persistence failure in `Drain()`, `firstErr` is set and `Receive()` is skipped for all remaining results. On the async path, `Receive()` at line 101-103 was the only place inflight gauges (`DecProcessorInflightRequests`, `DecModelInflightRequests`) and execution duration got decremented. After a write failure, remaining async results never decrement counters leak permanently.

## What does this PR do?

Moves the metrics decrement block from `Receive()` into `Drain()`, before the `firstErr` guard. This ensures counters are always cleaned up regardless of write success.

No behavioral change for the happy path same metrics, same timing, same values.

## How was this tested?

- [x] Unit test added: `TestResultCollector_DrainDecrementsMetricsAfterWriteFailure`
      (injects write failure, verifies all pending requests resolved)
- [x] Full test suite with race detector: 27/27 packages pass
- [x] Kind E2E: all tests pass
- [x] CKS GPU (Waldorf): 500/500 requests, Qwen2.5-7B on H200, 0 drops, 10s

## Checklist

### DIFF
--- a/internal/processor/pipeline/collector.go
+++ b/internal/processor/pipeline/collector.go
@@ -56,6 +56,11 @@ func (c *ResultCollector) Drain(ctx context.Context, resultCh <-chan ResultItem)
 		if !c.pending.Resolve(&msg) {
 			continue
 		}
+		if !msg.SubmittedAt.IsZero() {
+			metrics.DecProcessorInflightRequests()
+			metrics.DecModelInflightRequests(msg.ModelID)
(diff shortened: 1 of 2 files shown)


lioraron (approved):
Thanks!

PR #600, opened 2026-07-17
feat(benchmarks): implement scenario 5 async-processor deployment

lioraron:
## Summary
- Replaces the S5 "blocked on integration" stub with a working async-processor deployment using `llm-d-async` with endpoint-scrape gate
- Adds async-processor env vars (`DISPATCHER_VERSION`, `DISPATCHER_CHART`, etc.) to `setup.sh` with defaults pinned to v0.7.3
- Handles both sim mode (fake-metrics on inference-sim, `sim-pool` queue name) and GPU mode (vllm-metrics Service, `optimized-baseline` queue name)
- Removes the early-return stub in `benchmark.py` and adds S5 narrative generation
- Updates `README.md` scenario table and env var docs

## Test plan
- [x] `bash -n benchmarks/setup.sh` — syntax check (verified locally)
- [x] `python3 -m py_compile benchmarks/benchmark.py` — syntax check (verified locally)
- [x] Local sim-mode (Kind cluster):
  ```bash
  make dev-deploy                          # create Kind cluster + load images (if not already running)
  BENCHMARK_SCENARIO=5 make benchmark-local  # deploy S5 with inference-sim + async-processor, run benchmark
  ```
  Verify: inference-sim gets `--fake-metrics`, async-processor pod starts, batch-gateway processor in async mode, batch jobs make progress
- [ ] GPU-mode: `SCENARIO=5 KUBE_CONTEXT=<gpu-ctx> ./benchmarks/setup.sh` — verify vllm-metrics Service created, async-processor connects to Redis
- [ ] Teardown: `SCENARIO=5 KUBE_CONTEXT=<gpu-ctx> ./benchmarks/teardown.sh` — verify async-processor and vllm-metrics cleaned up


benchmarks/setup.sh
```diff
@@ -529,8 +553,67 @@ fi
...
+        -n "${NAMESPACE}" \
+        --set "ap.image.repository=${DISPATCHER_IMAGE_REPO}" \
+        --set "ap.image.tag=${DISPATCHER_IMAGE_TAG}" \
+        --set ap.messageQueueImpl=redis-sortedset \
+        --set ap.concurrency=1 \
+        --set ap.redis.enabled=true \
+        --set "ap.redis.url=redis://redis-master.${NAMESPACE}.svc.cluster.local:6379" \
+        --set ap.redis.pollIntervalMs=500 \
+        --set ap.redis.batchSize=10 \
+        --set-json "ap.redis.queuesConfig=[{\"queue_name\":\"llm-d-async:requests:${ASYNC_POOL_NAME}\",\"result_queue_name\":\"llm-d-async:results:${ASYNC_POOL_NAME}\",\"request_path_url\":\"/v1/completions\",\"igw_base_url\":\"${ASYNC_IGW_URL}\",\"gate_type\":\"endpoint-scrape\",\"gate_params\":{\"url\":\"${ASYNC_METRICS_URL}\",\"metric\":\"vllm:num_requests_waiting\",\"max_count_per_pod\":\"5\",\"fallback\":\"1.0\"}}]" \
```
j-mok-dev: nit: `request_path_url` is set to `/v1/completions` but the benchmark submits batch requests with `"url": "/v1/chat/completions"` (in both `benchmark.py` and `batch-submit.yaml`). Should this be `/v1/chat/completions` to match?

benchmarks/setup.sh
```diff
@@ -529,8 +553,67 @@ fi
...
+    - port: 8000
+      targetPort: 8000
+      name: http
+EOVLLMSVC
+    fi
+
+    DISPATCHER_HELM_VERSION_FLAG=()
+    if [ -n "${DISPATCHER_SOURCE}" ] && [ -d "${DISPATCHER_SOURCE}" ]; then
+        DISPATCHER_CHART="${DISPATCHER_SOURCE}/charts/async-processor"
+        DISPATCHER_IMAGE="ghcr.io/llm-d-incubation/async-processor:dev-local"
```
j-mok-dev: I think the `DISPATCHER_SOURCE` path is still incomplete here.

When `DISPATCHER_SOURCE` is set, this switches the chart to the local `async-processor` checkout and sets the image to `ghcr.io/llm-d-incubation/async-processor:dev-local`, but I don't see a corresponding local image build/load step in this script. So the documented local-source workflow looks likely to fail with `ImagePullBackOff` unless that image already exists and is reachable by the target cluster.

Would it make sense to either:
1. add the local build/load path here, or
2. drop `DISPATCHER_SOURCE` from the docs/env contract in this PR and keep the scenario on the pinned OCI image path only?

benchmarks/setup.sh
lioraron: Fixed, thanks!

benchmarks/setup.sh
lioraron: Fixed, thanks!

j-mok-dev (approved):

PR #617, opened 2026-07-23
feat(tls): add `--tls-min-version` and `--tls-cipher-suites` CLI flags to API server

vishbhat:
## Why is this PR needed?

The batch-gateway API server hardcodes `tls.VersionTLS12` as the minimum TLS version when TLS is enabled, with no cipher suite control and no ALPN (`NextProtos`). This means the server cannot honor the cluster TLS profile — if an administrator sets the profile to Modern or Custom, the batch-gateway ignores it. Per OCP TLS guidance, all TLS servers must respect the configurable cluster profile. The deploying operator needs a mechanism to project profile values into the batch-gateway at runtime.

## What does this PR do?

Adds `--tls-min-version` and `--tls-cipher-suites` CLI flags to the API server so the deploying operator can pass cluster TLS profile values as container args.

- **`--tls-min-version`**: Accepts `VersionTLS12` (default) or `VersionTLS13`. Empty string preserves the existing TLS 1.2 floor for backward compatibility.
- **`--tls-cipher-suites`**: Accepts a comma-separated list of secure cipher suite names (from `tls.CipherSuites()`). Rejects insecure suites explicitly. Empty string uses Go's default secure cipher list.
- Sets `NextProtos: ["h2", "http/1.1"]` for proper HTTP/2 ALPN negotiation when TLS is enabled.
- Warns and clears cipher suites when combined with TLS 1.3 (Go ignores `CipherSuites` for TLS 1.3 per spec).
- TLS flag validation only runs when SSL is enabled (`SSLCertFile` + `SSLKeyFile` provided), avoiding false errors or silent misconfiguration on plain-HTTP deployments.

## How was this tested?

## Checklist

### DIFF
--- a/internal/apiserver/common/config.go
+++ b/internal/apiserver/common/config.go
@@ -23,6 +23,7 @@ import (
 	"os"
 
 	sharedcfg "github.com/llm-d/llm-d-batch-gateway/internal/shared/config"
+	utls "github.com/llm-d/llm-d-batch-gateway/internal/util/tls"
 	"gopkg.in/yaml.v3"
 	"k8s.io/klog/v2"
 )
@@ -110,6 +111,8 @@ type ServerConfig struct {
 	ObservabilityPort string            `yaml:"observability_port"`
 	SSLCertFile       string            `yaml:"ssl_cert_file"`
 	SSLKeyFile        string            `yaml:"ssl_key_file"`
+	TLSMinVersion     string            `yaml:"-"`
+	TLSCipherSuites   string            `yaml:"-"`
 	InputHeaders      map[string]string `yaml:"input_headers"`
 
 	// HTTP server timeout configurations (in seconds)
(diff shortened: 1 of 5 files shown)


j-mok-dev (approved):
lgtm.
nit: `ParseMinVersion`/`ParseCipherSuites` are called both in `Validate()` and `Start()`. maybe we can store the parsed values in config during validation to avoid redundant parsing in Start(). not blocking.

vishbhat:
@j-mok-dev 
Good catch. My claude review flagged the same thing. The double-parse is intentional as a safety net: `Start()` can be called from tests that construct `ServerConfig` directly without going through `Load()` → `Validate()`. Since both calls only happen once at startup under `SSLEnabled()`, the cost is negligible. Either way is reasonable.

yizhaodev:
@vishbhat 
after quick glance, i don't see any changes in helm chart?
operator need to pass the new params via helm chart

vishbhat:
Duplicate of https://github.com/llm-d/llm-d-batch-gateway/pull/618/ , closing.

PR #621, opened 2026-07-26
refactor(demo): extract verify_dispatcher_runtime to common.sh

yizhaodev:
## What does this PR do?

- Move `verify_dispatcher_runtime` from `deploy-k8s.sh` to `common.sh`, parameterize it for reuse
- Consolidate Prometheus metric checks into the runtime verification function
- Rename `install_prometheus` → `install_prometheus_for_dispatcher` for clarity
- Add `--fake-metrics` to decode simulator so Prometheus verification finds data

## How was this tested?

### DIFF
--- a/examples/deploy-demo/deploy-k8s.sh
+++ b/examples/deploy-demo/deploy-k8s.sh
@@ -653,7 +653,7 @@ uninstall_llmd() {
 
 # ── Async Dispatcher (llm-d-async) ───────────────────────────────────────────
 
-install_prometheus() {
+install_prometheus_for_dispatcher() {
     local prom_name="prometheus"
 
     if kubectl get deployment "${prom_name}" -n "${LLM_NAMESPACE}" &>/dev/null; then
@@ -771,7 +771,7 @@ EOF
 deploy_dispatcher() {
     banner "Installing Async Dispatcher (llm-d-async)"
 
-    install_prometheus
+    install_prometheus_for_dispatcher
 
     # Reuse batch-internal-gateway: async-processor passes through the user's
     # Authorization header (captured at batch creation), so AuthPolicy works.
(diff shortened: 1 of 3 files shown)


examples/deploy-demo/llmd-sim/modelserver/decode-deployment.yaml
```diff
@@ -35,6 +35,8 @@ spec:
             - "random"
             - "--served-model-name"
             - "random"
+            - "--fake-metrics"
```
j-mok-dev, line 38: Enabling `--fake-metrics` here looks risky because the dispatcher's `prometheus-budget` gate reads these same metrics, but the demo flow never updates `/fake_metrics`. That means the simulator will keep reporting the static zero values from this JSON, so the gate can stop reflecting real backend load and stay effectively "open" under traffic.

examples/deploy-demo/deploy-k8s.sh
```diff
@@ -893,111 +893,13 @@ verify_dispatcher_config() {
         errors=$((errors + 1))
     fi
 
-    # 4. Prometheus has the metrics that prometheus-budget gate needs
```
j-mok-dev, line 896: This seems to move the Prometheus verification out of the install path and into `cmd_test()`. Does that mean `./deploy-k8s.sh install` no longer fails fast when the gate's required series are missing?

examples/deploy-demo/llmd-sim/modelserver/decode-deployment.yaml
yizhaodev: without --fake-metrics, simulator will not report metrics
i need to prove the whole flow is working with promethues-budget.

examples/deploy-demo/deploy-k8s.sh
yizhaodev: i need to verify the prometheus really able to collect metrics from epp/vllm
verfiy right after install may fail because scraping is not started yet.
so move to cmd_test for safer consideration

examples/deploy-demo/llmd-sim/modelserver/decode-deployment.yaml
j-mok-dev: Makes sense for scrape visibility. My concern is that in this demo we never mutate `/fake_metrics`, so the gate may end up reading a permanent zero signal.

examples/deploy-demo/deploy-k8s.sh
j-mok-dev: make sense. thanks for clarifying.

examples/deploy-demo/llmd-sim/modelserver/decode-deployment.yaml
yizhaodev: i understand. this is demo for setup, just want to verify the install command/config is correct

j-mok-dev (approved):

PR #624, opened 2026-07-29
test(e2e): add GIE flow control saturation scenarios

j-mok-dev:
## Why is this PR needed?

Issue #402 asked for GIE flow-control e2e beyond the existing EPP routing smoke tests: saturation shedding, priority bands, mixed-load metrics, and SLO-deadline ordering.

## What does this PR do?

Adds the remaining GIE flow-control scenarios that we can assert on the current Kind pin (GIE EPP v1.5.0), plus the harness to drive saturation deterministically.

- Saturation / recovery for batch traffic through EPP
- Priority-band coverage for interactive vs batch under pool saturation
- Mixed-load coverage for EPP-side and processor-side metrics
- Enable `--fake-metrics` on the GIE simulator path so tests can control saturation via `/fake_metrics`
- Quote simulator extra args in `dev-deploy.sh` so JSON-valued arguments stay strings in the generated manifest
- Increase the e2e timeout for saturation / TTL scenarios
- Raise `curlEPP --max-time` from 35s to 65s (EPP `defaultRequestTTL` 30s plus buffer) so EvictedTTL is not reported as a transport `"000"`

SLO-deadline dispatch order is **not** asserted. On GIE EPP v1.5.0, queued sheddable requests still dispatch first-come, first-served even with `slo-deadline-ordering-policy` and `x-slo-ttft-ms` on the wire. Batch `CompletedAt` also cannot observe EPP order (second granularity, parallel sim completion). A submit-gap sleep would put the long-deadline request in-flight first, so EPP never has both items queued to reorder. Restore that e2e after Kind deploys llm-d-router; do not bump `GIE_VERSION` to v1.6.0 (standalone EPP chart is already gone). Remaining-ms header formatting stays in `TestMergeHeaders`. There is a short coverage note in `test/e2e/flow_control_test.go`.

Saturation assertions check for **non-dispatch** rather than a single outcome label, because EPP may surface `EvictedTTL` or `DroppedOnSaturation` depending on queue state and timing.

## How was this tested?

## Checklist

## Related Issues

Related #402 (SLO-deadline ordering deferred until Kind e2e uses llm-d-router)


test/e2e/flow_control_test.go
```diff
@@ -432,8 +429,373 @@ func doTestBatchCompletionThroughEPP(t *testing.T) {
...
+	out, err := exec.Command("kubectl", "exec",
+		"-n", testNamespace,
+		e2eCurlPod,
+		"--",
+		"curl", "-sS", "-X", "POST",
+		"-H", "Content-Type: application/json",
+		"-H", fmt.Sprintf("x-gateway-inference-objective: %s", objective),
+		"-d", body,
+		"-w", "\n%{http_code}",
+		"--max-time", "35",
```
lioraron: The EPP `defaultRequestTTL` is 30s, so under saturation a queued request won't return until ~30s after it's admitted. `--max-time 35` leaves only 5s of slack — any cluster or scrape latency could push this over, causing `curlEPP` to return `"000"` (transport error). In `doTestPriorityBandInteraction` phase 2, a `"000"` response hits the `t.Fatalf` arm rather than the asserting arm, silently aborting phases 3 and 4 without a useful failure message. Consider 65s (TTL + a generous buffer).

test/e2e/flow_control_test.go
```diff
@@ -432,8 +429,373 @@ func doTestBatchCompletionThroughEPP(t *testing.T) {
...
+	}
+
+	if shortCompletedAt == 0 {
+		t.Fatal("short-deadline batch did not complete within timeout")
+	}
+	if longCompletedAt == 0 {
+		t.Fatal("long-deadline batch did not complete within timeout")
+	}
+
+	if shortCompletedAt > longCompletedAt {
```
lioraron: `CompletedAt` has second granularity and the assertion uses `>` (fails only if short is *strictly later* than long), so a same-second tie passes even if dispatch order was violated. With 3s inter-token-latency and `max_tokens=3` (~9s per request), sequential dispatch makes a tie unlikely, but it's not impossible on a fast/bursty cluster. A small sleep between `batchLong` and `batchShort` submissions (e.g. 2s) would force the short-deadline batch to always have a later submit time, making the EPP's ordering preference deterministically observable.

test/e2e/flow_control_test.go
j-mok-dev: Agreed. `curlEPP` now uses `--max-time 65` (EPP `defaultRequestTTL` 30s plus buffer) so a saturation-path EvictedTTL response is not cut off as `"000"`.

test/e2e/flow_control_test.go
j-mok-dev: A 2s gap between submits would let the long-deadline request go in-flight first, so EPP never has both items queued to reorder. `CompletedAt` is also the wrong signal: second granularity plus parallel sim completion, so a same-second tie can pass even when dispatch is first-come, first-served.
I tried asserting EPP enqueue/dispatch order with both requests queued first. On GIE v1.5.0 the overlay has `slo-deadline-ordering-policy` and Envoy does forward `x-slo-ttft-ms`, but dispatch is still first-come, first-served. That is an EPP queue contract, not something this gateway can assert on the current Kind pin, so I dropped `SLODeadlineOrdering`.
The test needs to be restored after Kind e2e deploys llm-d-router rather than bumping `GIE_VERSION`. v1.6.0 already dropped the standalone EPP chart. Remaining-ms header formatting stays in `TestMergeHeaders`. I left a TODO on the GIE coverage comment in `flow_control_test.go`.

scripts/dev-deploy.sh
```diff
@@ -1456,8 +1457,17 @@ main() {
...
+        # GIE flow control tests need --fake-metrics to dynamically control
+        # EPP utilization-detector saturation at runtime via /fake_metrics POST.
+        install_vllm_sim "${VLLM_SIM_NAME}" "${VLLM_SIM_MODEL}" "50ms" "100ms" \
+            "--fake-metrics" '{"kv-cache-usage": 0, "waiting-requests": 0, "running-requests": 0}'
+        install_vllm_sim "${VLLM_SIM_B_NAME}" "${VLLM_SIM_B_MODEL}" "200ms" "500ms" \
+            "--fake-metrics" '{"kv-cache-usage": 0, "waiting-requests": 0, "running-requests": 0}'
+    else
+        install_vllm_sim "${VLLM_SIM_NAME}" "${VLLM_SIM_MODEL}" "50ms" "100ms"
+        install_vllm_sim "${VLLM_SIM_B_NAME}" "${VLLM_SIM_B_MODEL}" "200ms" "500ms"
+    fi
```
zdtsw, line 1465: nit not blocker:
```suggestion
fake_metrics_args=()
if [ "${ENABLE_GIE}" = "true" ]; then
    # GIE flow control tests need --fake-metrics to dynamically control
    # EPP utilization-detector saturation at runtime via /fake_metrics POST.
    fake_metrics_args=("--fake-metrics" '{"kv-cache-usage": 0, "waiting-requests": 0, "running-requests": 0}')
fi

install_vllm_sim "${VLLM_SIM_NAME}" "${VLLM_SIM_MODEL}" "50ms" "100ms" "${fake_metrics_args[@]}"
install_vllm_sim "${VLLM_SIM_B_NAME}" "${VLLM_SIM_B_MODEL}" "200ms" "500ms" "${fake_metrics_args[@]}"
```

zdtsw (commented):
LGTM

scripts/dev-deploy.sh
j-mok-dev: updated. Thanks!

lioraron (approved):

PR #627, opened 2026-07-29
refactor(release): change release process

zdtsw:
## Why is this PR needed?

make release simple not rely on local code also create release branch in case we need z stream patch in the future

## What does this PR do?
- rename GHA workflow: ci-release.yaml to ci-build.yaml only build "lateset" image for new commits on main branch move build image function to create-release.yaml which now does build binary, helm chart, docker image, push
- update generate-release.sh do not rely on local SHA1, it uses "gh" remote create tag, and release branch if it is an offical release, and take SHA1 as input so user no need create tag just give a SHA1 to script.
- make RELEASE_TEMPLATE.md clear: not in use
- update docs

## How was this tested?

## Checklist

## Related Issues

cc @vishbhat

### DIFF
--- a/.github/workflows/create-release.yml
+++ b/.github/workflows/create-release.yml
@@ -1,6 +1,7 @@
-# Release workflow: creates a GitHub Release on version tags (v*.*.*),
-# builds multi-arch binaries (as .tar.gz to preserve execute bit), and publishes release notes.
-# Docker images for the same tag are built by ci-release.yaml.
+# Release workflow: triggered by a v*.*.* tag created via scripts/generate-release.sh.
+# Builds multi-arch binaries (.tar.gz), builds and pushes Docker images, packages and
+# publishes the Helm chart to GHCR, then creates the GitHub Release with generated notes.
+# Docker images are built before the Helm chart so image references are always resolvable.
 
 name: Release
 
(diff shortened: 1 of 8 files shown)


.github/workflows/create-release.yml
```diff
@@ -83,9 +84,61 @@ jobs:
...
+          REGISTRY: ghcr.io/llm-d
+          COMMIT_SHA: ${{ steps.meta.outputs.commit_sha }}
+          BUILD_DATE: ${{ steps.meta.outputs.build_date }}
+          VERSION: ${{ github.ref_name }}
+          TAG: ${{ github.ref_name }}
+
+      - name: Output image tags
+        run: |
+          echo "Images pushed:"
+          echo "  - ghcr.io/llm-d/batch-gateway-apiserver:${{ steps.meta.outputs.commit_sha }}"
```
yizhaodev: I don't think it will push image with sha here?

zdtsw: with https://github.com/zdtsw/llm-d-batch-gateway/blob/main/docker-bake.hcl#L40 it should

yizhaodev: so both ci-build.yaml and create-release.yml push images tagged with the same commit SHA
which cause image digest change after overwrite?

zdtsw: Good point. Need fix this create-release.yaml should skip tagging image with SHA1 but only tag.

yizhaodev (approved):

zdtsw:
can anyone help merge this?

PR #642, opened 2026-08-11
fix(processor): key AIMD endpoint state by route key, switch to route_key_method

### LINKED ISSUE #641: route_key_by_tenant silently disables AIMD and prefix-cache grouping
#630 updated only the inference endpoint lookup to use the tenant-scoped key, while plan preprocessing and AIMD dispatch still use the bare model ID. This mismatch means those features are computed but have no effect on actual dispatch when `route_key_by_tenant` is enabled.

Additionally, consider replacing the boolean `route_key_by_tenant` with a `route_key_method` enum for extensibility:

```
route_key_method: tenant   --> <tenant>/<model>
route_key_method: geography --> <geo>/<model>
```

See: https://github.com/llm-d/llm-d-batch-gateway/pull/630#issuecomment-5243537010

todayim:
## Summary

Fixes #641

Follow-ups to the #630 review by @yizhaodev (comments on the merged PR):

1. **Fix silently disabled AIMD under tenant-scoped routing.** `buildAIMDModels` resolved clients and keyed its per-endpoint map by the bare model ID, while dispatch items carry the route key (`<tenantID>/<modelID>`) as their ModelID. With `route_key_method: tenant` the resolver lookup missed every scoped entry, the endpoint map was built empty, and `acquireSlot` silently fell back to global-semaphore-only gating — per-endpoint concurrency limits and AIMD adaptive backoff never engaged. Resolver lookup, map keys and the dispatcher lookup now all derive from `routeKey()`.

2. **`route_key_by_tenant` → `route_key_method` enum.** As suggested in review, this leaves room for future scope methods (e.g. `geography`). Empty (default) keeps bare-model lookups; `tenant` scopes as `<tenantID>/<modelID>`. The boolean merged in #630 has not shipped in any release, so this is a clean rename before it becomes public API.

3. **Metric label alignment.** The preprocessing model-not-found error metric now uses the route key, matching the dispatch path labels.

Plan preprocessing (grouping / prefix-cache sorting) is intentionally unchanged: jobs are single-tenant units, so bare and scoped keys partition identically within a job.

## Test plan

- New `TestBuildAIMDModels` covers the regression: scoped entries register under `<tenantID>/<modelID>`, cross-tenant isolation holds, bare method + bare config keeps default behavior, and the previously-broken combination (bare keys against scoped config) is asserted to register nothing.
- `TestRouteKey` enum behavior incl. unknown methods; `TestValidate_RouteKeyMethod` rejects unknown values.
- Chart tests cover `route_key_method` rendering; `helm lint` + 100 chart unit tests pass locally.
- `go test ./internal/processor/... ./pkg/clients/...` green; `go vet` clean.

### DIFF
--- a/internal/processor/worker/route_key.go
+++ b/internal/processor/worker/route_key.go
@@ -1,13 +1,17 @@
 package worker
 
+import (
+	"github.com/llm-d/llm-d-batch-gateway/internal/processor/config"
+)
+
 // routeKey returns the model_gateways lookup key for a request model. When
-// byTenant is enabled and tenantID is non-empty, gateway entries are scoped
-// per tenant as "<tenantID>/<modelID>", letting identically-named models of
-// different tenants route to their own backends (e.g. per-InferSet gateways
-// sharing one batch-apiserver). Otherwise the bare model ID is used,
+// the route key method is "tenant" and tenantID is non-empty, gateway entries
+// are scoped per tenant as "<tenantID>/<modelID>", letting identically-named
+// models of different tenants route to their own backends (e.g. per-InferSet
+// gateways sharing one batch-apiserver). Otherwise the bare model ID is used,
 // preserving the default behavior.
-func routeKey(byTenant bool, tenantID, modelID string) string {
-	if !byTenant || tenantID == "" {
+func routeKey(method config.RouteKeyMethod, tenantID, modelID string) string {
+	if method != config.RouteKeyMethodTenant || tenantID == "" {
 		return modelID
 	}
 	return tenantID + "/" + modelID
--- a/internal/processor/worker/route_key.go
+++ b/internal/processor/worker/route_key.go
@@ -11,8 +11,8 @@ import (
 // gateways sharing one batch-apiserver). Otherwise the bare model ID is used,
 // preserving the default behavior.
 func routeKey(method config.RouteKeyMethod, tenantID, modelID string) string {
-	if method != config.RouteKeyMethodTenant || tenantID == "" {
-		return modelID
+	if method == config.RouteKeyMethodTenant && tenantID != "" {
+		return tenantID + "/" + modelID
 	}
-	return tenantID + "/" + modelID
+	return modelID
 }
--- a/charts/batch-gateway/templates/processor-configmap.yaml
+++ b/charts/batch-gateway/templates/processor-configmap.yaml
@@ -130,8 +130,8 @@ data:
     {{- if .Values.processor.config.sendFairnessHeader }}
     send_fairness_header: true
     {{- end }}
-    {{- if .Values.processor.config.routeKeyByTenant }}
-    route_key_by_tenant: true
+    {{- if .Values.processor.config.routeKeyMethod }}
+    route_key_method: {{ .Values.processor.config.routeKeyMethod | quote }}
     {{- end }}
     default_output_expiration_seconds: {{ .Values.processor.config.defaultOutputExpirationSeconds }}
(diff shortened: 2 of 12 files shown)


zdtsw (commented):
/LGTM

j-mok-dev (approved):

internal/processor/worker/route_key.go
```diff
@@ -1,13 +1,17 @@
...
+// models of different tenants route to their own backends (e.g. per-InferSet
+// gateways sharing one batch-apiserver). Otherwise the bare model ID is used,
 // preserving the default behavior.
-func routeKey(byTenant bool, tenantID, modelID string) string {
-	if !byTenant || tenantID == "" {
+func routeKey(method config.RouteKeyMethod, tenantID, modelID string) string {
+	if method != config.RouteKeyMethodTenant || tenantID == "" {
 		return modelID
 	}
 	return tenantID + "/" + modelID
```
yizhaodev: why the fall back is `tenantID + "/" + modelID`?

i am thinking something like
```
func routeKey(method config.RouteKeyMethod, tenantID, modelID string) string {
	if method == config.RouteKeyMethodTenant && tenantID != "" {
		return tenantID + "/" + modelID
	}
    // any other new method in the future
    // fallback
	return modelID
```

todayim: The guarded path is actually the opposite: scoping only applies when `method == RouteKeyMethodTenant` **and** the tenant ID is non-empty; the fallback in every other case is the bare model ID.

Keying on `tenantID != ""` alone would flip semantics for deployments where a tenant ID is always present (the apiserver persists it per job): the feature would silently turn on with the method unset, breaking existing bare-keyed model_gateways — the opt-in contract from #630. An unknown/future method (e.g. `geography`) also must not degrade into tenant scoping; falling through to bare plus the config-time rejection of unknown values keeps that safe. TestRouteKey covers the unknown-method → bare case as a guard.

yizhaodev: sorry, should be

```
func routeKey(method config.RouteKeyMethod, tenantID, modelID string) string {
	if method == config.RouteKeyMethodTenant && tenantID != "" {
		return tenantID + "/" + modelID
	}
    // any other new method in the future
    // fallback
	return modelID

```

todayim: Exactly — that's semantically identical to the current negated guard (method != tenant || tenantID == "" → bare). The current shape keeps the fallback first as the early return, but both are equivalent, so leaving as-is. Thanks for the review!

yizhaodev: feel like talking to ai :)
can you change to modelD as fallback at last?

todayim: Done in d5ec4b3 — positive guard, bare model ID as the final fallback. Behavior unchanged (TestRouteKey still covers all four cases).

zdtsw:
thanks @todayim for the fix, just curious what is your use case of using batch-gateway? mind if sharing

todayim:
Happy to! We run an internal model-as-a-service platform where multiple projects share one batch-apiserver. Each project deploys its models behind its own gateway (istio ingress in its own namespace), and different projects often serve models under the same name — which is exactly the routing collision #630 fixes for us.

On top of that we carry tenant identity end to end: the entry point injects a tenant header, the apiserver persists it on the job, and the processor re-injects it on dispatch, so we can do per-project metering and budget enforcement downstream (we're wiring usage into LiteLLM spend tracking off the output files).

Thanks for maintaining this project — the processor's plan/prefix-cache design and the orphan recovery story made it a very good fit, and we're glad to keep contributing back.

PR #647, opened 2026-08-19
feat(gc): initialize OTel tracer and trace orphan triage

wseaton:
batch-gc is the only binary without tracing, so the reconciler's interventions (re-enqueues, CAS transitions on orphans) are invisible in traces even though they change batch state. This initializes the OTel tracer in main, wraps `triageOrphan` in a `reconciler.triage` span carrying the batch ID and status, and wires the existing `otel.redis_tracing` / `otel.postgresql_tracing` config flags through to the gc's store clients, matching the apiserver and processor.

No behavior change when `OTEL_EXPORTER_OTLP_ENDPOINT` is unset (tracing stays disabled, same as the other binaries).

### DIFF
--- a/cmd/batch-gc/main.go
+++ b/cmd/batch-gc/main.go
@@ -43,6 +43,7 @@ import (
 	"github.com/llm-d/llm-d-batch-gateway/internal/util/clientset"
 	ucom "github.com/llm-d/llm-d-batch-gateway/internal/util/com"
 	"github.com/llm-d/llm-d-batch-gateway/internal/util/interrupt"
+	uotel "github.com/llm-d/llm-d-batch-gateway/internal/util/otel"
 )
 
 func main() {
@@ -70,6 +71,18 @@ func run() error {
 
 	logger.Info("Starting batch garbage collector", "dryRun", cfg.DryRun, "interval", cfg.Collector.Interval)
 
+	shutdownTracer, err := uotel.InitTracer(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to initialize tracer: %w", err)
+	}
+	defer func() {
+		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer shutdownCancel()
+		if err := shutdownTracer(shutdownCtx); err != nil {
+			logger.Error(err, "Failed to shutdown tracer")
+		}
+	}()
+
 	if err := gcmetrics.InitMetrics(); err != nil {
 		return fmt.Errorf("failed to initialize metrics: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


cmd/batch-gc/main.go
```diff
@@ -70,6 +71,18 @@ func run() error {
 
 	logger.Info("Starting batch garbage collector", "dryRun", cfg.DryRun, "interval", cfg.Collector.Interval)
 
+	shutdownTracer, err := uotel.InitTracer(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to initialize tracer: %w", err)
+	}
+	defer func() {
+		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 5*time.Second)
```
hexfusion, line 79: i would like to eventually see more general handling of all things graceful during termination. like how does this timing differ from catching a SIGTERM signal ordering etc. just food for thought

internal/gc/reconciler/reconciler.go
```diff
@@ -221,12 +223,17 @@ func (r *Reconciler) fetchNonTerminalJobs(ctx context.Context) ([]*db.BatchItem,
 func (r *Reconciler) triageOrphan(ctx context.Context, job *db.BatchItem, result *Result) {
 	logger := logr.FromContextOrDiscard(ctx).WithValues("jobId", job.ID)
 
+	ctx, span := uotel.StartSpan(ctx, "reconciler.triage")
+	defer span.End()
```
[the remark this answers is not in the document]
wseaton: copilot is right, this can be improved for log correlation

cmd/batch-gc/main.go
wseaton: agree, this pattern should be firmed up and made consistent throughout the codebase, opened https://github.com/llm-d/llm-d-batch-gateway/issues/650 for tracking

internal/gc/reconciler/reconciler.go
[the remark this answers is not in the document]
wseaton: fixed

Gregory-Pereira (approved):
/lgtm

internal/gc/reconciler/reconciler.go
```diff
@@ -219,14 +221,19 @@ func (r *Reconciler) fetchNonTerminalJobs(ctx context.Context) ([]*db.BatchItem,
...
+	span.SetAttributes(attribute.String(uotel.AttrBatchID, job.ID))
+	logger := logr.FromContextOrDiscard(ctx)
 
 	var statusInfo openai.BatchStatusInfo
 	if err := json.Unmarshal(job.Status, &statusInfo); err != nil {
 		logger.Error(err, "Reconciler: failed to unmarshal job status")
 		result.Errors++
 		return
 	}
+	span.SetAttributes(attribute.String("batch.status", string(statusInfo.Status)))
```
zdtsw, line 236: nit: 
add `batch.status` as a const in otel.go ?

Issue #648, opened 2026-08-19
Single-owner schema migrations via a migrate subcommand


wseaton:
Every component applies schema on boot: `pgCore` runs `CREATE TABLE IF NOT EXISTS` unconditionally during client construction (`internal/database/postgresql/db_core.go`), so apiserver, processor, and gc are all DDL writers. On a fresh database the three concurrent applies race inside Postgres and crash with spurious SQLSTATE 23505 catalog conflicts; a retry (#649) papers over that symptom, but the shape is the problem.

It also caps what migrations can be: three concurrent writers only work while the DDL is idempotent `CREATE IF NOT EXISTS`. Ordered, versioned, non-idempotent migrations (column normalization, new tables, backfills) need a single owner.

Proposal: a `migrate` subcommand on one of the existing binaries, run before the components start (helm pre-install/pre-upgrade hook or initContainer):

- `migrate` applies versioned migrations and records the schema version
- components stop running DDL; on boot they check the schema version and fail fast with a clear error if it's missing or behind
- local dev keeps a convenience path (`migrate` invoked by the dev-deploy/compose flow)

wseaton:
Consolidated into #652.

PR #649, opened 2026-08-19
fix(postgresql): retry ensureSchema on SQLSTATE 23505

wseaton:
Concurrent `CREATE TABLE IF NOT EXISTS` from multiple booting components can race inside Postgres and fail with spurious SQLSTATE 23505 unique violations on the `pg_type`/`pg_class` catalog indexes even though the `IF NOT EXISTS` guard holds. Seen as a real cold-boot crash when apiserver, processor, and gc start together against a fresh database. The loser now retries and finds the objects in place.

This treats the symptom at minimal size. The underlying shape — schema DDL running in every component — is #648; single-owner migrations make this retry unnecessary.

### DIFF
--- a/internal/database/postgresql/db_core.go
+++ b/internal/database/postgresql/db_core.go
@@ -23,12 +23,15 @@ package postgresql
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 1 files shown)


wseaton:
Superseded by #651 (same branch).

PR #651, opened 2026-08-19
fix(postgresql): retry ensureSchema on SQLSTATE 23505

wseaton:
Found this working on a batch gateway simulation harness.

When the apiserver, processor, and gc all boot against a fresh database, their concurrent `CREATE TABLE IF NOT EXISTS` calls can race inside postgres and one of them dies with a spurious SQLSTATE 23505 on the `pg_type`/`pg_class` catalog indexes, even though the `IF NOT EXISTS` guard holds. Rare in the wild because real deploys have startup jitter and k8s restarts hide it, but it's a guaranteed crash-loop lottery on any fresh install. The fix is just a bounded retry — the loser tries again and finds the tables in place.

A more durable fix would be to separate all database DDL into a utility that is run via a helm postInstall hook, instead of having 3 separate binaries try to do the same DDL operations — that's #652.

### DIFF
--- a/internal/database/postgresql/db_core.go
+++ b/internal/database/postgresql/db_core.go
@@ -23,12 +23,15 @@ package postgresql
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgconn"
 	"github.com/llm-d/llm-d-batch-gateway/internal/database/api"
 	"github.com/llm-d/llm-d-batch-gateway/internal/util/logging"
 )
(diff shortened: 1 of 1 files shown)


internal/database/postgresql/db_core.go
```diff
@@ -454,8 +457,26 @@ func (c *pgCore) delete(ctx context.Context, ids []string) (deletedIDs []string,
...
+		// in some environments, especially on fresh boot, different
+		// processes can contend with each other and produce a
+		// duplicate constraint error (23505)
+		if !errors.As(err, &pgErr) || pgErr.Code != "23505" {
+			return err
+		}
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-time.After(time.Duration(attempt+1) * 100 * time.Millisecond):
```
zdtsw, line 477: so in theory, all of them can retry after 100ms and fail again, then all retry again after another 200ms and might fail again :D 

should the After has a random seed so not all three retry at the same time?

wseaton: we can add some additional jitter sure, but the real fix here would be to properly externalize migrations and stop doing DDL in multiple places

lioraron, line 477: https://pkg.go.dev/github.com/cenkalti/backoff/v5 is used in this repo for retrying operations.
It can be upgraded.

acardace (approved):

lioraron (changes requested):
See the comments.

wseaton:
closing in favor of https://github.com/llm-d/llm-d-batch-gateway/pull/677

PR #655, opened 2026-08-19
feat: resume batches from Postgres result scratch after worker loss

wseaton:
Example fix for F4a driven by the consistency ratchet in #654. Worker scratch state moves to Postgres: each request's output line is written through to a batch_request_results table keyed by (batch_id, custom_id) as it completes, the reconciler re-enqueues non-expired in_progress orphans instead of failing them, and the replacement worker replays persisted rows and executes only the remainder. Rows are deleted after the terminal status write; a processor/before-result-persist failpoint covers the response-to-commit window.

Ratchet results: F4a flips broken -> fixed (worker loss now ends completed, not stranded). New scenario F4c kills the worker with 2 of 4 results persisted and requires exactly-once delivery; the engine serves 6 requests, not 8, proving the resume skips persisted work. A1 still reproduces, its symptom morphing from destroyed work to duplicate spend, and the judge now pins that. All other scenarios reproduce unchanged.

Known gaps, deliberate for a prototype: rows orphaned by non-finalize terminalization have no sweep yet, redis backends run with persistence disabled (full re-run on resume), and F4b's double-dispatch remains.

### DIFF
--- a/internal/database/api/result_item.go
+++ b/internal/database/api/result_item.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2026 The llm-d Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


wseaton:
Temporarily moving to my personal fork; upstream PR opened by mistake. This is a proof of concept of validating a correctness fix via the ratchet.

PR #661, opened 2026-08-25
test(e2e): run inference on vllm-vcr, shed batch traffic via the EPP

wseaton:
Replaces llm-d-inference-sim with [vllm-vcr](https://github.com/neuralmagic/vllm-vcr) in the e2e setup and removes the three fake-429 simulators. This removes all code that needs to interact with the "fake metrics" feature of the old sim, since the new simulator uses the real vllm frontend the metrics do not need to be faked.

More importantly, the shed/retry tests now go through the EPP for real: we can choke a model's engine via vcr's control API, run some interactive traffic against the same EPP, and let batch get shed. The EPP returns 503 on queue TTL expiry (429 only when a band's byte budget fills), so the tests and the flow-control guide now say that. The output file is checked against the EPP's batch-band outcome counters.

Other fix: The pass-through header test reads the header names from the `process-batch` span in Jaeger instead of grepping processor logs.

AIMD gauges are logged rather than asserted; we will get AIMD fidelity tests back up in a follow-up PR.

Notes:
- new image dep `ghcr.io/neuralmagic/vllm-vcr:0.2.2-vllm0.27` (amd64/arm64); pod downloads a small tokenizer from Hugging Face on start
- dev-deploy sets EPP `queueDepthThreshold` to 1 (test tuning)
- GIE leg is ~5 min slower; shed/AIMD tests skip on non-GIE legs
- 5 sim deployments → 2

Tested on kind: non-GIE suite 66/66, GIE suite green.


acardace (approved):

test/e2e/dispatcher_test.go
```diff
@@ -481,25 +483,87 @@ func testDispatcherEndpointScrapeGate(t *testing.T, rdb *redis.Client) {
...
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusNoContent {
-		t.Fatalf("fake_metrics returned %d", resp.StatusCode)
+
+	const parkedRequests = 10
+	patchEngineConfig(t, testSimService, `{"max_num_seqs": 1, "time_to_first_token": 0, "inter_token_latency": 30000}`)
+
+	ensureE2ECurlPod(t)
+	url := fmt.Sprintf("http://%s.%s.svc.cluster.local:8000/v1/completions", testSimService, testNamespace)
+	body := fmt.Sprintf(`{"model":%q,"prompt":"park","max_tokens":5}`, testModel)
+	script := fmt.Sprintf(`for i in $(seq 1 %d); do curl -sS -o /dev/null -m 600 -X POST %s -H 'content-type: application/json' -d '%s' & done; wait`, parkedRequests, url, body)
```
lioraron: `url` is interpolated directly into the shell script without quoting. If `testSimService` or `testNamespace` ever contained a space or shell metacharacter the command would silently misbehave. Consider single-quoting the URL: `... -X POST '%s' ...`. The `body` is already correctly single-quoted.

test/e2e/flow_control_test.go
```diff
@@ -124,180 +129,262 @@ func doTestSLOHeader(t *testing.T) {
...
+// creates, so the EPP queues rather than sheds them and dispatches them ahead
+// of batch traffic.
+func startInteractiveLoad(t *testing.T, model string, concurrency int) func() {
+	t.Helper()
+
+	ensureE2ECurlPod(t)
+
+	url := fmt.Sprintf("http://%s.%s.svc.cluster.local:8081/v1/chat/completions", eppDeploymentFor(model), testNamespace)
+	body := fmt.Sprintf(`{"model":%q,"max_tokens":%d,"messages":[{"role":"user","content":"interactive"}]}`, model, saturationMaxTokens)
+	script := fmt.Sprintf(`for i in $(seq 1 %d); do (while true; do curl -sS -o /dev/null -m 300 -X POST %s -H 'content-type: application/json' -H 'x-gateway-inference-objective: interactive-default-%s' -d '%s'; done) & done; wait`,
```
lioraron: Both `url` and `model` are interpolated into the shell script without quoting. `url` should be single-quoted (`'%s'`) and the inference-objective header value (`-H 'x-gateway-inference-objective: interactive-default-%s'`) relies on `model` containing no single-quote characters. All current callers are safe (k8s names), but a quote around the objective value (e.g. using a variable or separate string) would make the helper resilient to future model names.

test/e2e/dispatcher_test.go
wseaton: kubernetes service and namespace names have to be RFC 1123 compliant because of cluster dns, so they by definition can't contain those characters, nonetheless we can quote

test/e2e/flow_control_test.go
wseaton: addressed in [e3567bf](https://github.com/llm-d/llm-d-batch-gateway/pull/661/commits/e3567bf039c9dc0a123897a056f72751a8a816d2)

lioraron (approved):

Issue #645, opened 2026-08-18
[Feature]: Resume in-progress batches after Batch Processor pod or node loss


hhzhang16:
### Problem Statement

The current startup recovery works when the container restarts in the same pod and its `emptyDir` survives. If the pod is replaced or the node is lost, the new Processor no longer has the local plan, output, or progress files it needs to recover the batch.

I reproduced this by force-deleting the only Processor pod during a 64-request Async batch. The replacement pod came up, both Async queues drained, and the batch stayed `in_progress` at 27/64 with no output or error file. A later cancellation also stayed in `cancelling` because no live Processor owned the job.

This is particularly painful for long or expensive batches: some inference work may have completed, but the user cannot retrieve the partial output or resume the rest of the job.

The current behavior is documented in startup recovery:

https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/internal/processor/worker/recovery.go#L63-L109

The orphan reconciler helps detect this when GC is enabled, but an orphaned `in_progress` job is moved to `failed`; it is not resumed:

https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/internal/gc/reconciler/reconciler.go#L219-L265

### Proposed Solution

Persist enough execution state outside the Processor pod for another replica to take over the job. I think that includes:

- completed request IDs and their output/error records;
- a durable cursor or equivalent state for requests that have not been dispatched;
- job ownership/epoch information so a stale Processor cannot also finalize the job;
- idempotent output writes for requests that are redelivered after a failure;
- enough cancellation state for a replacement Processor to finish cancelling the job.

For Async mode, the result should only be acknowledged after this checkpoint succeeds. Requests with checkpointed terminal results should not run again; a request that was only in flight may need at-least-once replay, with the final output deduplicated by its internal request ID.

Pod-deletion E2E coverage should exercise this in both sync and Async dispatch modes.

### Alternatives Considered

- **Current orphan reconciler:** prevents an orphan from staying invisible forever when GC is enabled, but terminally fails an in-progress job and cannot restore partial output that was only in `emptyDir`.
- **Persistent workdir:** would preserve the current local files, but introduces shared filesystem/ownership concerns and still needs an explicit takeover and idempotency contract.
[...]

PR #662, opened 2026-08-25
fix(processor): isolate async result queues by replica

### LINKED ISSUE #644: [Bug]: Async results can be lost with multiple Batch Processor replicas
### Contact Details

hannahz@nvidia.com

### What happened?

I tested Async dispatch with two healthy Batch Processor replicas and one 64-request batch. The batch stopped at 33/64 even though both Redis queues were empty. A one-replica control on the same deployment completed 32/32. Cancelling the stuck batch produced 33 successful records and 31 cancellation/error records.

It looks like both replicas consume from the same result list, while only the replica that owns the job has its pending request IDs. If the other replica gets a result, its local pending map rejects it as belonging to another job, but the result has already been removed from Redis by `BRPOP`.

The relevant paths are:

- The resolver intentionally shares the result client: https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/pkg/clients/inference/async_inference_client_resolver.go#L74-L93
- Each process broadcasts only to its local subscribers: https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/internal/processor/pipeline/result_router.go#L40-L114
- A result that is not in the local pending map is rejected: https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/internal/processor/pipeline/pending.go#L48-L73
- The Producer API consumes results with `BRPOP`: https://github.com/llm-d/llm-d-async/blob/1f8c87672edbd2afec8ba50d816c80ef7390db6b/producer/redis_sortedset_producer.go#L260-L278

I think we need either a reply queue tied to the Processor/job owner, or a durable shared result stream where a result is only acknowledged after a Processor that can persist it accepts it. Whichever direction we take also needs an owner failover story; otherwise a unique queue fixes result stealing but still strands results when that owner dies.

A multi-replica E2E test should cover concurrent jobs and verify that every input produces exactly one terminal record.

Related request-side durability issue: llm-d/llm-d-async#404 tracks accepted requests being lost after a hard Async failure.

### Version

`e5d70e8978dcb288fd6729c73dccdc48e8ff5a99` (`main` on 2026-08-18)

### Steps to Reproduce

1. Configure Batch Gateway to use llm-d Async.
2. Run two Batch Processor replicas against the same request/result queues.
3. Submit one 64-request batch.
4. Wait for both Redis queues to drain.
5. Observe that the owning Processor is still waiting for results consumed by the other replica.

### Environment

```markdown
- Kubernetes
- Batch Processor replicas: 2
- llm-d Async: 1f8c87672edbd2afec8ba50d816c80ef7390db6b
- Shared Redis request/result queues
- Inference target: OpenAI-compatible Dynamo frontend/router from https://github.com/ai-dynamo/dynamo/pull/12784
```

### Relevant log output

```shell
request queue: 0
result queue: 0
batch progress: 33/64
missing results: 31
batch status: in_progress
```

hhzhang16:
## Why is this PR needed?

Each Batch Processor currently calls `BRPOP` on the same result list. A result can therefore be consumed by a replica that does not own the job. That replica rejects the unknown request ID after Redis has already removed the result, leaving the owning job stuck.

llm-d Async already carries the producer's `ResultQueueName` on each request and routes the result back to it. This change uses that existing contract: Processors still submit to the same request queue, but each Processor listens on `<result_queue_base>:<processor_id>`.

This fixes result stealing between healthy replicas. Recovery after the owning Processor is lost remains separate durability work in #645. This change does not depend on #582; if that ownership design is adopted later, it must preserve the per-Processor result-routing contract or replace it with equivalent durable routing.

When llm-d Async is configured with `queuesConfig`, its per-queue `result_queue_name` must be omitted because current Async releases give that static value precedence over the per-request result destination. A top-level result queue may remain configured as a fallback. For upgrades, remove the per-queue override and wait for the Async rollout before upgrading Batch Processors; reversing that order strands results even with one Processor replica. The repository's E2E, benchmark, demo, and deployment-guide configurations are updated accordingly.

## What does this PR do?

- Keep each pool's Async request queue shared across Batch Processor replicas.
- Derive a replica-specific result queue from the configured result queue base and the Processor ID.
- Add a Redis contract test proving that two Processors share requests while receiving only their own results.
- Add a two-replica E2E that submits two 32-request jobs and verifies that every input produces exactly one successful output.
- Update the Async deployment guidance to preserve the per-request result destination.

## How was this tested?

- `go test ./...`
- `go test -tags=integration ./...`
- `go vet ./...`
- `golangci-lint run ./...`
- `go test -race ./pkg/clients/inference`
- Replica-isolation contract test repeated 100 times
- `goimports` clean; `go mod tidy -diff` clean
- `gosec` introduced no findings relative to current `main`
- Linux/amd64 Batch Processor build
- E2E module compile and vet
- Helm lint for default and all CI values, template render, and 100 Helm unit tests
- Patch applies cleanly to current `main`
- Live Kubernetes E2E against a real Batch Gateway API, shared storage and Redis, llm-d Async (`1f8c876`), and NVIDIA Dynamo frontend/router (`Qwen/Qwen3-0.6B`): label-based discovery resolved the custom workload name `batch-repro-644-batch-gateway-processor`, and two patched replicas on separate nodes completed two concurrent 32-request jobs with 64/64 unique successful outputs, zero failed requests, and zero duplicate or missing results

## Checklist

## Related Issues

Closes #644

### DIFF
--- a/pkg/clients/inference/async_inference_client_resolver.go
+++ b/pkg/clients/inference/async_inference_client_resolver.go
@@ -41,6 +41,7 @@ type AsyncModelPoolConfig struct {
 // AsyncClientConfig holds the resolved configuration for async dispatch.
 type AsyncClientConfig struct {
 	RedisURL          string
+	ConsumerID        string // identity of this Batch Processor replica
 	Models            map[string]AsyncModelPoolConfig
 	ResultPollTimeout time.Duration // per-poll timeout in the result dispatcher loop
 }
@@ -72,8 +73,8 @@ func (r *AsyncGatewayResolver) Models() []string {
 }
 
 // SharedClientFor returns a shared client for the given model.
-// Reuses the same client across calls — results are not routed
-// per-request, any consumer can read them.
+// Reuses the same client across calls. Results for this Processor replica are
+// isolated on its consumer-specific result queue.
 func (r *AsyncGatewayResolver) SharedClientFor(modelID string) AsyncInferenceClient {
 	if r.clientFactories != nil {
 		if factory, ok := r.clientFactories[modelID]; ok {
@@ -113,6 +114,10 @@ func (r *AsyncGatewayResolver) Close() error {
 // NewAsyncResolver creates an AsyncGatewayResolver with one shared pool
 // (producer) per model/pool pair.
 func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatewayResolver, error) {
+	if config.ConsumerID == "" {
+		return nil, fmt.Errorf("consumerID must not be empty")
+	}
+
 	opts, err := redis.ParseURL(config.RedisURL)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse async inference Redis URL: %w", err)
@@ -147,6 +152,9 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		if resQueue == "" {
 			resQueue = asyncQueuePrefix + "results:" + mcfg.PoolName
 		}
+		// Each Processor consumes from its own result queue so another healthy
+		// replica cannot consume and discard its results.
+		resQueue = resQueue + ":" + config.ConsumerID
 
 		p, err := producer.NewRedisSortedSetProducer(
 			producer.RedisSortedSetConfig{
--- a/pkg/clients/inference/async_inference_client_resolver.go
+++ b/pkg/clients/inference/async_inference_client_resolver.go
@@ -155,6 +155,12 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		// Each Processor consumes from its own result queue so another healthy
 		// replica cannot consume and discard its results.
 		resQueue = resQueue + ":" + config.ConsumerID
+		logger.Info(
+			"Configured async inference queues",
+			"model", model,
+			"requestQueue", reqQueue,
+			"resultQueue", resQueue,
+		)
 
 		p, err := producer.NewRedisSortedSetProducer(
 			producer.RedisSortedSetConfig{
--- a/test/e2e/dispatcher_test.go
+++ b/test/e2e/dispatcher_test.go
@@ -422,8 +422,13 @@ func testDispatcherMultiReplicaBatch(t *testing.T) {
 		batchIDs = append(batchIDs, mustCreateBatch(t, fileID))
 	}
 
+	completionDeadline := time.Now().Add(3 * time.Minute)
 	for _, batchID := range batchIDs {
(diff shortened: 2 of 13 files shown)


pkg/clients/inference/async_inference_client_resolver.go
```diff
@@ -147,6 +152,9 @@ func NewAsyncResolver(config AsyncClientConfig, logger logr.Logger) (*AsyncGatew
 		if resQueue == "" {
 			resQueue = asyncQueuePrefix + "results:" + mcfg.PoolName
 		}
+		// Each Processor consumes from its own result queue so another healthy
+		// replica cannot consume and discard its results.
+		resQueue = resQueue + ":" + config.ConsumerID
```
lioraron, line 157: When `mcfg.ResultQueueName` is explicitly set, its semantics silently changed from "exact queue name" to "queue base name" — the effective queue the Processor listens on is now `{configured-value}:{pod-id}`, not `{configured-value}`. An operator who upgrades the Processor before removing the per-queue `result_queue_name` from Async will see stranded results with no log signal pointing at the mismatch. A single `logger.Info` line here (e.g. "async result queue for model X: {resQueue}") would make the effective name visible at startup and cut the diagnostic time on misconfigured upgrades.

test/e2e/dispatcher_test.go
```diff
@@ -339,6 +344,106 @@ func testDispatcherMultiRequestBatch(t *testing.T, rdb *redis.Client) {
...
+		}
+
+		fileID := mustCreateFile(t,
+			fmt.Sprintf("dispatcher-multi-replica-%d-%s.jsonl", batchIndex, testRunID),
+			strings.Join(lines, "\n"))
+		batchIDs = append(batchIDs, mustCreateBatch(t, fileID))
+	}
+
+	for _, batchID := range batchIDs {
+		batch, results := waitForBatchStatus(t, batchID, 2*time.Minute, openai.BatchStatusCompleted)
```
lioraron: The two batches are submitted sequentially and then also waited on sequentially. If batch 0 takes close to 2 minutes, the `waitForBatchStatus` for batch 1 starts with little time remaining even though batch 1 was submitted nearly simultaneously. With 2 replicas both batches should complete at roughly the same wall time, but there is no guarantee — in a loaded CI environment this could be flaky. Consider waiting for both concurrently (e.g. a single timeout from first submission) or increasing the per-batch timeout.

hhzhang16: Addressed

pkg/clients/inference/async_inference_client_resolver.go
hhzhang16: Addressed

lioraron (approved):
Thanks!

PR #664, opened 2026-08-31
fix(otel): support OTEL_TRACES_EXPORTER=none|otlp|console in InitTracer

gyliu513:
## Why is this PR needed?

InitTracer only supported an implicit on/off switch based on whether OTEL_EXPORTER_OTLP_ENDPOINT was set, with otlp as the only exporter. There was no way to select a console exporter for local development, and worse, when tracing was "off" (no endpoint configured) the function returned before installing the W3C trace-context propagator at all, so a request passing through batch-gateway with tracing disabled would drop its incoming trace context entirely instead of forwarding it downstream. 

## What does this PR do?

The propagator is now installed unconditionally at the top of InitTracer, before the exporter type is even resolved, so trace context received by apiserver, batch-processor, and batch-gc is always forwarded regardless of export configuration.

traceExporterType resolves the new OTEL_TRACES_EXPORTER variable against otlp/console/none, defaulting to otlp when unset (matching llm-d-router, llm-d-kv-cache, and llm-d-inference-payload-processor) and falling back to otlp with a logged warning on an unrecognised value instead of failing startup. 

Note the default changed: previously, tracing was off unless OTEL_EXPORTER_OTLP_ENDPOINT was explicitly set. Now, all three binaries default to attempting an otlp export to the SDK's default endpoint (localhost:4317) unless OTEL_TRACES_EXPORTER=none is set. This matches the other three llm-d repos' current default.

## How was this tested?

- [X] Unit tests added/updated/verified

## Checklist

- [ ] Commits are signed off (`git commit -s`) per [DCO](PR_SIGNOFF.md)

### DIFF
--- a/internal/util/otel/otel.go
+++ b/internal/util/otel/otel.go
@@ -23,6 +23,7 @@ import (
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc"
+	"go.opentelemetry.io/otel/exporters/stdout/stdouttrace"
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/sdk/resource"
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
@@ -33,6 +34,15 @@ import (
 )
 
 const defaultServiceName = "batch-gateway"
+
+// The exporter types OTEL_TRACES_EXPORTER selects between.
+const (
+	exporterTypeOTLP    = "otlp"
+	exporterTypeConsole = "console"
+	exporterTypeNone    = "none"
+
+	defaultExporterType = exporterTypeOTLP
+)
 
 // Span attribute keys for batch-gateway resources.
 const (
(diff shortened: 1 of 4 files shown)


internal/util/otel/otel.go
```diff
@@ -106,44 +107,98 @@ func DetachedContext(ctx context.Context, name string) (context.Context, trace.S
...
 }
+
+// The exporter types OTEL_TRACES_EXPORTER selects between.
+const (
+	exporterTypeOTLP    = "otlp"
+	exporterTypeConsole = "console"
+	exporterTypeNone    = "none"
+
+	defaultExporterType = exporterTypeOTLP
+)
```
zdtsw: nit: 
move const up , close to others

zdtsw (commented):
LGTM

In general, i would say this is a bug fix type change. exporter default flipped is a behavior change for existing deployments. worth calling it out explicitly so it's surfaced properly in the next release notes

internal/util/otel/otel.go
gyliu513: Done, thanks @zdtsw

zdtsw (commented):
/LGTM

zdtsw (approved):

Issue #656, opened 2026-08-20
[Feature]: Support configurable batch endpoints for compatible inference backends


hhzhang16:
### Problem Statement

Batch Gateway currently validates the batch endpoint against a fixed list of OpenAI Batch API endpoints. That is a good default, but it means an OpenAI-compatible backend cannot use Batch Gateway for additional endpoints it supports.

This is a concrete integration requirement for NVIDIA Dynamo. Dynamo supports both endpoints for vLLM pooling/classification deployments: its HTTP frontend serves `/v1/classify` and `/v1/pooling`, and a vLLM worker started with `--classify-worker` registers both endpoints for an engine using `--runner pooling`. Requests for `/v1/classify` and `/v1/pooling` are rejected by Batch Gateway with `400 invalid endpoint`, before they could reach the inference backend.

- HTTP frontend support: [ai-dynamo/dynamo#12139](https://github.com/ai-dynamo/dynamo/pull/12139)
- Classify/pooling model and protocol support: [ai-dynamo/dynamo#12140](https://github.com/ai-dynamo/dynamo/pull/12140)
- vLLM classify/pooling worker support: [ai-dynamo/dynamo#12142](https://github.com/ai-dynamo/dynamo/pull/12142)

The current allowlist is here:

https://github.com/llm-d/llm-d-batch-gateway/blob/e5d70e8978dcb288fd6729c73dccdc48e8ff5a99/internal/shared/openai/batch.go#L28-L57

### Proposed Solution

Keep the current OpenAI endpoint list as the default, but allow an operator to add explicit deployment-specific endpoints through configuration.

I don't think Batch Gateway should accept arbitrary paths by default because that widens the proxy/API surface. An extension allowlist would keep the current OpenAI-compatible behavior safe and backward-compatible while allowing endpoints such as `/v1/classify` and `/v1/pooling` when the backend supports them.

The request URL should still match the endpoint declared on the batch, and the endpoint should be forwarded unchanged in both sync and Async modes.

For validation, the two additional endpoints should be tested end to end against a Dynamo deployment using a compatible pooling/classification model. The earlier reproduction used a generation-only deployment, so it could confirm the Batch Gateway rejection but not downstream classify/pooling behavior.

### Alternatives Considered

- Keep the fixed OpenAI-only list. This is the strictest interpretation of the OpenAI Batch API, but prevents compatible backends from extending the batch surface.
[...]

PR #665, opened 2026-09-01
feat: support configurable batch endpoints

elizabetht:
## Why is this PR needed?

Batch Gateway currently validates batch endpoints against a fixed list of OpenAI Batch API endpoints. OpenAI-compatible inference backends may expose other batch-capable endpoints; for example, NVIDIA Dynamo supports `/v1/classify` and `/v1/pooling`, but Batch Gateway rejects batches using those endpoints before requests reach the backend.

This PR preserves the OpenAI-only default while allowing operators to explicitly enable deployment-specific endpoints.

## What does this PR do?

- Adds a reusable endpoint allowlist containing the standard OpenAI Batch API endpoints plus explicitly configured extensions.
- Validates configured paths at startup and rejects absolute URLs, query strings, fragments, escaped paths, and other unsafe values.
- Adds standalone API-server configuration under `batch_api.extra_endpoints` and processor configuration under `extra_endpoints`.
- Adds shared Helm configuration under `global.extraEndpoints`, with component-specific API-server and processor overrides.
- Documents a processor-first rollout: configure processor endpoints, wait for the processor rollout, enable them globally, then remove the processor override so it inherits the global value.
- Applies the same allowlist during batch creation and processor JSONL preprocessing.
- Requires every request URL in an input JSONL file to exactly match the endpoint declared on its batch.
- Preserves configured endpoints unchanged through synchronous and asynchronous inference dispatch.
- Preserves existing OpenAI-only behavior when no additional endpoints are configured.

Example Helm configuration:

```yaml
global:
  extraEndpoints:
    - /v1/classify
    - /v1/pooling
```

## How was this tested?

  - Covered default, configured, unconfigured, malformed, and unsafe endpoints.
  - Covered API-server and processor configuration, batch creation, JSONL validation, and exact request-to-batch endpoint matching.
  - Added an inference-client regression proving `/v1/classify` and `/v1/pooling` reach `InferenceHTTPClient.Generate` unchanged.
  - Full unit suite: `1173 passed, 0 failed, 17 skipped`.
  - Local kind E2E suite: `66 passed, 0 failed, 4 skipped`.
  - Added Helm rendering tests for global inheritance, component overrides, and empty endpoint lists.
  - Ran `helm lint` and verified the rendered API-server and processor ConfigMaps.
  - Validated a processor-first rollout against a real Dynamo deployment using `cross-encoder/nli-MiniLM2-L6-H768`.
  - Before configuration, `/v1/classify` and `/v1/pooling` batch creation returned HTTP 400.
  - After configuration, batches for both endpoints completed with `total=1, completed=1, failed=0`; output records returned HTTP 200 with no error, and Dynamo logs confirmed both paths arrived unchanged.
  - Restored the Helm release and GPU deployment and removed the temporary namespace and SSH tunnel after validation.

Detailed Dynamo commands, topology, results, and cleanup evidence are recorded in [this PR comment](https://github.com/llm-d/llm-d-batch-gateway/pull/665#issuecomment-5555606756).

## Checklist

## Related Issues

#656

### DIFF
--- a/charts/batch-gateway/values.yaml
+++ b/charts/batch-gateway/values.yaml
@@ -2,6 +2,10 @@
 global:
   imagePullSecrets: []
 
+  # Additional inference API paths accepted by both the API server and processor.
+  # Standard OpenAI Batch API endpoints are always allowed.
+  extraEndpoints: []
+
   # Shared pod security context for all components.
   # All three components (apiserver, processor, gc) must run with the same UID/GID
   # when sharing files on a PVC. Per-component overrides are possible via
--- a/charts/batch-gateway/values.yaml
+++ b/charts/batch-gateway/values.yaml
@@ -3,7 +3,10 @@ global:
   imagePullSecrets: []
 
   # Additional inference API paths accepted by both the API server and processor.
-  # Standard OpenAI Batch API endpoints are always allowed.
+  # Standard OpenAI Batch API endpoints are always allowed. Component-specific
+  # values override this default and can be used for a processor-first rollout:
+  # first set processor.config.extraEndpoints, wait for the processor rollout,
+  # then set this value, and finally reset the processor override to null.
   extraEndpoints: []
 
   # Shared pod security context for all components.
@@ -203,6 +206,8 @@ apiserver:
       eventTTLSeconds: 2592000
       passThroughHeaders:
         - Authorization
+      # null inherits global.extraEndpoints. A list, including [], overrides it.
+      extraEndpoints: null
     fileAPI:
       defaultExpirationSeconds: 7776000
       maxSizeBytes: 209715200
@@ -314,6 +319,10 @@ processor:
     taskWaitTime: "1s"
     pollInterval: "5s"
     numWorkers: 20
+    # null inherits global.extraEndpoints. Set this first for a processor-first
+    # rollout, then reset it to null after global.extraEndpoints is enabled.
+    # An explicit [] disables extensions for this component.
+    extraEndpoints: null
     concurrency:
       # Fixed ceiling for total in-flight requests across all endpoints.
       # When AIMD is enabled, per-endpoint limits self-regulate via backpressure,
--- a/internal/apiserver/batch/batch_handler.go
+++ b/internal/apiserver/batch/batch_handler.go
@@ -51,18 +51,16 @@ type BatchAPIHandler struct {
 	endpointAllowlist openai.EndpointAllowlist
 }
 
-func NewBatchAPIHandler(config *common.ServerConfig, clients *clientset.Clientset) *BatchAPIHandler {
+func NewBatchAPIHandler(config *common.ServerConfig, clients *clientset.Clientset) (*BatchAPIHandler, error) {
 	endpointAllowlist, err := openai.NewEndpointAllowlist(config.BatchAPI.ExtraEndpoints)
 	if err != nil {
-		// ServerConfig.Validate rejects this at startup. Fail closed if a caller
-		// constructs an unvalidated config directly.
-		endpointAllowlist = openai.EndpointAllowlist{}
+		return nil, fmt.Errorf("extra endpoints: %w", err)
 	}
 	return &BatchAPIHandler{
 		config:            config,
 		clients:           clients,
 		endpointAllowlist: endpointAllowlist,
-	}
+	}, nil
 }
 
 func (c *BatchAPIHandler) GetRoutes() []common.Route {
--- a/internal/apiserver/batch/batch_handler.go
+++ b/internal/apiserver/batch/batch_handler.go
@@ -46,14 +46,22 @@ import (
 var _ common.ApiHandler = (*BatchAPIHandler)(nil)
 
 type BatchAPIHandler struct {
-	config  *common.ServerConfig
-	clients *clientset.Clientset
+	config            *common.ServerConfig
+	clients           *clientset.Clientset
+	endpointAllowlist openai.EndpointAllowlist
 }
 
 func NewBatchAPIHandler(config *common.ServerConfig, clients *clientset.Clientset) *BatchAPIHandler {
+	endpointAllowlist, err := openai.NewEndpointAllowlist(config.BatchAPI.ExtraEndpoints)
+	if err != nil {
+		// ServerConfig.Validate rejects this at startup. Fail closed if a caller
+		// constructs an unvalidated config directly.
+		endpointAllowlist = openai.EndpointAllowlist{}
+	}
 	return &BatchAPIHandler{
-		config:  config,
-		clients: clients,
+		config:            config,
+		clients:           clients,
+		endpointAllowlist: endpointAllowlist,
 	}
 }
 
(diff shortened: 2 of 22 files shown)


charts/batch-gateway/values.yaml
```diff
@@ -2,6 +2,10 @@
 global:
   imagePullSecrets: []
 
+  # Additional inference API paths accepted by both the API server and processor.
+  # Standard OpenAI Batch API endpoints are always allowed.
+  extraEndpoints: []
```
hhzhang16, line 10: Can this be rolled out safely? During a Helm upgrade, a new API-server replica can start accepting `/v1/classify` while an old processor replica still has the fixed allowlist and can fail the job during preprocessing. I think we need a processor-first staging path, such as component-specific overrides with the global value as the default. Otherwise, submissions need to be paused while enabling this feature.

internal/apiserver/batch/batch_handler.go
```diff
@@ -46,14 +46,22 @@ import (
...
+	endpointAllowlist openai.EndpointAllowlist
 }
 
 func NewBatchAPIHandler(config *common.ServerConfig, clients *clientset.Clientset) *BatchAPIHandler {
+	endpointAllowlist, err := openai.NewEndpointAllowlist(config.BatchAPI.ExtraEndpoints)
+	if err != nil {
+		// ServerConfig.Validate rejects this at startup. Fail closed if a caller
+		// constructs an unvalidated config directly.
+		endpointAllowlist = openai.EndpointAllowlist{}
+	}
```
hhzhang16, line 58: ServerConfig.Validate already validates these endpoints, but this constructor validates them again and silently falls back to the defaults on failure, which gives us two validation paths with different error behavior. I think it would be better if we construct the allowlist once per component and pass the validated value through; if this constructor still owns the construction, it should return the error rather than changing the configured behavior.

internal/processor/worker/preprocessor.go
```diff
@@ -303,9 +303,12 @@ func extractAndValidateLine(line []byte) (requestMeta, error) {
 	if !strings.HasPrefix(req.URL, "/") || strings.HasPrefix(req.URL, "//") || strings.Contains(req.URL, "://") {
 		return requestMeta{}, fmt.Errorf("url must be a relative path: %s", req.URL)
 	}
-	if !openai.IsValidEndpoint(req.URL) {
+	if !endpointAllowlist.IsValid(req.URL) {
 		return requestMeta{}, fmt.Errorf("invalid endpoint: %s", req.URL)
 	}
+	if batchEndpoint != "" && req.URL != batchEndpoint {
```
hhzhang16: Can we require the request URL to match the batch endpoint unconditionally? Batches created through the API always have an endpoint, so the empty case looks like it mainly exists to preserve the old test calls. I’d rather treat an empty stored batch endpoint as invalid (and update the tests to pass the endpoint they expect)

internal/shared/openai/batch.go
```diff
@@ -37,23 +39,90 @@ const (
...
+	EndpointEmbeddings,
+	EndpointCompletions,
+	EndpointModerations,
+}
+
+// EndpointAllowlist reports which batch endpoints are accepted. The zero value accepts
+// exactly the OpenAI Batch API endpoints.
+type EndpointAllowlist struct {
+	extra map[Endpoint]struct{}
+}
```
hhzhang16: This package-level array is mutable -- while it can't be appended to or resliced, its elements can be reassigned. I would instead keep the existing switch (which was smaller and harder to misuse) as the canonical check for default endpoints and store only configured extensions in the allowlist.

hhzhang16 (commented):
Before marking it as ready, could you run end to end against a Dynamo pooling/classification deployment, include the commands and results in the PR, and add a regression that verifies that a configured non-default endpoint reaches the inference client unchanged?

charts/batch-gateway/values.yaml
elizabetht: Addressed by adding component-specific overrides with `global.extraEndpoints` as the default. null inherits the global value, while a list including [] overrides it.

This supports a processor-first staged rollout:

1. Configure `processor.config.extraEndpoints` and wait for the processor rollout to complete.
2. Enable the endpoints through `global.extraEndpoints`.
3. Reset the processor override to null.

The rollout procedure is documented in the chart README and values.yaml, with Helm tests covering inheritance and override behavior.

pkg/clients/inference/inference_client_test.go
```diff
@@ -240,36 +240,42 @@ func testGenerate(t *testing.T) {
...
 
-	t.Run("should use endpoint from request", func(t *testing.T) {
-		endpoint := ""
-		testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			endpoint = r.URL.Path
-			w.WriteHeader(http.StatusOK)
-			_ = json.NewEncoder(w).Encode(map[string]interface{}{"id": "test"})
-		}))
-		t.Cleanup(testServer.Close)
+	t.Run("should preserve configured non-default endpoint", func(t *testing.T) {
```
hhzhang16: This test still starts inside `InferenceHTTPClient` with the endpoint already populated, so it doesn’t verify the new configuration and processor wiring; the client already forwarded `req.Endpoint` before this PR

charts/batch-gateway/values.yaml
hhzhang16: Thanks for addressing; what about the corresponding disable/rollback path?

hhzhang16 (commented):
Can you update the PR description?

internal/shared/openai/batch.go
```diff
@@ -56,6 +89,36 @@ func IsValidEndpoint(endpoint string) bool {
...
+		return errors.New("must be an absolute path beginning with exactly one slash")
+	}
+	if strings.ContainsAny(endpoint, "?#") {
+		return errors.New("must not contain a query string or fragment")
+	}
+	parsed, err := url.ParseRequestURI(endpoint)
+	if err != nil {
+		return fmt.Errorf("must be a valid URL path: %w", err)
+	}
+	if parsed.Path != endpoint || parsed.RawQuery != "" || parsed.Fragment != "" {
```
lioraron, line 106: Please reject dot-segment paths here (for example, require `path.Clean(endpoint) == endpoint`, and add coverage). `/v1/allowed/../admin` passes the current checks and then is forwarded as a request URL; common proxies normalize it to `/v1/admin`. That defeats the intended canonical endpoint allowlist and can dispatch to a different backend route than the operator configured.

lioraron (commented):
Please resolve the code conflicts?

internal/processor/worker/worker.go
```diff
@@ -89,18 +90,24 @@ func NewProcessor(
...
-		asyncInference: clients.AsyncInference,
-		files:          newFileManager(clients.File, clients.FileDB),
+		cfg:               cfg,
+		endpointAllowlist: endpointAllowlist,
+		processorID:       processorID,
+		poller:            poller,
+		updater:           updater,
+		batchDB:           clients.BatchDB,
+		event:             clients.Event,
+		inflight:          clients.InFlight,
```
lioraron, line 107: Current `Processor` and `Clientset` no longer have the `inflight: clients.InFlight` field.

lioraron (commented):
The API server and processor independently apply the runtime allowlist, so a batch accepted before an endpoint is removed can later fail preprocessing after a processor rollout. Any suggestions on this?

internal/processor/worker/preprocessor.go
```diff
@@ -282,7 +282,7 @@ type requestMeta struct {
 
 // extractAndValidateLine parses and validates a request line and returns the
 // metadata needed during ingestion.
-func extractAndValidateLine(line []byte) (requestMeta, error) {
+func extractAndValidateLine(line []byte, endpointAllowlist openai.EndpointAllowlist, batchEndpoint string) (requestMeta, error) {
```
hhzhang16, line 285: This signature change wasn’t applied to the callers in `planner_content_test.go`; four calls still use the old one-argument form, so the worker package doesn’t compile. Please update those tests!

Issue #657, opened 2026-08-20
[Feature]:  Add shared Batch Gateway + llm-d Async end-to-end coverage


hhzhang16:
### Problem Statement

The Batch Gateway → llm-d Async integration landed in #458, and the basic path works, but we do not currently have one place defining and testing the full end-to-end contract across both components:

Batch API client → Batch Gateway API/storage → Batch Processor → request queue → llm-d Async → OpenAI-compatible inference endpoint → result queue → Batch Processor → durable output/error files

I validated the released components against an OpenAI-compatible NVIDIA Dynamo frontend/router using [ai-dynamo/dynamo#12784](https://github.com/ai-dynamo/dynamo/pull/12784) as the deployment baseline:

- Batch Gateway chart/images: `0.3.0`
- llm-d Async chart/image: `v0.9.0`
- Dynamo runtime: `1.3.0`
- Model: `Qwen/Qwen3-0.6B`
- Sync control: 2/2 requests completed and produced the expected output file.
- Async path: 2/2 requests completed, Async reported 2/2 successful requests, and both Redis queues drained to zero.
- The Dynamo frontend logs confirmed that both Async requests traversed the Dynamo router and were assigned to a worker.
- The same client also verified request-level error-file handling and queued cancellation in both modes.

The happy path is there. Failure testing with larger batches also found gaps where accepted work can disappear, inference can continue after cancellation, or a batch can remain stuck:

- [llm-d/llm-d-async#404](https://github.com/llm-d/llm-d-async/issues/404): requests can be lost after Async removes them from Redis and the process exits before publishing a durable result.
- [#644](https://github.com/llm-d/llm-d-batch-gateway/issues/644), fixed by [#662](https://github.com/llm-d/llm-d-batch-gateway/pull/662): multiple healthy Batch Processor replicas could consume one another's results.
- [#645](https://github.com/llm-d/llm-d-batch-gateway/issues/645): a replacement Processor cannot resume an in-progress batch after the original pod/node and its local working files are lost.
- [llm-d Async in-flight cancellation](https://github.com/llm-d/llm-d-async/issues/422): cancellation is checked before inference dispatch, but does not currently cancel an already-executing inference request.
- [llm-d Async durable result delivery](https://github.com/llm-d/llm-d-async/issues/423): `Producer.GetResult` destructively removes a result before the consumer durably checkpoints it, leaving a separate result-loss window if the Processor exits.

[...]

PR #666, opened 2026-09-02
test(e2e): cover composed batch dispatch gate

hhzhang16:
## Why is this PR needed?

The existing test exercised llm-d-async’s Producer directly. This test verifies the real consumer path: Files API → Batch API → batch processor → llm-d-async dispatch gate → inference

## What does this PR do?

Adds the missing composed dispatch-gate E2E tracked under #657.

## How was this tested?

- Focused composed dispatch-gate E2E passed
- Full dispatcher E2E passed: 12 passed, 0 failed
- `make test` passed: 1,156 passed, 0 failed, 17 skipped
- `gofmt`, shell syntax, and `git diff --check` passed

## Checklist

## Related Issues

Related to #657

### DIFF
--- a/test/e2e/dispatcher_test.go
+++ b/test/e2e/dispatcher_test.go
@@ -44,6 +44,7 @@ var (
 	gatePool              = "sim-pool-gate"
 	gateReqQueue          = "llm-d-async:requests:" + gatePool
 	gateResultQueue       = "llm-d-async:results:" + gatePool
+	gateModel             = "sim-model-gate"
 	dispatchGateBudgetKey = "dispatch-gate-budget"
 
 	scrapePool        = "sim-pool-scrape"
@@ -149,6 +150,9 @@ func TestDispatcher(t *testing.T) {
 	})
 	t.Run("DispatchGate", func(t *testing.T) {
 		testDispatcherRedisGate(t, rdb)
+	})
+	t.Run("BatchAPIDispatchGate", func(t *testing.T) {
+		testBatchAPIDispatchGate(t, rdb)
 	})
 	t.Run("EndpointScrapeGate", func(t *testing.T) {
 		testDispatcherEndpointScrapeGate(t, rdb)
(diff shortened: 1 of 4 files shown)


test/e2e/dispatcher_test.go
```diff
@@ -522,6 +526,171 @@ func testDispatcherRedisGate(t *testing.T, rdb *redis.Client) {
...
+	if err != nil && err != redis.Nil {
+		t.Fatalf("read prior dispatch gate budget: %v", err)
+	}
+	t.Cleanup(func() {
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		defer cancel()
+
+		if err := rdb.Set(cleanupCtx, dispatchGateBudgetKey, "1.0", 0).Err(); err != nil {
+			t.Errorf("cleanup: open dispatch gate: %v", err)
+		} else {
```
lioraron: The cleanup drain loop uses a 5-second deadline, but the drain path here traverses the full batch API stack (batch processor → llm-d-async dispatch gate → inference), which is considerably slower than a direct producer enqueue. Every other gate drain wait in this file uses 30 seconds (lines 495, 565, 718). Under CI load this will fire spuriously. Suggest raising this to 30 seconds to match the established pattern.

test/e2e/dispatcher_test.go
```diff
@@ -522,6 +526,171 @@ func testDispatcherRedisGate(t *testing.T, rdb *redis.Client) {
...
+				if depth == 0 {
+					break
+				}
+				time.Sleep(200 * time.Millisecond)
+			}
+			if depth != 0 {
+				t.Errorf("cleanup: gated queue did not drain after opening: depth=%d", depth)
+			}
+		}
+
```
lioraron, line 571: The budget restore runs unconditionally even when the drain loop above times out with `depth != 0`. When that happens and `budgetExisted` is `false` (which it will be in normal test ordering, since `testDispatcherRedisGate` deletes the key via `defer rdb.Del`), this branch runs `rdb.Del(dispatchGateBudgetKey)`, closing the gate on a request still sitting in the queue. If the dispatcher treats a missing key as a closed gate, the zombie request never drains and the next subtest sees a dirty queue. Only restore the previous budget after confirming the queue is empty:

```go
if depth != 0 {
    t.Errorf("cleanup: gated queue did not drain after opening: depth=%d; leaving budget at 1.0", depth)
    return
}
if budgetExisted {
    ...
} else ...
```

test/e2e/dispatcher_test.go
```diff
@@ -522,6 +526,171 @@ func testDispatcherRedisGate(t *testing.T, rdb *redis.Client) {
 	}
 }
```
lioraron, line 528: Missing coverage: cancel-while-gated. The existing `BatchCancel`/`BatchCancelBeforeProcessing` tests don't use the dispatch gate, and this test doesn't cancel. The untested path is: close gate → create batch → verify request lands in `gateReqQueue` → cancel the batch → open gate → wait for queue to drain → assert sim received 0 new requests. This exercises whether the llm-d-async consumer correctly checks the cancellation marker before dispatching a gated request once the gate opens. A cancelled batch could otherwise complete successfully if the check is missing or happens at the wrong point.

test/e2e/dispatcher_test.go
hhzhang16: Addressed

test/e2e/dispatcher_test.go
hhzhang16: llm-d-async treats a missing Redis budget key as 1.0, so deleting it keeps the gate open. That said, the same risk exists if the previous value was 0, so returning while leaving the gate open is right

test/e2e/dispatcher_test.go
hhzhang16: This is a useful scenario, but I think it’s separate from the #657 acceptance item this MR addresses. #657 tracks queued cancellation and dispatch-budget pause/resume independently; this test specifically proves the latter resumes and completes exactly once. I can add cancel-while-gated in a different MR and don't think it should block this one

lioraron (approved):

PR #679, opened 2026-09-09
fix(processor): keep model plan filenames unique

git-jxj:
## Why is this PR needed?

A batch containing `org/model`, `org_model`, and `org_model_2` assigns the last two models the same plan filename. Reversing the model mapping then loses one model, and requests are dispatched with a route model that differs from `body.model`. A naturally suffixed name allocated before the collision also triggers the bug.

## What does this PR do?

Check every suffixed candidate against all allocated names and reserve the final candidate. Existing names remain unchanged when there is no collision. Add a regression that writes real plan/input files and verifies each request is produced exactly once with its original model, across all six insertion orders and multiple occupied suffixes.

## How was this tested?

- The new routing regression failed on main before the fix and passes after it.
- `go test -race ./...` passes, including regression tests.
- `go test -tags=integration ./...` passes; infrastructure-dependent tests may skip using the suite's existing checks.
- `go test -race ./internal/processor/worker -count=1` passes on the final patch.
- `golangci-lint run --timeout=5m ./...` passes with v2.11.4 (0 issues).
- No deployed-model or cluster e2e run: this bug is reproduced through the real planner and plan-file source locally.

## Checklist

- [x] Commits are signed off (`git commit -s`) per DCO.
- [x] Code follows the contributing guidelines.
- [ ] Full `make ci`.
- [ ] Deployed `make test-e2e`.


wseaton (approved):
LGTM, thanks for the fix!

wseaton (approved):
LGTM, thanks for the thorough test! Pulled it locally, the new test fails on main in every ordering and passes with the fix, and the full e2e suite is green with it applied.

PR #680, opened 2026-09-09
fix(processor): accept text content parts during preprocessing

git-jxj:
## Why is this PR needed?

Batch preprocessing rejects valid chat content arrays. For example, `"content":[{"type":"text","text":"hello"}]` fails metadata decoding with `cannot unmarshal array ... content of type string`, while the equivalent string succeeds. One such request aborts preprocessing for the batch.

## What does this PR do?

Keep message content as `json.RawMessage` during metadata extraction. Decode only system messages needed to find the first non-empty system-prefix text, accepting strings or arrays and concatenating text parts. User/assistant content and later messages remain encoded, avoiding unnecessary parsing. Invalid system content encountered during prefix extraction returns a contextual error; other content is left for the inference endpoint to validate.

The original request body is read from the input file and forwarded unchanged. Regression tests exercise real plan/input files and `PlanFileSource`, checking prefix hashes and decoded-body preservation for string, text-part, mixed, empty and null content. Additional tests cover missing/empty system messages, lazy decoding and invalid system content.

## How was this tested?

Validation of the final revision:

- `go test ./internal/processor/worker -count=1` passes.
- `golangci-lint run ./internal/processor/worker/...` passes with 0 issues.
- `git diff --check` passes.
- `go test ./internal/processor/worker -run '^$' -bench '^BenchmarkContentPreprocessing$' -benchmem -count=3` compares this revision with the previous eager-decoding implementation on the same machine. Median time for a long user string drops from 11,136 to 6,393 ns/op; for 100 user text parts, from 88,109 to 26,021 ns/op. Text-part allocations drop from 247 to 22 and bytes from 15,584 to 4,984 per operation. These are metadata-extraction microbenchmarks, not end-to-end throughput measurements.

The original content-array regression was also verified to fail before the fix. No deployed-model or cluster e2e run was performed.

## Checklist

- [x] Commits are signed and signed off per DCO.
- [x] Code follows the contributing guidelines.
- [ ] Full `make ci`.
- [ ] Deployed `make test-e2e`.

## Related Issues

#657 tracks broader test coverage; this PR fixes the narrower text content-array preprocessing failure.


wseaton (changes requested):
Thanks for the report and the proposed fix! But I think this needs a second look for performance reasons, compared to a slightly different approach (using json.RawMessage)

decoder_relative

preprocess_cpu

ex.

```go
type planRequestLine struct {
      // ...
      Messages []struct {
              Role    string          `json:"role"`
              Content json.RawMessage `json:"content"` // was string
      } `json:"messages"`
}

for _, msg := range req.Body.Messages {
      if msg.Role != "system" {
              continue
      }
      text, err := messageText(msg.Content)
      if err != nil {
              return requestMeta{}, fmt.Errorf("system message content: %w", err
      }
      if text != "" {
              prefixHash = fnv32a(text)
              break
      }
}


func messageText(content json.RawMessage) (string, error) {
      content = bytes.TrimSpace(content)
      if len(content) == 0 || bytes.Equal(content, []byte("null")) {
              return "", nil
      }
      switch content[0] {
      case '"':
              var s string
              err := json.Unmarshal(content, &s)
              return s, err
      case '[':
              var parts []struct {
                      Type string `json:"type"`
                      Text string `json:"text"`
              }
              if err := json.Unmarshal(content, &parts); err != nil {
                      return "", err
              }
              var b strings.Builder
              for _, p := range parts {
                      if p.Type == "text" {
                              b.WriteString(p.Text)
                      }
              }
              return b.String(), nil
      default:
              return "", errors.New("must be a string or an array of content parts")
      }
}

```

wseaton (approved):

PR #681, opened 2026-09-09
feat(shutdown): consolidate graceful shutdown into internal/util/shutdown

### LINKED ISSUE #650: [Feature]: consolidate graceful shutdown logic
### Problem Statement

From https://github.com/llm-d/llm-d-batch-gateway/pull/647#discussion_r3813861034 — apiserver, processor, and batch-gc each hand-roll termination with their own ordering and timeout knobs. It all works, but the sequence is implicit and each are implemented slightly differently in each main.go.

### Proposed Solution

one shared pattern:

signal → stop intake → drain → close clients → flush observability, with a normative shutdown budget. Small helper in internal/util that the three subcomponent mains can call

### Alternatives Considered

_No response_

### Willingness to Contribute

Yes, I can submit a PR

### Additional Context

_No response_

highpon:
## Why is this PR needed?

apiserver, batch-processor, and batch-gc each shutdown logic, with slightly different

## Why is this PR needed?

apiserver, batch-processor, and batch-gc each shutdown logic, with slightly different ordering and timeout handling in every `main.go` (#650). The shutdown sequence was implicit and duplicated three times,
making it hard to reason about.

## What does this PR do?

- Introduces `internal/util/shutdown`, a sharedse`, `Guard`, `WatchIntake`) that implements one normative shutdown sequence — signal → stop intake → drain → close clients → flush observability — bounded by an overall shutdown budget.
- Adopts this shared package in `cmd/apiserver`processor`, and`internal/apiserver/server/server.go`, replacing the ad hoc `context.WithTimeout` + deferred-cleanup code that existed in each binary.
- Adds `Guard` as a safety net that closes/flushes a resource on any early return before the `Coordinator` takes ownership of it, preventing resource leaks and double-close bugs.
- Also fixes several issues found during review
  - `Guard.Cleanup` declared a timeout on `Close()`-style functions that don't accept a `context.Context`, so the timeout
could never actually take effect; these now run consistent with how the `Coordinator`'s own"close clients" phase already handles this.
  - `batch-processor`'s "flush observability" ano explicit per-phase timeout, unlike theequivalent phases in the other two binaries.
  - `batch-gc` unconditionally set `ready.Store(true)` without checking `ctx.Err()`, which could clobber the
shutdown-triggered `ready.Store(false)` from `Wd during client/reconciler setup, leaving thereadiness probe reporting "ready" throughout the drain.

## How was this tested?

- [x] Unit tests added/updated/verified — added `internal/util/shutdown/shutdown_test.go` (phase ordering, error
aggregation, budget enforcement, Guard arm/disa(e.g. `internal/apiserver/server`) still pass.

## Checklist

## Related Issues

Fixes #650


internal/apiserver/server/server.go
```diff
@@ -145,6 +153,7 @@ func (s *Server) Start(ctx context.Context) error {
 			logger.Error(err, "observability server failed")
 		}
 	}()
+	obsGuard := shutdown.NewGuard(logger, "flush observability", time.Duration(s.config.GetObservabilityShutdownTimeoutSeconds())*time.Second, obsServer.Shutdown)
```
lioraron, line 156: `obsGuard` is never deferred. If `net.Listen` (or a later TLS/startup step) returns an error, `Start` returns after the observability goroutine has started but before the guard is disarmed, so that server keeps listening. Add `defer obsGuard.Cleanup()` immediately after constructing the guard; it becomes a no-op on the normal coordinated path.

highpon: Thank you for your review
fix: https://github.com/llm-d/llm-d-batch-gateway/pull/681/changes/cf7f70d41d784c0413d428d05c4c0968b17a787f

cmd/batch-gc/main.go
```diff
@@ -107,14 +105,52 @@ func run() error {
...
 		return err
 	}
 
+	// Stop intake as soon as the shutdown signal arrives, ahead of the drain
+	// that follows from g.Wait() below.
+	shutdown.WatchIntake(ctx, func() { ready.Store(false) })
+
+	const flushObservabilityTimeout = 5 * time.Second
+	const flushTracerTimeout = 5 * time.Second
+	coordinator := shutdown.New(logger, flushObservabilityTimeout+flushTracerTimeout+shutdown.DefaultSlack)
```
acardace: DefaultSlack is manually added to all new coordinators, this means that either you can just do it internally in the Coordinator code or that is not really needed. Why did you add it?

acardace: I think a better design is to have all Phases declare their timeouts (like already done now) and then the coordinator can simply compute the whole budget.

Now you repeat the timeouts twice just to assign the `budget` field in the Coordinator.

internal/util/shutdown/shutdown.go
```diff
@@ -0,0 +1,134 @@
...
+		<-ctx.Done()
+		stop()
+	}()
+}
+
+// Guard is a safety net for a resource a Coordinator will eventually take
+// ownership of. If the caller returns before Disarm is called (e.g. a later
+// initialization step fails), Cleanup runs fn once as a fallback so the
+// resource isn't leaked; once Disarm has been called, Cleanup is a no-op.
+type Guard struct {
```
acardace: I'm not convinced we need a new `Guard` abstraction, I'd drop it and use plain defers.

In batch-gc/main.go both guards are disarmed right after the coordinator is built, so they only cover 1-3 early returns. Their only job is exactly-once, but nothing guarded needs it: Clientset.Close() already closes the same redis client 2–3x per call and is sync.Once-safe (redis_ds_client.go:131, db_core.go:477), and both http.Server.Shutdown and TracerProvider.Shutdown return nil on second call.

A missed Disarm (or a plain defer + coordinator) is at worst a no-op second pass, not a bug. Not worth a new NewGuard/Disarm/Cleanup API with inverted Timeout <= 0 semantics vs Phase, 3 error paths (or even plain defers) are clearer.

internal/util/shutdown/shutdown.go
```diff
@@ -0,0 +1,134 @@
...
+	Name    string
+	Timeout time.Duration
+	Run     func(context.Context) error
+}
+
+// WatchIntake calls stop as soon as ctx is done, ahead of any drain that
+// follows. It runs independently of a Coordinator: for workloads whose
+// drain loop blocks the goroutine that will eventually call Coordinator.Run,
+// waiting for the coordinator's own phases would stop intake too late.
+func WatchIntake(ctx context.Context, stop func()) {
```
acardace: I don't think we need this either, context.AfterFunc does exactly this but without using an additional goroutine.

cmd/batch-gc/main.go
```diff
@@ -107,14 +105,52 @@ func run() error {
...
 	var ready atomic.Bool
-	metricsErrCh, err := startMetricsServer(ctx, cfg.MetricsAddr, logger, &ready)
+	metricsErrCh, shutdownMetrics, err := startMetricsServer(ctx, cfg.MetricsAddr, logger, &ready)
 	if err != nil {
 		return err
 	}
 
+	// Stop intake as soon as the shutdown signal arrives, ahead of the drain
+	// that follows from g.Wait() below.
+	shutdown.WatchIntake(ctx, func() { ready.Store(false) })
```
acardace:
```
stopIntake := context.AfterFunc(ctx, func() { ready.Store(false) })
defer stopIntake()
```

internal/util/shutdown/shutdown_test.go
```diff
@@ -0,0 +1,141 @@
+package shutdown
```
acardace: Tests look good for the basics, but a couple of gaps worth covering:

- No test for per-phase Timeout. Everything uses Timeout: 0 except the overall budget expiry. Worth one case showing a short phase ctx cancels inside a long budget while the next phase still gets a live ctx.
- Cleanup twice, Run == nil, and multi-error Join are untested branches. Mainly want to lock in whether double Cleanup running fn twice is intended.

acardace (changes requested):

PR #685, opened 2026-09-11
feat: guard resumable batch takeover

tryambak2019:
## Why is this PR needed?

Issue #645 requires coordination between GC and startup recovery. Without a database-enforced guard, GC can reset, re-enqueue, or terminalize a Batch while a replacement Processor is claiming it.

Startup recovery errors are also currently logged and ignored, allowing the Processor to become ready and begin polling after recovery has failed.

## What does this PR do?

* Adds `batch_items.resumable BOOLEAN NOT NULL DEFAULT FALSE`, preserving legacy behavior for existing rows.
* Makes PostgreSQL re-enqueue require the expected `processor_id`, status, and `epoch`, with `resumable=false`.
* Prevents normal dequeue, queued cancellation, and GC reconciliation from mutating resumable Batches.
* Returns startup claim or recovery errors from `Run` before readiness, polling, or result consumption, allowing Kubernetes to restart the Processor.
* Keeps migrated epoch-zero jobs recoverable.

This is the first step toward #645. It adds the takeover guards but does not yet create resumable Batches or implement manifest/checkpoint recovery.

## How was this tested?

Added focused tests for the PostgreSQL predicates, GC exclusion, startup failure propagation, and legacy epoch-zero migration. Execution is pending repository CI because the current local environment does not include the required Go toolchain.

* [ ] Unit tests added/updated/verified
* [ ] Integration/e2e tests added/updated/verified
* [ ] Manual testing performed

## Checklist

* [x] Commits are signed off (`git commit -s`) per [DCO](https://github.com/llm-d/llm-d-batch-gateway/blob/main/PR_SIGNOFF.md)
* [ ] Code follows project [contributing guidelines](https://github.com/llm-d/llm-d-batch-gateway/blob/main/CONTRIBUTING.md)
* [ ] CI checks pass (`make ci`)
* [ ] E2E tests pass (`make test-e2e`)

## Related Issues

Progress towards #645


hhzhang16 (approved):

internal/processor/worker/recovery.go
```diff
@@ -95,16 +97,25 @@ func (p *Processor) recoverOwnedJobs(ctx context.Context) {
...
 					jlogger.Error(failErr, "Startup recovery: failed to fail job past its recovery budget")
+					if task.Resumable || errors.Is(failErr, errResumableRecoveryUnavailable) {
+						return failErr
+					}
 				}
 				return nil
 			}
 			if recoverErr := p.recoverJob(jctx, task.ID); recoverErr != nil {
 				jlogger.Error(recoverErr, "Startup recovery: failed to recover owned job")
+				if task.Resumable || errors.Is(recoverErr, errResumableRecoveryUnavailable) {
```
lioraron, line 108: An ErrConflict here is not safe to ignore. If cancellation updates a claimed legacy row after recoverJob reads it, PQEnqueue fails its expected-status check and UpdateFailedStatus now also preserves cancelling. This branch lets Run begin polling, but the row remains owned and cancelling, so it cannot be dequeued and the already-sent cancel event is never consumed. Detect the conflict and refetch/recover the cancelling state so the batch reaches a terminal state.

internal/processor/worker/recovery_epoch_pg_test.go
```diff
@@ -84,7 +84,9 @@ func TestRecoverOwnedJobsFencesPreviousEpoch(t *testing.T) {
 		t.Fatalf("seed: %v", err)
 	}
 
-	p.recoverOwnedJobs(ctx)
+	if err := p.recoverOwnedJobs(ctx); err != nil {
```
lioraron, line 104: Please add a deterministic PostgreSQL regression test that pauses recovery after it reads a claimed validating row, sends cancellation, then resumes recovery. Assert that it refetches and finalizes cancellation with no owned non-terminal row left behind. This test starts with cancellation already persisted, so it misses the failing CAS-interleaving path.

internal/processor/worker/recovery.go
tryambak2019: Addressed in signed commit `59d37ac`. On `db.ErrConflict`, recovery refetches the batch; if cancellation won the race, it finalizes cancellation instead of continuing polling with an owned non-terminal row.

internal/processor/worker/recovery_epoch_pg_test.go
tryambak2019: Addressed in `59d37ac`. Added a deterministic PostgreSQL regression test that pauses after recovery reads the claimed validating row, persists cancellation before the re-enqueue CAS, then verifies recovery refetches and leaves no owned non-terminal row. Passed against PostgreSQL 16.

lioraron (commented):
Please address ad resolve the Copilot comments?

internal/database/postgresql/batch_queue.go
```diff
@@ -73,27 +74,39 @@ func (c *PostgresBatchQueueClient) PQEnqueue(ctx context.Context, jobPriority *a
...
 			UPDATE batch_items
 			SET processor_id = NULL,
 			    status = jsonb_set(status, '{status}', '"validating"'),
 			    epoch = epoch + 1
 			WHERE id = $1
-			  AND processor_id IS NOT NULL
+			  AND epoch = $2
+			  AND processor_id = $3
+			  AND status::jsonb->>'status' = $4
+			  AND resumable = FALSE
```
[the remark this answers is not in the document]
tryambak2019: Addressed by the recovery conflict retry in `d8d788c`. `recoverJob` treats `db.ErrConflict` as a lost-state race and refetches/re-routes the job once, so a cancellation that wins before re-enqueue is finalized rather than overwritten through stale failure handling. The PostgreSQL interleaving regression test covers this path.

.github/workflows/pre-commit.yml
```diff
@@ -43,7 +57,11 @@ jobs:
         uses: pre-commit/action@v3.0.1
 
       - name: Run integration tests
-        run: make test-integration
+        run: make test-integration GOFLAGS=-p=1
```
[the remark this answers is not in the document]
tryambak2019: Addressed in `58a7f17`. The `test-integration` Make target now expands `$(GOFLAGS)`, and the workflow invokes it with `GOFLAGS=-p=1`; PostgreSQL-backed integration packages therefore run serially while they share the destructive test schema.

internal/database/postgresql/batch_queue.go
```diff
@@ -215,6 +233,7 @@ func (c *PostgresBatchQueueClient) PQDelete(ctx context.Context, jobPriority *ap
 			SELECT id FROM batch_items
 			WHERE id = $1
 			  AND processor_id IS NULL
+			  AND resumable = FALSE
```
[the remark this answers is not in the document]
tryambak2019: Addressed in `bcbf35d`. After `PQDelete` returns zero, `CancelBatch` writes with `ExpectedResumable=false`; a queued resumable row produces `ErrConflict`/HTTP 409 without a status change. `TestCancelBatchDoesNotMutateQueuedResumableBatch` verifies the row remains `validating`.

Issue #672, opened 2026-09-07
[Feature]: Standardize on async dispatching as the sole execution path


acardace:
### Problem Statement

The Batch Processor currently supports two parallel execution paths:
1. Synchronous HTTP dispatching directly to inference endpoints.
2. Asynchronous dispatching via `llm-d-async` request and result queues.

Maintaining synchronous dispatch duplicates complex flow control and rate-limiting logic inside the processor:
- In-gateway `AIMDController` for adaptive concurrency control.
- Per-endpoint adaptive semaphores.
- Direct inference HTTP client management, retry backoff, and traffic shedding logic.

This duplicate functionality adds thousands of lines of complex concurrency, backoff, and retry code that duplicates what GIE and `llm-d-async` already provide.

### Proposed Solution

- Standardize on `llm-d-async` request/result queues as the sole execution path for batch jobs.
- Remove synchronous HTTP dispatching from the processor.
- Delete the in-gateway `AIMDController`, per-endpoint adaptive semaphores, and direct inference clients.
- Delegate traffic shedding, queueing, and backpressure directly to GIE and `llm-d-async`, significantly simplifying the processor codebase.

lioraron:
While there is agreement that `batch-gateway + async-processor` should be the path to production, we should avoid removal of the sync mechanism until benchmark results clearly show that `batch-gateway + async-processor` is stable, and that it performs better than `batch-gateway + flow-control` and `batch-gateway + admission-control` at various scale levels.

acardace:
@lioraron of course we'll take this step by step.

PR #689, opened 2026-09-14
fix: harden async dispatch while retaining sync default

aneeshkp:
## Summary

Harden the existing async dispatch path while retaining synchronous HTTP
dispatch as the upstream default.

This is an incremental step toward #672. Sync removal and any future
default change are deferred until async stability and benchmark comparisons
are complete.

## Changes

- Retain `dispatch_mode: sync` as the default.
- Preserve explicit `dispatch_mode: async`.
- Fix async cancellation and expiry error classification.
- Add stronger async submission, ordering, multi-model, cancellation, and
  expiry coverage.
- Improve async deployment and dispatcher configuration.
- Add Helm validation for missing async model-to-pool mappings.
- Keep sync/AIMD/direct HTTP support unchanged.

## Verification

- [x] Full Go test suite
- [x] `go vet ./...`
- [x] 101 Helm tests
- [x] Script tests
- [x] Local PostgreSQL/Redis async success
- [x] Local async cancellation and expiry
- [x] Local explicit sync
- [x] AKS async success with real `llm-d-async v0.9.1`
- [x] AKS async cancellation
- [x] AKS explicit sync
- [ ] Production-scale benchmark comparison

## Follow-up

Benchmark async against flow-control and admission-control at multiple scale
levels before considering sync removal or changing the upstream default.
Also change the PR title to:
fix: harden async dispatch while retaining sync default


lioraron (changes requested):
This looks premature given this requirement https://github.com/llm-d/llm-d-batch-gateway/issues/672#issuecomment-5629811909

aneeshkp:
> This looks premature given this requirement [#672 (comment)](https://github.com/llm-d/llm-d-batch-gateway/issues/672#issuecomment-5629811909)

@lioraron  That's fair, My intent  was to make  this an incremental step towards #672 , defaulting to to async while keeping sync available until the stability and benchmark requirements are met.

aneeshkp:
@wseaton @vMaroon @evacchi

acardace:
@lioraron do we have benchmarks of the sync path with a production deployment of llm-d + vllm?

lioraron:
> @lioraron do we have benchmarks of the sync path with a production deployment of llm-d + vllm?

https://github.com/llm-d/llm-d-batch-gateway/issues/491

acardace:
> > @lioraron do we have benchmarks of the sync path with a production deployment of llm-d + vllm?
> 
> #491

From those benchmarks it seems like AIMD doesn't work with production vLLM. So what is the ask here? I don't think that's a very compelling benchmark to block this PR.

lioraron:
> > > @lioraron do we have benchmarks of the sync path with a production deployment of llm-d + vllm?
> > 
> > 
> > #491
> 
> From those benchmarks it seems like AIMD doesn't work with production vLLM. So what is the ask here? I don't think that's a very compelling benchmark to block this PR.

Provide benchmark results that clearly show that `batch-gateway + async-processor` is stable, and that it performs better than `batch-gateway + flow-control` and `batch-gateway + admission-control` at various scale levels -- before switching the default.

internal/processor/config/config.go
```diff
@@ -363,7 +363,7 @@ func NewConfig() *ProcessorConfig {
 		DefaultOutputExpirationSeconds: 90 * 24 * 60 * 60, // 90 days
 		ProgressTTLSeconds:             24 * 60 * 60,      // 24 hours
 
-		DispatchMode: DispatchModeSync,
+		DispatchMode: DispatchModeAsync,
```
lioraron, line 366: This also changes the checked-in default processor config into an invalid configuration. `cmd/batch-processor/config.yaml` omits `dispatch_mode` but configures only `global_inference_gateway`; it is decoded into `NewConfig()` and now validates as async, which rejects that gateway. `go run ./cmd/batch-processor --config cmd/batch-processor/config.yaml` exits with that validation error. Add `dispatch_mode: sync` to the bundled config (and migrate/update the equivalent chart examples) so the documented default command remains runnable.

aneeshkp: @lioraron Good catch, thanks.
Fixed the bundled processor config to explicitly select sync, and updated the sync Helm fixtures/examples and documentation accordingly. 

Added a regression test that loads and validates the bundled config. Helm now explicitly defaults to async and fails during rendering if model mappings are missing. 

Go tests and all 101 Helm tests pass.

charts/batch-gateway/templates/processor-configmap.yaml
```diff
@@ -1,4 +1,8 @@
 {{- if .Values.processor.enabled -}}
+{{- $dispatchMode := default "async" .Values.processor.config.dispatchMode -}}
+{{- if and (eq $dispatchMode "async") (empty .Values.processor.config.asyncDispatch.models) -}}
+{{- fail "processor.config.asyncDispatch.models must be configured when dispatchMode is async" -}}
```
madhugoutham, line 4: Could we keep omitted dispatchMode compatible with existing sync releases until the benchmark gate is met, since an upgrade with an existing globalInferenceGateway now hits this fail unless users first add dispatchMode: sync?

internal/processor/config/config.go
aneeshkp: @lioraron I have kept synchronous HTTP dispatch as the upstream default and retained async dispatch as an explicit option.  

This PR now focuses on hardening the async path: cancellation/expiry classification, async submission and ordering tests, multi-model coverage, deployment improvements, and Helm validation for missing async mappings.

charts/batch-gateway/templates/processor-configmap.yaml
aneeshkp: Sync is restored as the upstream and Helm default, so existing releases that omit dispatchMode and configure globalInferenceGateway remain compatible. Async remains available through explicit dispatchMode: async with required model-to-pool mappings. I also added regression coverage for the bundled sync configuration and Helm async validation

internal/processor/pipeline/dispatcher_async.go
```diff
@@ -78,8 +80,9 @@ func (d *AsyncDispatcher) Run(ctx context.Context, requestCh <-chan RequestItem,
 
 	// Drain submitted-but-uncollected requests as errors so that
 	// output_lines + error_lines == total_requests.
+	code, message := cancelCode(ctx)
```
madhugoutham, line 83: Could unresolved async requests keep a non-user error on SIGTERM or a source failure, since cancelCode currently writes batch_cancelled for every non-expiry cause and startup recovery will upload that error file when the job later fails with partial output?

internal/processor/config/config.go
lioraron: Thanks @aneeshkp looks good.
Pls address https://github.com/llm-d/llm-d-batch-gateway/pull/689#discussion_r4029490489 when available.

Makefile
```diff
@@ -337,13 +338,13 @@ test-all: test test-regression test-integration test-scripts
...
-## dev-deploy: Deploy batch-gateway to a local kind cluster with all dependencies
+## dev-deploy: Deploy batch-gateway in sync mode (set ENABLE_DISPATCHER=true for async fixtures)
 dev-deploy:
 	@bash scripts/dev-deploy.sh
 
-## dev-deploy-gie: Deploy with GIE integration (per-model EPP + InferenceObjectives)
+## dev-deploy-gie: Deploy with sync GIE integration (per-model EPP + InferenceObjectives)
 dev-deploy-gie:
-	@ENABLE_GIE=true bash scripts/dev-deploy.sh
+	@ENABLE_DISPATCHER=false ENABLE_GIE=true bash scripts/dev-deploy.sh
```
lioraron, line 347: Is explicit `ENABLE_DISPATCHER=false` required here?

memory.md
```diff
@@ -0,0 +1,40 @@
+# Working Context
```
lioraron: This file looks not relevant for adding to the repo?

Makefile
aneeshkp: ENABLE_DISPATCHER=false ensures the GIE deployment cannot accidentally inherit async mode from the caller’s environment.

internal/processor/pipeline/dispatcher_async.go
aneeshkp: Fixed it

lioraron (approved):
Thanks

PR #690, opened 2026-09-14
fix: handle transient file download errors in e2e tests

RishabhSaini:
## Problem

The AIMD Recovery e2e test fails intermittently with HTTP 500 when downloading the output file via `GET /v1/files/{file-id}/content`. The batch completes successfully (12/12 requests), but the file download returns a transient 500 from the Istio/ingress layer or S3/MinIO eventual consistency. The test treats any HTTP error during polling as fatal (`t.Fatalf`), making it fragile to transient errors.

## Root Cause

1. **Server**: `DownloadFile` handler returns HTTP 500 for all storage retrieval errors, including `os.ErrNotExist` (file missing from storage). Callers cannot distinguish permanent "not found" from transient storage errors.

2. **Tests**: `fetchBatchResults`, `fetchOutputFile`, `waitForBatchStatus`, and `waitForCompletedRequests` all use `t.Fatalf` on the first error with no retry logic.

## Fix

**Server** (`file_handler.go`):
- `DownloadFile` now returns 404 when storage returns `os.ErrNotExist` (file truly missing)
- Returns 500 for all other errors (transient storage failure)

**Tests** (`helpers_test.go`):
- New `downloadFileContent()` retries file downloads with exponential backoff (500ms->8s, max 5 attempts)
- Retries only on 5xx (transient), fast-fails on 404 (permanent)
- `waitForBatchStatus` and `waitForCompletedRequests` retry transient poll errors instead of `Fatalf`


test/e2e/helpers_test.go
```diff
@@ -432,19 +432,62 @@ type batchResultLine struct {
...
+
+	const (
+		maxRetries = 5
+		baseDelay  = 500 * time.Millisecond
+	)
+	client := newClient()
+
+	var lastErr error
+	for attempt := range maxRetries {
+		resp, err := client.Files.Content(context.Background(), fileID)
```
lioraron, line 459: `Files.Content` returns a non-nil error for every HTTP status >= 400 (while also populating `resp`). Consequently this `err == nil` branch never sees 404, 500, 502, etc.: 404 is retried five times instead of fast-failing, and all HTTP errors are retried indiscriminately rather than using `isRetryableStatus`. Please classify `resp.StatusCode` whenever `resp != nil` (or extract it from the API error) before deciding to retry, and add 404/500/502 coverage.

internal/apiserver/file/file_handler.go
```diff
@@ -592,6 +592,17 @@ func (c *FileAPIHandler) DownloadFile(w http.ResponseWriter, r *http.Request) {
 	storageName := ucom.FileStorageName(fileObj.ID, fileObj.Filename)
 	fileReader, fileMeta, err := c.clients.File.Retrieve(ctx, storageName, folderName)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
```
lioraron, line 595: `os.ErrNotExist` is not specific to a missing object here: the S3 `Retrieve` implementation also maps `types.NoSuchBucket` to this sentinel. A deleted or misconfigured bucket will therefore be reported as a client-side 404 for this file, rather than a storage/server failure. Please distinguish missing keys from missing buckets (for example with a storage-specific sentinel) and retain a 5xx response for the latter.

test/e2e/helpers_test.go
RishabhSaini: Fixed. Now uses `errors.As(*openai.Error)` to extract status code from the API error. 404 → fail fast, 500 → fail fast, 502/503/504 → retry with backoff.

internal/apiserver/file/file_handler.go
RishabhSaini: Fixed. S3 `Retrieve` now only maps `NoSuchKey` to `os.ErrNotExist`. `NoSuchBucket` returns the raw AWS error, so the handler returns 500 instead of 404 for bucket misconfiguration.

test/e2e/helpers_test.go
```diff
@@ -432,19 +433,65 @@ type batchResultLine struct {
...
```
lioraron, line 459: `newClient` leaves openai-go at its default of two retries for every >=500 response. Thus a 500/501 is retried before this code can fast-fail it, while each retryable 502/503/504 outer attempt can issue three requests (up to 15 total), bypassing this helper’s stated policy and backoff. Disable the SDK retry for this call with `option.WithMaxRetries(0)` so `isRetryableStatus` controls the attempts.

PR #691, opened 2026-09-16
fix(redis): propagate pipeline errors

wseaton:
## What

Propagate errors returned by Redis `Pipelined` immediately at the call site.

Previously, a pipeline-level failure was assigned to the named return error and then could be overwritten while inspecting individual command errors, allowing `PQEnqueue` to log success even though the enqueue did not reach Redis.

## Testing

- `go test ./internal/database/redis`


internal/database/redis/redis_queue.go
```diff
@@ -74,7 +74,7 @@ func (c *ExchangeDBClientRedis) PQEnqueue(ctx context.Context, item *db_api.Batc
 		return nil
 	})
 	if lerr != nil {
-		err = lerr
+		return lerr
```
lioraron, line 77: Use
```
err = lerr
return
```
to align with the rest of the code in this file?

lioraron (commented):
Thanks

lioraron:
The fix in this PR has been applied within another PR that was merged, so closing this. Thank you.

PR #692, opened 2026-09-17
feat(processor): derive worker count from input disk budget

lioraron:
## Summary

- derive the batch processor effective worker count at startup from the configured worker ceiling, work-directory capacity, input-file disk budget, and maximum input-file size
- default the processor worker ceiling to 5 and reserve 10% of the 10Gi work directory for staged input files
- pass the work-directory limit and API input-file maximum through the Helm processor configuration; add unit and chart coverage
- small added fix: return immediately when the Redis pipeline reports an error, preventing subsequent handling of an invalid command-result slice

## Validation

- [x] `go test ./... -count=1`
- [x] `go vet ./internal/processor/config ./internal/processor/worker`
- [x] `helm unittest charts/batch-gateway`
- [x] `git diff --check`
- [x] `make test-helm`
- [x] `make test-all`
- [x] `make pre-commit`
- [x] `make test-e2e`


j-mok-dev:
Thanks @lioraron ! I think `total_workers` should reflect `EffectiveNumWorkers()` so worker saturation (active_workers / total_workers) matches the actual semaphore capacity when the disk budget limits `num_workers`.

charts/batch-gateway/templates/processor-configmap.yaml
```diff
@@ -17,6 +17,9 @@ data:
     task_wait_time: {{ .Values.processor.config.taskWaitTime | quote }}
     poll_interval: {{ .Values.processor.config.pollInterval | quote }}
     num_workers: {{ .Values.processor.config.numWorkers }}
+    work_dir_size_limit: {{ .Values.processor.workDirVolume.sizeLimit | quote }}
+    input_file_disk_budget_percent: {{ .Values.processor.config.inputFileDiskBudgetPercent }}
+    input_file_max_size_bytes: {{ .Values.apiserver.config.fileAPI.maxSizeBytes | int64 }}
```
yizhaodev: api server also has this in values.yaml `Values.apiserver.config.fileAPI.maxSizeBytes`
I suggest we define `Values.global.maxInputFileSizeBytes`, and have api server/process to use same value.

lioraron:
> Thanks @lioraron ! I think `total_workers` should reflect `EffectiveNumWorkers()` so worker saturation (active_workers / total_workers) matches the actual semaphore capacity when the disk budget limits `num_workers`.

Thanks @j-mok-dev for catching this! Addressed in [b45feb4](https://github.com/llm-d/llm-d-batch-gateway/pull/692/commits/b45feb4fcf8665dfd54d3c405f76e77d542fb800): `total_workers` now reports `EffectiveNumWorkers()` rather than the configured ceiling, so the dashboard and saturation alert denominator matches the actual worker semaphore capacity. I also added coverage for a configured 20 workers capped to 5 by the input-file disk budget.

internal/processor/config/config.go
```diff
@@ -343,8 +372,11 @@ func NewConfig() *ProcessorConfig {
 				AdditiveIncrease: 1,
 			},
 		},
-		NumWorkers: 1,
-		Addr:       ":9090",
+		NumWorkers:                 1,
```
yizhaodev: change to 5? same as values.yaml

yizhaodev (approved):

j-mok-dev (approved):

charts/batch-gateway/templates/processor-configmap.yaml
lioraron: Good suggestion, thanks! Addressed in [3ac3080](https://github.com/llm-d/llm-d-batch-gateway/pull/692/commits/3ac3080d47f09993cb657a9819cd3c985323ea5a): `global.maxInputFileSizeBytes` is now the shared source of truth. Both the API server’s `file_api.max_size_bytes` and the processor’s `input_file_max_size_bytes` are rendered from this value, and chart tests verify a custom value reaches each component.

internal/processor/config/config.go
lioraron: Done, thanks.
Updated this to use `DefaultNumWorkers = 5`.

PR #695, opened 2026-09-19
test(tls): cover Certificates.IsEmpty and JoinCertPath

magic-peach:
## Why is this PR needed?

`Certificates.IsEmpty()` and `JoinCertPath()` in `internal/util/tls` had no direct test coverage.

## What does this PR do?

Adds table driven tests for `IsEmpty()` (zero value, dir only, all fields set) and `JoinCertPath()` (empty file, non empty file, empty dir).

## How was this tested?

`go test ./internal/util/tls/... -v -cover` passes, package coverage went from 30.8% to 46.2%. `go vet` and `gofmt -l` are clean.

## Checklist

## Related Issues

N/A


internal/util/tls/tls_test.go
```diff
@@ -5,6 +5,74 @@ import (
...
+			want:  true,
+		},
+		{
+			name:  "cert dir set is not empty",
+			certs: Certificates{Dir: "/etc/certs"},
+			want:  false,
+		},
+		{
+			name:  "all fields set is not empty",
+			certs: Certificates{Dir: "/etc/certs", CertFile: "tls.crt", KeyFile: "tls.key", CaCertFile: "ca.crt"},
```
lioraron: Please add cases where CertFile, KeyFile, and CaCertFile are each set independently (with Dir empty). This row still passes if IsEmpty is accidentally changed to inspect only Dir, because Dir is populated here, even though the Redis client relies on any non-zero certificate field to enable TLS loading. The new direct coverage would then miss a regression in the helper it is meant to protect.

